PR #3, opened 2023-05-23
Adding the concept of custom OpenAI

jespino:


crspeller:
@jespino Not sure I like the name "Custom OpenAI". Maybe "OpenAI Compatible"?

jespino:
@crspeller done, PTAL

jespino:
@crspeller Yes, probably the MattermostAI settings are going to disappear and use directly the OpenAI Compatible.

PR #8, opened 2023-05-31
Simplify config and bring back OpenAI compatible service

jespino:


crspeller (approved):
Thanks! Forgot to update the config with that refactor.

PR #9, opened 2023-06-02
Change default model to GPT-3.5

chenilim:
#### Summary
Since GPT-4 is still in Beta, the AI bot will fail if the dev key doesn't have access to it, and a default model isn't specified in the settings. This change will use GPT-3.5-turbo as the default if a model isn't specified in settings.

The workaround is to set "gpt-3.5-turbo" as the default model in settings.


crspeller (approved):
Makes sense!

PR #11, opened 2023-06-14
Add Gitpod demo

azigler:
I added a working Gitpod demo for this project, and the instructions for setting it up. This works out-of-the-box with OpenAI as OpenAI longer seems to be invalidating API keys used in Gitpod workspaces.

cc @crspeller @chenilim 

Next steps could be adding the `mattermost-ai-server` to this setup and getting consistent behavior from a local LLM, but this example works for remote LLMs as-is.


crspeller (approved):
Tested and works for me. Thanks!

PR #12, opened 2023-06-16
Feat/change tone plus spellcheck

fmartingr:
#### Summary

This commit adds to commands: 
- `/spellcheck [message]`: Corrects the spell check for the message.
- `/change_tone [tone] [message]`: Changes the tone for the provided message (allowed: `casual` & `professional`).

My main idea was to add some buttons/menus on the post editor to toggle this directly from there and modify the user input directly, before sending the message, but since the Mattermost Plugin Framework doesn't allow customizing the post editor, there is that. (Maybe for another OSF?)

### DIFF
--- a/server/ai/prompts/change_tone.tmpl
+++ b/server/ai/prompts/change_tone.tmpl
@@ -1,0 +1,10 @@
+{{define "change_tone.system"}}
+Change tone and voice to {{.Tone}}.
+{{if eq .Tone "professional"}}
+Use industry-specific language and terminology.
+Avoid complex words.
+{{end}}
+Provide only the resulting text as an answer.
+Avoid greetings and sign-offs.
+The original text: "{{.Message}}"
+{{end}}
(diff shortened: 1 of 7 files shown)


server/ai/prompts/change_tone.tmpl
```diff
@@ -0,0 +1,10 @@
+{{define "change_tone.system"}}
+Change tone and voice to {{.Tone}}.
+{{if eq .Tone "professional"}}
+Use industry-specific language and terminology.
+Avoid complex words.
+{{end}}
+Provide only the resulting text as an answer.
+Avoid greetings and sign-offs.
+The original text: "{{.Message}}"
```
crspeller: Can you put the message into the user message? I don't think we really want to have raw user input into the system message.

server/ai/prompts/spellcheck.tmpl
```diff
@@ -0,0 +1,3 @@
+{{define "spellcheck.system"}}
+Correct the spelling for the following: "{{.Message}}"
```
crspeller: Same thing here. Avoid raw user input in the system message.

crspeller (changes requested):
This is great! Thanks for the contribution! 🎉

crspeller (approved):
Awesome!

PR #26, opened 2023-07-10
Add support for Azure OpenAI

jespino:
#### Summary
Adding a special condition to make Custom API work in OpenAI from Azure


crspeller (changes requested):
Do you mind adding some documentation for this? Probably just a note under the "Supported Backends" section in the README would be sufficient.

crspeller (approved):

PR #35, opened 2023-07-14
Adding the code needed for independent post menu for AI

### LINKED ISSUE #15: Webapp capability: Add directly to post context menu.
It would be valuable to be able to add the AI tools directly to the post menu. Currently this is not supposed and plugins are relegated to the "message actions" menu.

Should support doing something like this: 
image

[Figma Link](https://www.figma.com/file/4Dq6Kl0kq8Zf186Ygf4NWT/AI-Explorations?type=design&node-id=1-140762&t=mxIQX6xYlJa94dZD-4)

jespino:
#### Summary
This depends on changes that are going to be in the mattermost server

#### Ticket Link
Fixes #15

#### Screenshots

Before:

imagen

After:
imagen


crspeller (approved):
Looks great! Fixed the styling and changed a couple places to using styled components.

Will old off merging until the webapp PR is submitted so I can see it in action.

jespino:
@crspeller I've done a huge amount of changes, but now I think is in good shape to get this reviewed. Still there are things that I want to improve, but I think we can do it later.

webapp/src/components/ask_ai_input.tsx
```diff
@@ -0,0 +1,42 @@
...
+	padding: 5px
+	right: 7px
+`;
+
+const Input = styled.input`
+	border: none
+	backgroundColor: transparent
+`;
+
+export default function AskAiInput() {
```
crspeller, line 20: I am guessing this will be used in a follow up PR?

crspeller (commented):
Looks great! Just needs a few style fixes.

webapp/src/components/ask_ai_input.tsx
jespino: Yes

PR #38, opened 2023-07-18
Adding tool that can look up the most recent posts in a channel.

crspeller:
Also a fix to prevent double responces on mentions in DM channels.

### DIFF
--- a/server/ai/conversation.go
+++ b/server/ai/conversation.go
@@ -28,6 +28,7 @@ type ConversationContext struct {
 	CompanyName      string
 	RequestingUser   *model.User
 	Channel          *model.Channel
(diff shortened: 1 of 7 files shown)


tlinnyc:
How to get this to work? I'm not in the DM with the bot, I went into a public channel within a team, the bot is also added to both team and that specific channel, I asked it to retrieve the most recent 10 posts from "Channel Name", but it keeps telling me it doesn't have the ability to. I tried with OpenAI and Anthropic.

image

I know for summarizing a thread, I have to click the AI button on the thread, but not sure if there's something else I'm suppose to click for channel access.

Server is 9.11.8
Desktop: 5.10.2

PR #40, opened 2023-07-19
Update README with some new features.

crspeller:


azigler (approved):
Looks great to me @crspeller 

Do you want to add these to the OpenOps repo, as well?

crspeller:
@azigler Yes that makes sense.

PR #42, opened 2023-07-27
Azure OpenAI Bugfixes

bartoszpijet:
#### Summary
Azure integration never did work IMO. I've created bugfixes for docs + azure handler in openai compatible. Had to bump dependency to a version that supports new azure API version. I've also added fix in stream response checking, but I would love for someone to take look at openai.go line 143, I'm not sure if it won't cause memory leak, not an expert in go. Tested it and it looks good


jespino (approved):
This looks great, thanks! :confetti_ball:

crspeller (approved):
Awesome! Thanks for the fixes!

PR #48, opened 2023-08-04
Minor typo in responses

M-ZubairAhmed:
#### Summary
May this is need not changing because LLM will figure out what we mean by `responces` anyways


crspeller (approved):
Thanks!
It does seem to know what that means anyway, but better to have it correct.

PR #49, opened 2023-08-04
Adding whisper contect extraction

jespino:
#### Summary
Using whisper to extract audio/video files text to use in searches, and probably in AI prompts.


server/plugin.go
```diff
@@ -232,3 +233,15 @@ func (p *Plugin) MessageHasBeenPosted(c *plugin.Context, post *model.Post) {
 		return
 	}
 }
+
+func (p *Plugin) FileWillBeUploaded(c *plugin.Context, info *model.FileInfo, file io.Reader, output io.Writer) (*model.FileInfo, string) {
```
crspeller, line 270: We should add a configuration setting for this. 
It also needs to follow the security restrictions.

crspeller (changes requested):

server/plugin.go
streamer45: We may also want to devise an easy way to skip this in case of already transcribed files (e.g. calls transcriptions).

server/plugin.go
```diff
@@ -265,3 +266,33 @@ func (p *Plugin) handleAutoCallsRecording(post *model.Post, postingUser *model.U
 
 	return nil
 }
+
+func (p *Plugin) FileWillBeUploaded(c *plugin.Context, info *model.FileInfo, file io.Reader, output io.Writer) (*model.FileInfo, string) {
+	if info.Content != "" {
```
jespino, line 271: This address @streamer45's concern about not regenerating if the file info already have a content. But keep in mind that you have to generate the file info already with the `Content` field properly populated, if not, the plugin is going to try to generate it anyway.

server/plugin.go
```diff
@@ -265,3 +266,33 @@ func (p *Plugin) handleAutoCallsRecording(post *model.Post, postingUser *model.U
...
+		return nil, ""
+	}
+
+	if !p.getConfiguration().Config.EnableAudioContentExtraction {
```
jespino, line 275: This is the new setting requested by @crspeller

server/plugin.go
```diff
@@ -265,3 +266,33 @@ func (p *Plugin) handleAutoCallsRecording(post *model.Post, postingUser *model.U
...
+	if !strings.HasPrefix(info.MimeType, "audio/") && !strings.HasPrefix(info.MimeType, "video/") {
+		return nil, ""
+	}
+
+	channel, appErr := p.API.GetChannel(info.ChannelId)
+	if appErr != nil {
+		return nil, ""
+	}
+
+	if err := p.checkUsageRestrictions(info.CreatorId, channel); err != nil {
```
jespino: And here are the restriction checks, basically I'm checking the uploader restrictions and the channel restrictions there.

server/plugin.go
streamer45: I wish we had a `FileInfo.Type` field to handle custom files because as things stand, calls recordings won't ever have the `Content` set as the transcription process would happen independently, after they have been uploaded. So in the best case scenario we'd be patching the file info at a later time (not even sure we can do it without API additions), or use post props.

server/plugin.go
```diff
@@ -265,3 +266,28 @@ func (p *Plugin) handleAutoCallsRecording(post *model.Post, postingUser *model.U
...
+
+	if !p.getConfiguration().Config.EnableAudioContentExtraction {
+		return nil, ""
+	}
+
+	if !strings.HasPrefix(info.MimeType, "audio/") && !strings.HasPrefix(info.MimeType, "video/") {
+		return nil, ""
+	}
+
+	if err := p.checkUsageRestrictionsForUser(info.CreatorId); err != nil {
```
crspeller, line 283: Can we check channel restrictions here too? Or is it not assigned a channel yet when we get it here?

crspeller (changes requested):

server/plugin.go
jespino: Yes, exactly, You upload the file without the channel assigned yet, we start doing all the processing, and then we assign the post later, so we don't have enough information for the channel permission checking.

crspeller: Hmm. Unfortunate. We won't be able to enable this on community. I don't think we should add code that unexpectedly sends the contents of files in private channels or on teams where you have the plugin disabled. Can we disable this feature when channel/team restrictions are enabled? Also want to add a note in the system console about that behavior.

jespino: Very good point, I think we don't have a plugin hook that could help there. What we have is the possibility of being smarter and do something like "whenever you upload the plugin, put that in the queue of files to process, and every 5 seconds check for the file to have a channelID or not, if after 1 or 2 or X minutes the file doesn't have a channelId, discard the processing, if it has a channelId do the processing after checking the permissions.

That would be way easier of we have a "FileUpdated" hook. I can move in that direction too.

crspeller: Either one of those could be good. I can see a "FileUpdated" hook being generally useful.

crspeller:
Closing as stale. We might get back to this.

Issue #28, opened 2023-07-10
Support previewing jupyter notebooks within MM.


crspeller:
It would be great if MM could support preview for jupyter notebooks in file uploads. Maybe use somthing like `nbconvert` to display?

jespino:
I can take care of this using `notebook.js` that is javascript library that allows to preview ipynb files. Let me give it a try.

crspeller:
This is harder to make secure then it seems. Probably suitable for a separate plugin.

PR #50, opened 2023-08-10
Adding Jupyter notebook preview support

jespino:
#### Summary
This adds a lot of code, but we are taking most of the json data included in
the python notebook and putting it here. Also I'm including a lot of libraries
like react-markdown, dompurify and katex.

#### Ticket Link
#28

#### Screenshots
imagen


webapp/src/components/notebook_preview/notebook_preview.tsx
```diff
@@ -0,0 +1,44 @@
...
+    overflow: scroll;
+    max-width: 1024px;
+`;
+
+const NotebookPreview = ({fileInfo}: Props) => {
+    const [notebook, setNotebook] = useState<any>(null);
+
+    useEffect(() => {
+        const client = new Client4();
+        fetch(client.getFileUrl(fileInfo.id, new Date().getTime())).then(async (response: any) => {
```
crspeller: Isn't the second argument of getFileUrl supposed to be the update at of the fileinfo? We can also just leave it out to not break caching I think.

webapp/src/components/notebook_preview/notebook_preview.tsx
```diff
@@ -0,0 +1,44 @@
...
+            const data = await response.json();
+            setNotebook(data);
+        });
+    }, []);
```
crspeller: We should probably add `fileInfo.id` as a dependency here.

webapp/src/components/notebook_preview/code_cell_output_data.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import 'katex/dist/katex.min.css';
+
+type Props = {
+    format: string
+    content: string|string[]
+}
+
+const CodeCellOutputData = ({format, content}: Props) => {
+    const sanitizedContent = DOMPurify.sanitize(joinText(content));
```
crspeller, line 18: @jupenur Do you have some time to make sure we are not missing an injection attack here?

crspeller (changes requested):
This looks great! Tested it locally with some of my notebooks and seen no issue.

webapp/src/components/notebook_preview/notebook_preview.tsx
jespino: Fixed

webapp/src/components/notebook_preview/notebook_preview.tsx
jespino: Fixed

crspeller:
Too many possible security issues for now. Closing as stale.

Issue #30, opened 2023-07-10
Design Needed: Better system console UI


crspeller:
The current system console UI for configuration is a bit confusing. Replace with a custom component that hides fields not applicable to the currently selected LLMs would be a good start.

PR #51, opened 2023-08-11
Adding a custom configuration to the plugin in the admin console

jespino:
#### Summary
The new configuration settings in the admin console are fully customizable
through our components on the plugin side.

Also, we include the possibility of having any number of AI backends used for different things.

#### Ticket Link
#30 

#### Screenshots
imagen


crspeller (approved):
This is great! I merged my test changes. Slightly annoying we need to have a separate config struct, but it's fine, I just changed it to an embed so its easier. 

We will need to update the OpenOps demo to use this new config. Created https://github.com/mattermost/openops/issues/18 for that.

I also think we might want to change to looking up the LLM services by server side IDs rather than names to make things slightly more user friendly.

PR #71, opened 2023-09-08
Adding files to context of conversations

### LINKED ISSUE #61: Add ability for LLM to use files from content extraction.
Currently the LLM does not have access to file contents even though the MM server adds the extracted content. 

This is not straightforward as you need to balance what the LLM will pay attention too and the LLM context limit. Many files will not fit in the context.

crspeller:
Basic addition of file contents to conversations. 

Fixes #61 

## TODO 
- [ ] Work with token limits. How do we handle very large files?


crspeller:
Closing as stale for now. Needs work on token limits.

PR #83, opened 2023-10-17
Adding the new RHS panel

jespino:
#### Summary
This PR adds the new RHS panel, it includes backward compatibility support based on the existence (or not) of the `AdvancedCreateComment` component exposed in the mattermost plugin api. If that component exists all the new RHS is loaded and used, if not it is going to use the old mechanism to show the AI messages.

#### Screenshots


webapp/src/components/rhs/rhs.tsx
```diff
@@ -0,0 +1,99 @@
...
+import {makeGetPostsInChannel, getPost} from 'mattermost-redux/selectors/entities/posts';
+import {getPosts} from 'mattermost-redux/actions/posts';
+
+import {manifest} from '@/manifest';
+
+import ThreadItem from './thread_item';
+import RHSHeader from './rhs_header';
+import RHSNewTab from './rhs_new_tab';
+
+const ThreadViewer = styled((window as any).Components.ThreadViewer)`
```
crspeller: This crashes if the component is not available.

webapp/src/components/rhs/rhs.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const botChannelId = useSelector((state: any) => state['plugins-' + manifest.id].botChannelId);
+    const getPostsInChannel = makeGetPostsInChannel();
+    let posts = useSelector((state) => getPostsInChannel(state as any, botChannelId || '', -1)) || [];
+    posts = posts.sort((a, b) => b.update_at - a.update_at);
+
+    const selectedPostId = useSelector((state: any) => state['plugins-' + manifest.id].selectedPostId);
+
+    useEffect(() => {
+        if (currentTab === 'threads') {
+            dispatch(getPosts(botChannelId || '', 0, 60, false, true, true) as any);
```
crspeller: If CRT is not enabled this fails to get the root posts. 
This needs paging. 
If the bot channel ID is not available we should probably skip instead of fetching with an empty string. Also botChannelId is missing from the useEffect dependency list

webapp/src/index.tsx
```diff
@@ -11,20 +12,95 @@ import EditorMenu from './components/editor_menu';
...
+                store.dispatch({type: 'SET_AI_BOT_CHANNEL', botChannelId} as any);
+            });
+        }
+
+        store.subscribe(() => {
+            const state = store.getState();
+            if (state && state.entities.users.currentUserId !== currentUserId) {
+                currentUserId = state.entities.users.currentUserId;
+                if (currentUserId) {
+                    getAIDirectChannel(currentUserId).then((botChannelId) => {
```
crspeller, line 98: Not sure about this janky client side solution to getting the channel name so we can get the posts later. Maybe it would be better if we created an API endpoint on the plugin side so we can just ask for what to display. Then we can fetch the posts easily server side.

Also allows for possible integrations with external systems.

webapp/src/components/rhs/rhs.tsx
```diff
@@ -0,0 +1,99 @@
...
+export default function RHS() {
+    const dispatch = useDispatch();
+    const [currentTab, setCurrentTab] = useState('new');
+
+    const botChannelId = useSelector((state: any) => state['plugins-' + manifest.id].botChannelId);
+    const getPostsInChannel = makeGetPostsInChannel();
+    let posts = useSelector((state) => getPostsInChannel(state as any, botChannelId || '', -1)) || [];
+    posts = posts.sort((a, b) => b.update_at - a.update_at);
+
+    const selectedPostId = useSelector((state: any) => state['plugins-' + manifest.id].selectedPostId);
```
crspeller, line 38: There doesn't seem to be a need for this to go though redux. Maybe `useState` would be fine here?

webapp/src/components/rhs/rhs_new_tab.tsx
```diff
@@ -0,0 +1,119 @@
...
+`;
+
+type Props = {
+    botChannelId: string
+    selectPost: (postId: string) => void
+    setCurrentTab: (tab: string) => void
+}
+
+const setEditorText = (text: string) => {
+    const replyBox = document.getElementById('reply_textbox');
```
crspeller, line 67: I guess this works, but maybe we should have an official way to do this?

webapp/src/components/rhs/rhs_new_tab.tsx
```diff
@@ -0,0 +1,119 @@
...
+    PlaylistCheckIcon,
+} from '@mattermost/compass-icons/components';
+
+import {createPostImmediately} from 'mattermost-redux/actions/posts';
+
+import RHSImage from '../assets/rhs_image';
+
+import {Button} from './common';
+
+const CreatePost = styled((window as any).Components.CreatePost)`
```
crspeller: Again this will break if the component is not available.

webapp/src/components/rhs/rhs_new_tab.tsx
```diff
@@ -0,0 +1,119 @@
...
+                <OptionButton onClick={addMeetingAgenda}><FormatListNumberedIcon/>{'Meeting agenda'}</OptionButton>
+                <OptionButton onClick={addToDoList}><PlaylistCheckIcon/>{'To-do list'}</OptionButton>
+                <OptionButton onClick={addProsAndCons}><PlusMinus className='icon'>{'±'}</PlusMinus>{'Pros and Cons'}</OptionButton>
+            </QuestionOptions>
+            <CreatePost
+                placeholder={'Ask AI Assistant anything...'}
+                onSubmit={async (p: any) => {
+                    p.channel_id = botChannelId || '';
+                    p.props = {};
+                    const data = await dispatch(createPostImmediately(p) as any);
```
crspeller: Why as any?

webapp/src/components/rhs/rhs_new_tab.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    selectPost(data.data.id);
+                    setCurrentTab('thread');
+                }}
+            />
+        </NewQuestion>
+    );
+};
+
+export default React.memo(RHSNewTab);
```
crspeller, line 117: Why memo? It won't work anyway because we pass functions that are not wrapped with useCallback.

webapp/src/components/rhs/thread_item.tsx
```diff
@@ -0,0 +1,67 @@
...
+    repliesCount: number;
+    lastActivityDate: number;
+    onClick: () => void;
+}
+
+export default function ThreadItem(props: Props) {
+    return (
+        <ThreadItemContainer onClick={props.onClick}>
+            <Title>
+                <div className='title-text'>{props.postMessage}</div>
```
crspeller: Use a styled component here and not a className

crspeller (changes requested):

webapp/src/components/rhs/rhs.tsx
jespino: The lack of paging here, for now, it is intentional to not complicate this too much. It will get the last 60 conversations.

webapp/src/index.tsx
jespino: Good point, let me explore that.

webapp/src/components/rhs/rhs.tsx
jespino: Looks like it if it is not enable it gets root and not-root posts, what is not great, but it works, I just need to filter out not-root posts, and also it reduces the amount of posts that we are going to be able to get upfront without paging.

webapp/src/index.tsx
jespino: This can solve multiple problems here, like the fact that we have to relay on CRT or not CRT enable to be able to get a certain amount of root posts, and after that, that can lead to more complex pagination. But having something like "AI Conversations" endpoint would help. I don't see the need of doing it right away, because this PR already have enough complexity around UI, but I would love to have that new endpoint in another PR later.

webapp/src/components/rhs/rhs.tsx
jespino: The reason for this is because in different places we can select an AI post (using the `selectPost` action), that is not easy accessible from that places, like the post menu. I think this requires to be at redux or any other kind of global state to be able to be modified from other places.

webapp/src/components/rhs/rhs_new_tab.tsx
jespino: This was the less intrusive way that I found to do it. I think would be great to have an official way to do it, I don't know if this PR should solve that. Again this PR is already complex enough, we can polish things later, but I don't think we should polish everything here.

webapp/src/components/rhs/rhs_new_tab.tsx
jespino: For whatever reason that wasn't liking the `createPostImmediatly` as an `AnyAction` from redux, but it is an action.

webapp/src/components/rhs/rhs_new_tab.tsx
jespino: I think the solution is to use the useCallback outside, but you are right, the memo made no sense before, I'm adding the useCallback outside where it is needed.

webapp/src/components/rhs/thread_item.tsx
jespino: Good point

webapp/src/components/rhs/rhs.tsx
jespino: Fixed by only generating the styled component if the underneath component exists. That is ok, because we only show the RHS if the components exists in first place, so it shouldn't fail in runtime.

webapp/src/components/rhs/rhs.tsx
jespino: I checked and now it works for both, CRT enabled and CRT disabled, but in CRT disabled the list of threads is going to be less. We can add pagination later, or the dedicated API approach that you proposed in another comment.

webapp/src/components/rhs/rhs_new_tab.tsx
jespino: Fixed.

webapp/src/components/rhs/rhs_new_tab.tsx
jespino: Done

webapp/src/components/rhs/thread_item.tsx
jespino: Done.

crspeller:
@jespino Working on a few improvements myself before we merge this.

crspeller (approved):
Merging unrefined RHS to be improved.

azigler:
Love the new RHS. Was this added by this PR? Are there accompanying examples for customizing these prompts? I could see it be really appealing to config these per deployment.

image

crspeller:
@azigler This was the first PR yes. No, there is no support for customizing those.

PR #94, opened 2023-11-27
Fix max token func

nosyn:
#### Summary

Fix function typo which causes broken build
https://github.com/mattermost/mattermost-plugin-ai/pull/93


crspeller (approved):
Merge conflict. Thanks!

PR #108, opened 2024-01-17
Introduce source available folder. Move multiple LLMs to enterprise.

crspeller:
Adds a source available license under and `/enterprise` folder. Copied from https://github.com/mattermost/mattermost-plugin-playbooks/tree/master/server/enterprise

Moves the ability to have multiple LLMs configured to enterprise. The intention is to make this move useful in the future.


server/plugin.go
```diff
@@ -108,12 +113,18 @@ func (p *Plugin) getLLM() ai.LanguageModel {
...
+
+	if p.licenseChecker.IsMultiLLMLicensed() {
+		for _, service := range cfg.Services {
+			if service.Name == cfg.LLMGenerator {
+				llmService = service
+				break
+			}
 		}
+	} else {
+		llmService = cfg.Services[0]
```
lieut-data, line 130: `cfg.Services` is always non-empty?

lieut-data (approved):

PR #130, opened 2024-02-01
Restrict external data access

crspeller:
Restricts external data access to only when the AI bot is responding to a DM.
External data access is:
- Fetching posts from out of channel.
- Fetching data from GitHub
- Fetching user profile data.

The only exception is user profile data is fetched when SAML or LDAP is enabled for the fetched user and all fields are controlled by LDAP or SAML.

### DIFF
--- a/server/ai/conversation.go
+++ b/server/ai/conversation.go
@@ -24,6 +24,7 @@ type Post struct {
 }
 
 type ConversationContext struct {
+	BotID            string
 	Time             string
 	ServerName       string
 	CompanyName      string
@@ -34,7 +35,7 @@ type ConversationContext struct {
 	PromptParameters map[string]string
 }
 
-func NewConversationContext(requestingUser *model.User, channel *model.Channel, post *model.Post) ConversationContext {
+func NewConversationContext(botID string, requestingUser *model.User, channel *model.Channel, post *model.Post) ConversationContext {
 	// Get current time and date formatted nicely with the user's locale
 	now := time.Now()
 	nowString := now.Format(time.RFC1123)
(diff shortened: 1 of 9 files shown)


server/ai/conversation.go
```diff
@@ -24,6 +24,7 @@ type Post struct {
 }
 
 type ConversationContext struct {
+	BotID            string
```
jupenur, line 28: You never assign this, meaning `IsDMWithBot` returns true for *all* DM channels. Please add unit tests and return false from `IsDMWithBot` when the bot ID is empty.

server/built_in_tools.go
```diff
@@ -38,6 +38,13 @@ func (p *Plugin) toolResolveLookupMattermostUser(context ai.ConversationContext,
 		return "", errors.Wrap(err, "failed to lookup user")
 	}
 
+	// If this isn't a DM fail for users who are not SSO users.
+	if !context.IsDMWithBot() {
+		if user.AuthService != model.UserAuthServiceSaml && user.AuthService != model.UserAuthServiceLdap {
```
jupenur: At least with LDAP it looks like switching from email auth does not immediately update the profile, meaning you could have `user.AuthService == model.UserAuthServiceLdap` and user attributes that were manually set while on email auth. I *think* that's the same for SAML too, but I need to confirm.

Anyway, this check isn't good enough.

jupenur (changes requested):

server/ai/conversation.go
crspeller: Good catch. Updated to prevent this.

server/built_in_tools.go
crspeller: Don't see a way to check for this case. Removing the exception for now.

jupenur (approved):
Other than the issue concerning DMs I shared in-thread, LGTM.

PR #138, opened 2024-02-16
Remove hashtags and latex.

crspeller:
## Description

Removes hashtags and latex from AI posts.


jupenur (approved):
LGTM, but this will need to be implemented on mobile too, right?

crspeller:
@jupenur Yes. Mobile PR has been updated.

PR #175, opened 2024-05-03
Adding abiliy to add multiple AI bots.

crspeller:
Adds the ability to configure multiple AI bots. Each bot can be configured with a different backed and custom instructions.

### DIFF
--- a/webapp/src/components/system_console/enterprise_chip.tsx
+++ b/webapp/src/components/system_console/enterprise_chip.tsx
@@ -10,14 +10,15 @@ const Chip = styled.div`
     align-items: center;
     padding: 3px 8px 3px 22px;
     margin-left: 8px;
-    background: var(--button-bg);
     border-radius: 10px;
     height: 20px;
 
     font-size: 10px;
     font-weight: 600;
     line-height: 15px;
-    color: var(--center-channel-bg);
+
+    color: var(--button-bg);
+    background: rgba(var(--button-bg-rgb), 0.12);
 
     &:before {
         left: 7px;
--- a/webapp/src/components/system_console/enterprise_chip.tsx
+++ b/webapp/src/components/system_console/enterprise_chip.tsx
@@ -47,6 +47,7 @@ const SubText = styled.div`
 
 type Props = {
     subtext?: string;
+    text?: string;
 };
 
 const EnterpriseChip = (props: Props) => {
@@ -61,7 +62,7 @@ const EnterpriseChip = (props: Props) => {
             }
         >
             <Chip>
-                {'Enterprise'}
+                {props.text || 'Enterprise'}
             </Chip>
         </OverlayTrigger>
     );
--- a/webapp/src/components/system_console/no_bots_page.tsx
+++ b/webapp/src/components/system_console/no_bots_page.tsx
@@ -36,7 +36,7 @@ const StyledPanelContainer = styled(PanelContainer)`
 	flex-direction: column;
 	align-items: center;
 	gap: 16px;
-	padding-bottom: 32px;
+	padding-bottom: 56px;
 `;
 
 const Title = styled.div`
--- a/webapp/src/components/system_console/no_bots_page.tsx
+++ b/webapp/src/components/system_console/no_bots_page.tsx
@@ -1,0 +1,55 @@
+import {PlusIcon} from '@mattermost/compass-icons/components';
+import React from 'react';
+import styled from 'styled-components';
+
+import {PrimaryButton} from 'src/components/assets/buttons';
+import SparklesGraphic from 'src/components/assets/sparkles_graphic';
+
+import {PanelContainer} from './panel';
+
+type Props = {
+    onAddBotPressed: () => void;
+};
+
+const NoBotsPage = (props: Props) => {
+    return (
+        <StyledPanelContainer>
+            <SparklesGraphic/>
+            <Title>{'No AI bots added yet'}</Title>
+            <Subtitle>{'To get started with AI copilot, add an AI bot'}</Subtitle>
+            <PrimaryButton onClick={props.onAddBotPressed}>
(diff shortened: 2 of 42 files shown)


matthewbirtch (changes requested):
Awesome work on this @crspeller. A few things I noticed in my review on corpus:

1. The blinking cursor often is showing in the wrong place - in the middle of the streaming post rather than the end. I don't think this is related to this PR necessarily though. Do we have a ticket open for fixing this already?

2. Can we update the UI for the bot selection on the channel and thread summarization menus? This is my bad for not making it more consistent with the bot selector on the RHS. Rather than showing the chevron icon to the right of the bot label, I'd like to incorporate the chevron in to the bot label
image

3. For the default selected bot, in the RHS, it doesn't seem to remember what your last selected bot was. It always defaults to `AI Copilot`. Also, there seems to be inconsistent behavior between the channel summarization (which remembers the last-selected bot), and the thread summarization (which always defaults to AI Copilot). My instinct is to always remember the last-used bot.

4. I just noticed in the thread summarization menu, when you open it and then hover off, it closes. This is not consistent with the other menus on posts. It should hide when you click away or make a selection from the menu, not when you hover off. Again, not sure if this is related to this PR.

5. Opened a [new ticket](https://github.com/mattermost/mattermost-plugin-ai/issues/177) for a new problem I discovered while testing this: When in the global `Threads` view and then open the AI RHS, if I click th 'Brainstorm ideas' button, it populates the Thread input rather than the AI RHS input. 

6. In the system console, for the expand/collapse header of each bot, can we change the cursor to the pointer cursor when hovering?

7. In the system console, I created a new bot without the complete information (API key field blank). It let me save it and then the new bot showed up in the UI on the chat facing side. It initially shows 'Missing Information', but all it seemed to require was the Display name, username and avatar. We should have checks in place to not show an incorrectly configured bot.

8. I created a bot with the same username as another existing bot and didn't get any error. When I switched back to the chat-facing side, in the bot list, it replaced the old bot with the same username with the new one I created and no longer showed the old bot in the list. We will need to check for unique usernames.

crspeller:
1. Opened a ticket here: https://github.com/mattermost/mattermost-plugin-ai/issues/181
2. Makes sense. Updated. 
3. Updated as we discussed. Will now remember your preferences across sections, but not update RHS immediately.
4. Not related, bug in the webapp. https://github.com/mattermost/mattermost-plugin-ai/issues/183
5. Strange! Thanks. 
6. Updated
7+8. Added some validation for these cases. Bots without full configuration will not be created or updated. Duplicates will cause errors until duplication is resolved. (The UX is kinda bad around the duplication case, but it's a little complicated to do correctly because of how the system console works with plugins)

matthewbirtch (changes requested):
Thanks for resolving those items. A few more things I noticed:

1. The dropdown component seems to cut off the bottom of text. Notice the bottom of the `g` is getting cutoff in `Ask Sage`
image

2. When I click 'add an AI bot', I wonder if we should consider having the section expanded right away. It's adding another click to expand it which seems unncessesary right now. 

3. Also, it feels odd to call out immediately that there is missing information before I've even started entered anything. Could we only do this if I've clicked 'save' with missing information? And then are we able to call out the fields that are missing information at that point?
    image

4. This is odd, but when I create a new bot and start typing in the characters `wiz` the bot icon automatically changes on its own somehow.
    https://github.com/mattermost/mattermost-plugin-ai/assets/2040554/8608e446-5238-41da-9dcf-17eb5dfc85c8

5. I wasn't able to test to see if the empty state for bots was added because I didn't want to blow away the bots from corpus. Could you just share a screenshot of that?

6. I also wasn't able to see what it looks like if I don't have an enterprise license. Could you share a screenshot for that as well?

7. Are we going to implement a confirmation when deleting a bot? We had this in the designs, but I don't see it yet here in the latest.
image

8. In the chat history view, can adjust `x replies` so when there is only 1 it says ` reply`. Maybe you already have a ticket for this?
    image

crspeller:
1. Only happens on Chrome apparently. Fixed by just making the text field slightly bigger. 
2+3. We can make an improvement ticket for those. 
4. Haha fixed.
5. 
image
You can just delete all the bots on corpus, just don't press the save button.
6. 
image
7. It feels like it might be a bit redundant given the requirement to press save. 
8. Fixed

webapp/src/components/system_console/no_bots_page.tsx
```diff
@@ -0,0 +1,55 @@
...
+	width: 18px;
+	height: 18px;
+`;
+
+const StyledPanelContainer = styled(PanelContainer)`
+	display: flex;
+	flex-direction: column;
+	align-items: center;
+	gap: 16px;
+	padding-bottom: 32px;
```
matthewbirtch: Assuming this is only used for the empty state, can we make it `padding-bottom: 56px` to center things a bit better in the empty state?

crspeller: This is the result of that change: 
image

webapp/src/components/system_console/enterprise_chip.tsx
matthewbirtch, line 1: Can we update this chip with the following styles?

```
color: var(--button-bg);
background: rgba(var(--button-bg-rgb), 0.12);
```
image

matthewbirtch (approved):
Looks great @crspeller. Just a few minor tweaks noted, but I think we can make any other improvements in follow-ups

webapp/src/components/system_console/no_bots_page.tsx
matthewbirtch: Looks good. Thanks!

webapp/src/components/system_console/enterprise_chip.tsx
crspeller: image
Result.

Issue #176, opened 2024-05-04
🐛 bug: timeout streaming error when used with LocalAI


lhchew:
## Description

Encountered "timeout streaming" when configured to point to LocalAI.  Error is shown below.  Works fine when configured to point to OpenAI.

{
  "caller": "app/plugin_api.go:1003",
  "error": "timeout streaming",
  "level": "error",
  "msg": "Streaming result to post failed",
  "plugin_id": "mattermost-ai",
  "timestamp": "2024-05-04 05:45:23.445 Z"
}

## Steps to reproduce

Run Mattermost using Azure Container Instance.  Image used is mattermost/mattermost-preview.   
Mattermost server version is 9.7.3.
AI Copilot version is 0.6.3. 

Run LocalAI using Azure Container Instance.  Image used is localai/localai:latest-aio-cpu.

Able to get chat response from LocalAI using Postman with the following API Post call.

http://xx.xx.xx.xx:8080/v1/chat/completions

<Body>
{
  "model": "gpt-4",
   "messages": [
    {
        "role": "user",
        "content": "How are you?"
    }
    ]
}

kaakaa:
I encountered the same issue with ollama as well. I think it's because the default value for `StreamTimeoutSecond` is too short. Is it possible to configure this value?
https://github.com/mattermost/mattermost-plugin-ai/blob/2d13eb69da506ca83e1845bc6f309fc2b42b86f6/server/ai/openai/openai.go#L51

tjdalsry:
Me too.

How exactly do we solve this?

I know I'm lengthening that time, but each version has a different source, so I don't know what to change. I'm using the latest version of 0.8.

lhchew:
I refer to the configuration file of this plugin, there is a parameter to configure the streaming timeout in seconds.

**StreamingTimeoutSeconds int    `json:"streamingTimeoutSeconds"`**

https://github.com/mattermost/mattermost-plugin-ai/blob/master/server/ai/configuration.go

After you have enabled this plugin in MatterMost, open up MatterMost configuration file (config.json) and look for the section belonging to this plugin.  It should look something like this.

"service":{
    "apiKey":  ,
    "apiURL":  ,
    "defaultModel":  ,
    "id":  ,
    "orgId":  ,
    "password":  ,
    "tokenLimit":  ,
    "type":  ,
    "username":
}

[...]

PR #179, opened 2024-05-11
Make `StreamingTimeoutSecond` configurable

kaakaa:
# Refs #176

## Description
Make `StreamingTimeoutSecond` configurable for OpenAI and OpenAI Compatible. With setting it to `60`, I could fix the timeout issue with ollama refered in #176.

Screenshot from 2024-05-11 22-54-00


azigler:
Thank you for this PR @kaakaa! Pinging @crspeller for review.

crspeller:
@kaakaa Hey! Thanks for the contribution. Couple of changes to the system console went though. Do you mind fixing the conflicts? Let me know if you need any help.

kaakaa:
@crspeller Thanks for letting me know it. I've updated the code with the new system console.

**NOTE**: When deploying master branch to test my changes, the AI chat was not working due to the following error.

> {"timestamp":"2024-05-25 04:48:14.655 Z","level":"error","msg":"failed to get bot: ","caller":"app/plugin_api.go:1003","plugin_id":"mattermost-ai"}

The cause would be that `Config.DefaultBotName` is empty. Switching `Default bot` in sytem console would update `Config.DefaultBotName`, but I can't do it because I don't have Enterprise plan and can create only one bot. Updating and deleting existing bot didn't resolve this error. It seems that users who are not on the Enterprise Plan may encounter this issue.

By adding code before the following location to assign the default name ('ai') when defaultBotName is empty, I was able to temporarily work around the issue. However, this method requires changing the settings once, so it does not seem to be the best solution.
https://github.com/mattermost/mattermost-plugin-ai/blob/39e09762b8788622783b22efabb3e48e81cde05b/webapp/src/components/system_console/config.tsx#L156-L160

crspeller:
@kaakaa Thanks! 
I did noticed that issue while working on another pr and have a partial fix here: https://github.com/mattermost/mattermost-plugin-ai/pull/187/files#diff-efcd221be1d5948ef999c98624dc2d4a15dedc19b0945ca1b819a3d51480c607R142
But yes I need to fix the migration there.

tjdalsry:
When will this issue be patched?

PR #184, opened 2024-05-15
Adding support for zoom plugin summarization of chats and transcriptions

jespino:
Adding the summarization support for zoom plugin for transcriptions and chat history.


server/api_post.go
```diff
@@ -156,7 +156,7 @@ func (p *Plugin) handleSummarizeTranscription(c *gin.Context) {
 		c.AbortWithError(http.StatusInternalServerError, fmt.Errorf("unable to get calls user: %w", err))
 		return
 	}
-	if !targetPostUser.IsBot || targetPostUser.Username != CallsBotUsername {
+	if !targetPostUser.IsBot || (targetPostUser.Username != CallsBotUsername && targetPostUser.Username != ZoomBotUsername) {
 		c.AbortWithError(http.StatusBadRequest, errors.New("not a calls bot post"))
```
crspeller: add "or zoom"

server/meeting_summarization.go
```diff
@@ -161,9 +161,17 @@ func (p *Plugin) newCallTranscriptionSummaryThread(requestingUser *model.User, t
 			return fmt.Errorf("unable to read calls file: %w", err)
 		}
 
-		transcription, err := subtitles.NewSubtitlesFromVTT(transcriptionFileReader)
-		if err != nil {
-			return fmt.Errorf("unable to parse transcription file: %w", err)
+		var transcription *subtitles.Subtitles
+		if transcriptionFilePost.Type == "custom_zoom_chat" {
```
crspeller, line 168: This is kinda an awkward way to do this. It's not really a meeting we are summarizing here, just the chat from it. 
I wonder if we could just incorporate the chat into the main summary somehow?

server/ai/subtitles/subtitles.go
```diff
@@ -13,6 +14,38 @@ type Subtitles struct {
...
+		}
+		item.StartAt = startAt.Sub(zeroTime)
+		item.EndAt = startAt.Add(5 * time.Second).Sub(zeroTime)
+		item.Lines = append(item.Lines, astisub.Line{Items: []astisub.LineItem{{Text: text}}})
+		storage.Items = append(storage.Items, item)
+	}
+	return storage, nil
+}
+
+func NewSubtitlesFromChat(chat io.Reader) (*Subtitles, error) {
```
crspeller: Include zoom here because it's a zoom specific format. `NewSubtitlesFromZoomChat` same for above helper function.

crspeller (changes requested):

server/meeting_summarization.go
jespino: We could try that, but my concern is sometimes the chat deviates from the regular conversation and it is out-of-sync with the regular conversation. I think having a summary of both independently can work well. Also, the transcription and the chat are delivered in different webhooks, so it makes the things way more complicated. I would suggest to go with this approach and we can address that later and have another thing that is combining both. But I think for now this is good enough.

crspeller (approved):

PR #188, opened 2024-05-27
Implement conversation API endpoint

streamer45:
#### Description

PR implements a bots-only `/conversation` API endpoint to allow other plugins to programmatically start and continue conversations with Copilot without resorting to hacks such as posting in DMs between bots.

The general idea is to leverage the AI plugin as much as possible while eliminating the requirement of a real thread existing in a channel. 

The proposed handler expects a list of posts (i.e., a virtual thread) that will be fed as context. The last post in the list will be the current request.

#### Example

```sh
curl -v -H 'Authorization: Bearer BOT_AUTH_TOKEN' http://localhost:8065/plugins/mattermost-ai/conversation --data @payload.json
```

```json
{
    "bot_name": "ai",
    "thread": [
    {
        "id": "a",
        "message": "There's an incident ongoing",
        "channel_id": "oi18bftr738ntn35xnyjyed9uh",
        "user_id": "zjm6x7y4n3gjzfj7kw3xp9c9se",
        "create_at": 1
    },
    {
        "id": "b",
        "message": "A customer is experiencing db query timeouts",
        "channel_id": "oi18bftr738ntn35xnyjyed9uh",
        "user_id": "zjm6x7y4n3gjzfj7kw3xp9c9se",
        "create_at": 2
    }
    ],
    "request":     {
        "id": "c",
        "message": "can you summarize the current conversation?",
        "channel_id": "oi18bftr738ntn35xnyjyed9uh",
        "user_id": "zjm6x7y4n3gjzfj7kw3xp9c9se",
        "create_at": 3
    },
    "use_system_role": true
}
```


server/api_conversation.go
```diff
@@ -0,0 +1,118 @@
...
+		return
+	}
+
+	// Don't respond to ourselves
+	if p.IsAnyBot(post.UserId) {
+		c.AbortWithError(http.StatusBadRequest, errors.New("not responding to ourselves"))
+		return
+	}
+
+	bot := p.GetBotMentioned(post.Message)
```
streamer45: I am keeping the message requirements intact, such as needing to mention a specific bot.

server/api.go
```diff
@@ -26,6 +26,8 @@ func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Req
 	router.GET("/ai_threads", p.handleGetAIThreads)
 	router.GET("/ai_bots", p.handleGetAIBots)
 
+	router.POST("/conversation", p.handlePostConversation)
```
streamer45, line 30: I'm not overly worried about naming or scoping. Just started with this to get something working.

server/api_conversation.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	for {
+		select {
+		case msg := <-result.Stream:
+			if _, err := c.Writer.WriteString(msg); err != nil {
+				c.AbortWithError(http.StatusInternalServerError, fmt.Errorf("error while writing result: %w", err))
+			}
+			// Flushing lets us stream partial results without requiring the client to wait for the full response.
+			c.Writer.Flush()
```
streamer45, line 149: For streaming to work correctly we'll need https://github.com/mattermost/mattermost/commit/51f42717b62e3b98668b5ddc00059c59fa45ceb9.

server/api_conversation.go
```diff
@@ -0,0 +1,118 @@
...
+	// We only allow bots to use this API handler for the time being.
+	if _, err := p.pluginAPI.Bot.Get(userID, false); errors.Is(err, pluginapi.ErrNotFound) {
+		c.AbortWithError(http.StatusForbidden, errors.New("forbidden"))
+		return
+	} else if err != nil {
+		c.AbortWithError(http.StatusInternalServerError, fmt.Errorf("failed to get bot: %w", err))
+		return
+	}
+
+	var conv []*model.Post
```
streamer45: I thought a simple slice would be clean enough, but I'm happy to consider a more advanced approach, such as passing some sort of custom object that wraps the data we need.

streamer45:
@crspeller Not urgent. I'd love to get some feedback. This is the minimum addition I could think of that would meet the basic requirements. Happy to review any of it.

server/api_conversation.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/pluginapi"
+
+	"github.com/gin-gonic/gin"
+)
+
+func (p *Plugin) handlePostConversation(c *gin.Context) {
+	userID := c.GetHeader("Mattermost-User-Id")
+
+	// We only allow bots to use this API handler for the time being.
```
crspeller, line 31: If you are using the inter-plugin API to use this API we could restrict it to other plugins for now with the header: https://github.com/mattermost/mattermost-plugin-github/blob/5aa2450cc65254054872fc3ca6917b1c9354d543/server/plugin/api.go#L265

server/api_conversation.go
crspeller: Maybe we want to just wrap that in a struct so we can introduce other parameters if need be without a breaking change.

server/api_conversation.go
crspeller: Maybe we should take the bot name as a parameter instead of requiring a mention in the post?

crspeller (changes requested):
I suspect you will probably want a way to manipulate the system message using this API. Instructions to the bot are more powerful from the system message.

server/api_conversation.go
streamer45: Yeah, given https://github.com/mattermost/mattermost-plugin-ai/pull/188#discussion_r1619525511, I think it makes sense to make this a bit more structured. It may also be cleaner to decouple the current request/prompt from any past context.

server/api_conversation.go
streamer45: Yep, that's probably a healthier choice in the long run. Will update :+1:

server/api_conversation.go
streamer45: Thanks. I'll look into it although that would probably mean an additional roundtrip in my case (i.e. go through the plugin)  since the bot is just a client running outside of Mattermost.

streamer45:
> I suspect you will probably want a way to manipulate the system message using this API. Instructions to the bot are more powerful from the system message.

@crspeller Slowly getting back to this. Are you referring to the prompt used?

crspeller:
@streamer45 Yes having a way to specify the system prompt

server/api_conversation.go
```diff
@@ -0,0 +1,158 @@
...
+		return
+	}
+	prompt.AppendConversation(p.ThreadToBotConversation(bot, threadData.Posts))
+
+	// Overriding post role if requested.
+	if reqData.UseSystemRole {
+		for i := range prompt.Posts {
+			prompt.Posts[i].Role = ai.PostRoleSystem
+		}
+	}
```
streamer45, line 134: A bit brutal but it gets the job done. I can turn this into a method of `BotConversation` and refine further of course.

streamer45:
@crspeller Updated. Please give it another pass whenever you get a chance.

server/api_conversation.go
```diff
@@ -0,0 +1,158 @@
...
+
+	prompt, err := p.prompts.ChatCompletion(ai.PromptDirectMessageQuestion, p.MakeConversationContext(bot, postingUser, channel, post))
+	if err != nil {
+		c.AbortWithError(http.StatusInternalServerError, fmt.Errorf("failed to generate prompt: %w", err))
+		return
+	}
+	prompt.AppendConversation(p.ThreadToBotConversation(bot, threadData.Posts))
+
+	// Overriding post role if requested.
+	if reqData.UseSystemRole {
```
crspeller, line 130: Sorry I think I was unclear about what this means. The system prompt is special instructions to the LLM that are weighted more highly then user instructions. Usually there is only one system prompt. You can look at OpenAI's explanation of what it means here: https://platform.openai.com/docs/guides/text-generation/chat-completions-api

I guess what I really want is to be able to call this API from other plugins and be able to customize what prompt is provided on line 122. So that we can provide some custom instructions to the LLM about how the rest of the conversation should be handled like telling it it's in a call and talking live, the participants, etc.

crspeller (changes requested):

server/api_conversation.go
streamer45: @crspeller Are you thinking to go as far as letting plugins pass/override a whole prompt template or more like letting them choose which prompt to use and possibly augment existing ones?

crspeller: I am imagining them passing the whole prompt template. That way they can do whatever they want. Ideally they would be able to utilize the default personalities and whatnot as well.

streamer45:
I wish I had time to get back to this because I still think it's an interesting addition, but realistically, I have no time to look into it. Will close for now.

PR #194, opened 2024-05-30
Adding i18n support to the frontend

jespino:
Adding the i18n support to the fronted and adding the spanish translations

# Closes #


webapp/src/components/rhs/rhs_new_tab.tsx
```diff
@@ -91,41 +92,71 @@ const setEditorText = (text: string) => {
...
-
-const addProsAndCons = () => {
-    setEditorText('Write a pros and cons list about ');
-};
-
 const RHSNewTab = ({botChannelId, selectPost, setCurrentTab}: Props) => {
     const dispatch = useDispatch();
+    const intl = useIntl();
     const [draft, updateDraft] = useState<any>(null);
+    const addBrainstormingIdeas = useCallback(() => {
```
jespino, line 80: This changes here have certain implications, the main one is the fact that we are translating the buttons label, but also the generated text, that is because makes sense that if you click a button about "brainstorm ideas" in Spanish, the text suggested is also in Spanish, and the expectation is that the answer is going to be also in Spanish from the IA. If your IA doesn't support Spanish, it can be a problem.

webapp/src/components/bot_slector.tsx
```diff
@@ -27,7 +28,10 @@ export const DropdownBotSelector = (props: DropdownBotSelectorProps) => {
         >
             <>
                 <SelectMessage>
-                    {'Generate With:'}
+                    <FormattedMessage
+                        id='dropdown_bot_selector.generate_with'
```
crspeller: Do we need the id field? We didn't use it on Playbooks and never ran into any issues.

webapp/src/components/rhs/rhs_new_tab.tsx
crspeller: I think that makes sense. I think most LLMs are multi-lingual so most users won't run into this issue.

webapp/src/index.tsx
```diff
@@ -74,9 +77,33 @@ export default class Plugin {
...
+        const usersState = store.getState().entities.users;
+        const currentUser = usersState.profiles[usersState.currentUserId] || {};
+        const userLocale = currentUser.locale || 'en';
+
+        const translations: any = {
+            en: localeEn,
+            es: localeEs,
+        };
+
+        const withIntl = (Component: any) => {
```
crspeller: I think we can avoid this and register the translations with the webapp like playbooks does: https://github.com/mattermost/mattermost-plugin-playbooks/blob/be874bbd629b16379a13bafcb67ecdde4ff67608/webapp/src/index.tsx#L142

crspeller (changes requested):
This is great! Just a couple comments.

webapp/src/components/bot_slector.tsx
jespino: I think is not strictly needed, I can generate the sha of the text as IDs on the fliy, but it is consistent with mattermost-webapp. The only other drawback is the accidental collation of translations, sometimes, same text in different places have different translations.

webapp/src/index.tsx
jespino: Good point, I'll take a look.

jespino: Applied, it is cool. It is tricky because you have to tell webpack to use the shared `ReactIntl` for that, if not it is not finding the `IntlProvider`. But now it is way simpler, for sure.

webapp/src/components/bot_slector.tsx
jespino: I did the changes because I think it is also less collision prone with other services, and even to ensure that more, I'm using 8 characters ids, instead of 6, to avoid any accidental collition on translations between playbooks and copilot. Could be that something have the same text but totally different context, this would collide ever.

jprusch:
What would be the best way to add another language? Clone & create a PR? Hook the plugin up to Weblate?

jespino:
@jprusch the idea is to hook the plugin into Weblate as soon as this gets merged into the master branch.

crspeller (approved):

PR #195, opened 2024-05-30
Adding support for locale aware AI

### LINKED ISSUE #167: 💡 idea: Take into account the user locale for AI actions
## Description

Users having a non-english locale in their Mattermost profile profile would greatly benefit from having the answers in their chosen language.

AI Copilot responds in the language of the user input when using the RHS to chat, however action buttons (Summarize Thread, Open Questions, Actions Items etc.) in the UI will by default return a response in English from my testing.

These actions' answers being only available to the user triggering them, I would love to see them match the user locale, which is available to the AI Copilot already (checked by asking "who am I" to get my locale)

image

A valuable use case that I use often is conversations with our customers across multiple regions. Multiple languages are spoken in threads as there is often a language barrier. AI Copilot helps me get a summary in English of threads. Follow ups with AI copilot also help me draft responses in the appropriate language.

jespino:
This PR adds support for locale aware AI prompts. The AI will now be able to respond in the user's preferred language.

Closes #167


server/meeting_summarization.go
```diff
@@ -100,8 +101,9 @@ func (p *Plugin) createTranscription(recordingFileID string) (*subtitles.Subtitl
 
 func (p *Plugin) newCallRecordingThread(bot *Bot, requestingUser *model.User, recordingPost *model.Post, channel *model.Channel, fileID string) (*model.Post, error) {
 	siteURL := p.API.GetConfig().ServiceSettings.SiteURL
+	localizer := i18n.NewLocalizer(p.i18n, requestingUser.Locale)
 	surePost := &model.Post{
-		Message: fmt.Sprintf("Sure, I will summarize this recording: %s/_redirect/pl/%s\n", *siteURL, recordingPost.Id),
+		Message: fmt.Sprintf(localizer.MustLocalize(&i18n.LocalizeConfig{DefaultMessage: &i18n.Message{ID: "ai.summarize_recording", Other: "Sure, I will summarize this recording: %s/_redirect/pl/%s\n"}}), *siteURL, recordingPost.Id),
```
crspeller: I wonder if it would be better to avoid translation on the server here. We could instead send a prop or another post type that tells the client how to render the post. That way we could do all the translations on the client and the posts wouldn't get stuck in whatever language the user happened to be using at the time.

crspeller (changes requested):

server/meeting_summarization.go
jespino: I was thinking exactly the same with that. Anyway, about getting stuck in the language of the user, that is not that bad because the message is expected to be in the user language, so if the user change the language later, what you are going to get is an inconsistent header written on english, and a reply in Spanish (because that header looks like is in the message itself).

jespino: I just realized that there are a lot of other strings that I wasn't translating, probably 10 more, passing it identifiers and generating the post type, is fine, but can be a bit weird.

crspeller (approved):

PR #201, opened 2024-06-06
Adding telemetry

jespino:
## Description

This PR adds telemetry to the plugin. It sends a list of events, but stills needs to add the metrics for the current settings.


server/api_channel.go
```diff
@@ -103,6 +103,17 @@ func (p *Plugin) handleSince(c *gin.Context) {
 		return
 	}
 
+	p.track(evSummarizeUnreadMessages, map[string]interface{}{
+		"channel_id": channel.Id,
+		"user_id":    user.Id,
+		"since":      data.Since,
+		"type":       promptPreset,
+		"feature": map[string]string{
```
crspeller: Seems strange to add this here? The SKU is implied and all our events should be tagged with AI.

server/configuration.go
```diff
@@ -85,6 +85,15 @@ func (p *Plugin) setConfiguration(configuration *configuration) {
 
 // OnConfigurationChange is invoked when configuration changes may have been made.
 func (p *Plugin) OnConfigurationChange() error {
+	serverConfig := p.API.GetConfig()
```
crspeller, line 88: I think this can deadlock. If you move initTelemetry down below where EnsureBots is we are guaranteed to have the config.
You also have to stick it in OnActivate.

crspeller (changes requested):

server/api_channel.go
jespino: That comes from the ticket here: https://mattermost.atlassian.net/browse/MM-58196

jespino: @esethna What do you think?

server/configuration.go
jespino: This was replicated from calls code. And I think the OnConfigurationChange is called at least one on first load, so it should be fine in that sense.

server/api_channel.go
```diff
@@ -103,6 +103,17 @@ func (p *Plugin) handleSince(c *gin.Context) {
 		return
 	}
 
+	p.track(evSummarizeUnreadMessages, map[string]interface{}{
+		"channel_id": channel.Id,
+		"user_id":    user.Id,
```
ifoukarakis: Other parts of telemetry (in other plugins and MM code) use `user_actual_id` for identifying the user. It would be nice to stick to the convention. A more detailed list of conventions is avaiilable [here](https://mattermost.atlassian.net/wiki/spaces/DATAENG/pages/2766831623/Event+payload).

server/api_channel.go
jespino: From the jira ticked I think this is consider the right way to do it.

server/api_channel.go
jespino: Done! Thanks for the heads up.

server/api_channel.go
```diff
@@ -103,6 +103,18 @@ func (p *Plugin) handleSince(c *gin.Context) {
...
+	p.track(evSummarizeUnreadMessages, map[string]any{
+		"channel_id":     channel.Id,
+		"user_actual_id": user.Id,
+		"since":          data.Since,
+		"type":           promptPreset,
+		"feature": map[string]any{
+			"name":    "AI",
+			"is_free": false,
+			"skus":    []string{"enterprise"},
+		},
```
ifoukarakis: Features should be part of `context` rather than properties. Something similar is happening in [calls](https://github.com/mattermost/mattermost-plugin-calls/pull/798/files). Is it something that's possible with current plugin API?

server/api_channel.go
```diff
@@ -103,6 +103,18 @@ func (p *Plugin) handleSince(c *gin.Context) {
...
 	}
 
+	p.track(evSummarizeUnreadMessages, map[string]any{
+		"channel_id":     channel.Id,
+		"user_actual_id": user.Id,
+		"since":          data.Since,
+		"type":           promptPreset,
+		"feature": map[string]any{
+			"name":    "AI",
+			"is_free": false,
```
ifoukarakis: `is_free` is not needed anymore, as `skus` is used instead.

ifoukarakis (commented):
Left a couple of comments.

server/api_channel.go
```diff
@@ -103,6 +103,18 @@ func (p *Plugin) handleSince(c *gin.Context) {
 		return
 	}
 
+	p.track(evSummarizeUnreadMessages, map[string]any{
+		"channel_id":     channel.Id,
+		"user_actual_id": user.Id,
+		"since":          data.Since,
+		"type":           promptPreset,
+		"feature": map[string]any{
+			"name":    "AI",
```
ifoukarakis: Shouldn't this be `Copilot`?

jespino: good call

server/api_channel.go
jespino: Yes, I followed the calls approach and I think should work fine, actually, make the calls fro tracking events simpler.

jespino:
@ifoukarakis thanks, addressed your suggestions.

ifoukarakis (approved):
Approving after checking data on test schema in destination.

server/api_channel.go
```diff
@@ -103,6 +103,13 @@ func (p *Plugin) handleSince(c *gin.Context) {
 		return
 	}
 
+	p.track(evSummarizeUnreadMessages, map[string]any{
```
crspeller: Can we have it be `evUnreadMessages` since it does more than summary?

server/plugin.go
```diff
@@ -260,6 +277,20 @@ func (p *Plugin) handleDMs(bot *Bot, channel *model.Channel, postingUser *model.
...
 	}
 
+	if post.RootId == "" {
+		p.track(evUserStartedConversation, map[string]any{
+			"user_actual_id":   postingUser.Id,
+			"bot_id":           bot.mmBot.UserId,
+			"bot_service_type": bot.cfg.Service.Type,
+		})
+	} else {
+		p.track(evContextualInterrogation, map[string]any{
```
crspeller: Can we change this to `evContinuedConversation` as this is more than contextual interrogation.

server/api_channel.go
jespino: Done

server/plugin.go
jespino: Done

crspeller (approved):

PR #213, opened 2024-06-27
🐛 bug: typos in initial prompt templates

### LINKED ISSUE #212: 🐛 bug: typos in initial prompt templates
## Description

I really like how the prompts are split out into the new `.tmpl` files and I see a lot of potential for building mix-and-match prompts. I noticed a few typos and missing punctuation so I think these files need a review.

azigler:
# Closes #212

## Description

I cleaned up some typos and punctuation and standardized some of the ways we request things from the LLM.

I also think in `standard_personality_without_locale.tmpl` we should consider pulling the bot's name as a variable, instead of hardcoding it as "Copilot"

### DIFF
--- a/server/ai/prompts/emoji_select.tmpl
+++ b/server/ai/prompts/emoji_select.tmpl
@@ -1,5 +1,5 @@
 {{define "emoji_select.system"}}
-You are an emoji selector. You will receive a chat message. Determine which emoji from the following list is the best to react with. Do not answer questions. Do not respond with emoji. Respond only with one name of an emoji from the list:
+You are an emoji selector. You will receive a chat message and your job is to determine which emoji from the following list is the best reaction to the message. Do not answer questions or respond with emoji. Respond only with one name of an emoji from the following list:
 
 grinning
 smiley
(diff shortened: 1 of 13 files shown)


crspeller:
@azigler Merge conflicts from the other PR. 
Also wondering if you have done some testing to see if these new prompts are better? The prompts are surprisingly finicky.
I am weary of merging so many random fixes at once without having a good idea that it is going to improve things.

azigler:
@crspeller I agree, and wanted to get your opinion on how to best A/B test these. Since each model could handle each given prompt differently, I'm not sure what's the best benchmark to measure against. Should we test the prompts against a few models to get a sense of variety? I've seen a lot of things like [LLM "tournaments"](https://docs.titanml.co/blog/llm-arena-tournament/) but they compare two models using an identical prompt. In this case, we want to compare multiple models with multiple prompts, so it's more like a matrix.

crspeller:
@azigler Ideally yes, but we can probably start with just evaluating against one model. So far all my evaluations have been manual but it would be great if we could setup some framework like https://github.com/confident-ai/deepeval or something of the sort to test prompt changes. 

At the bare minimum we should manually test prompt changes to make sure they are improvements.

azigler:
Going to close this PR for now and revisit the ask for testing, since these prompts were updated in a parallel PR.

PR #218, opened 2024-07-05
Adding the cd workflow to generate releases

jespino:
## Description

This is a good practice to generate releases including environment variables like the RUDDER URL/KEY

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,19 @@
+name: cd
+on:
+  workflow_run:
+    workflows: ["ci"]
+    types:
+      - completed
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+permissions:
+  contents: read
+
+jobs:
+  plugin-cd:
+    uses: mattermost/actions-workflows/.github/workflows/plugin-cd.yml@main
+    with:
+      golang-version: "1.21"
+    secrets: inherit


crspeller:
@jespino Is there some context to this change? 
Asking for a review from @phoinixgrr since he has been working on the CI/CD for plugins.

jespino:
@crspeller the idea is to be able to generate release through the standard CI and automatically set the environment variables for the production rudder whenever the telemetry PR is merged.

jespino:
@phoinixgrr so... how do you recommend to include the RUDDER key for the AI plugin? i has been added to other plugins/projects using a CI secret. How would you apply that secret here?

phoinixgrr:
I would highly recommend against hardcoding the key directly into the code. Instead, consider injecting the key at runtime, although I am not sure if that's feasible in this situation. We need to understand the specifics. 

Configuration settings, including keys, should be stored in the environment(injected during runtime) rather than within the code. You can refer to [The Twelve-Factor App methodology](https://12factor.net/config) for more details on best practices for configuration management.

Moreover security wise, if the Key is burned into runtime, the key can be extracted with a tool like [Ghidra](https://ghidra-sre.org/) or `strings` linux command. 

Let's discuss this further, to find the best way to approach this problem. 

cc: @saturninoabril

jespino:
Following a different path for this

PR #221, opened 2024-07-11
Updating the rudder key using the plugins key environment variable

jespino:
## Description

This fixes the problem of setting up the rudder key in the plugin in the CI process


crspeller (approved):

ifoukarakis (commented):
This will lead to confusion between plugin data and copilot data.. A separate write key has been provisioned for AI plugin. Let's configure this properly. CC @jespino @crspeller

phoinixgrr:
> This will lead to confusion between plugin data and copilot data.. A separate write key has been provisioned for AI plugin. Let's configure this properly. CC @jespino @crspeller

Will tackle this

PR #222, opened 2024-07-15
[feat] updating the rudder key using the plugins key environment variable

phoinixgrr:
Reverting https://github.com/mattermost/mattermost-plugin-ai/pull/221
To be merged after: https://github.com/mattermost/delivery-platform/pull/156 

Ticket: https://mattermost.atlassian.net/browse/CLD-8054

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ INCLUDE_FFMPEG ?=
 BUILD_HASH = $(shell git rev-parse HEAD)
 LDFLAGS += -X "main.buildHash=$(BUILD_HASH)"
 LDFLAGS += -X "main.isDebug=$(MM_DEBUG)"
-LDFLAGS += -X "main.rudderWriteKey=$(MM_RUDDER_PLUGINS_PROD)"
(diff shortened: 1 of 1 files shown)


toninis (approved):

saturninoabril (approved):

ifoukarakis (commented):
nit: shouldn't we be using `copilot` rather than `plugin-ai` for naming?

phoinixgrr:
> nit: shouldn't we be using `copilot` rather than `plugin-ai` for naming?

@ifoukarakis  Refers to `mattermost-plugin-ai` plugin , should be easier to be correlated with the actual usage. 
We can change it if you think that is needed though. 

Let me know.

ifoukarakis:
> > nit: shouldn't we be using `copilot` rather than `plugin-ai` for naming?
> 
> @ifoukarakis Refers to `mattermost-plugin-ai` plugin , should be easier to be correlated with the actual usage. We can change it if you think that is needed though.
> 
> Let me know.

I'm ok with either. `mattermost-plugin-ai` is the name of the repo, but `README.md` mentions `Mattermost Copilot Plugin`, so I was thinking if there's an agreed naming so that we're consistent. Again, that's nitpicking, so not a blocker.

PR #225, opened 2024-07-29
Migrate Createpost to v10 version

crspeller:
## Description

See https://github.com/mattermost/mattermost/pull/27766 for server side fixes.

Compatability was broken by https://github.com/mattermost/mattermost/pull/26419


webapp/src/components/rhs/rhs_new_tab.tsx
```diff
@@ -92,6 +100,66 @@ const RHSNewTab = ({botChannelId, selectPost, setCurrentTab}: Props) => {
...
+    let editorComponent;
+    if (AdvanceTextEditor) {
+        editorComponent = (
+            <AdvanceTextEditor
+                data-testid='rhs-new-tab-create-post'
+                channelId={botChannelId}
+                placeholder={intl.formatMessage({defaultMessage: 'Ask Copilot anything...'})}
+                isThreadView={true}
+                afterSubmit={(result: {created?: {id: string}}) => {
+                    if (result.created?.id) {
```
hmhealey, line 111: I commented on https://github.com/mattermost/mattermost/pull/27766 how we need to re-add the optimistic/pending post behaviour which that accidentally removed. I suggested over there that we would wait to call `afterSubmit` until the post is actually created on the server which would mean that we won't have to wait for the pending post logic, so there might be nothing to do here, but I wanted to make sure to mention that.

If we were to call `afterSubmit` after the pending post was created, we could potentially select the pending post and then have the reducer that handles `SELECT_AI_POST` and handle a `RECEIVED_NEW_POST` to have it update the pending ID to the real ID, but that's likely overcomplicating things for now

hmhealey (approved):

hmhealey (approved):

PR #228, opened 2024-08-05
Add valiation to username for bot.

crspeller:


enzowritescode (changes requested):
Front-end validation is good because it can eliminate the number of API requests but it can always be bypassed. Validation also needs to be done on the back-end.

crspeller:
@enzowritescode Validation is already done on the back-end. It won't create an invalid bot.

enzowritescode (approved):

PR #229, opened 2024-08-05
Prevent bad query strings from LLM

crspeller:
## Description

It is possible for the LLM to produce a function call that would result in a bad query string. Proper URL Query escapting prevents this.

### DIFF
--- a/server/built_in_tools.go
+++ b/server/built_in_tools.go
@@ -166,7 +166,7 @@ func (p *Plugin) toolGetGithubIssue(context ai.ConversationContext, argsGetter a
 		return "invalid parameters to function", errors.New("invalid issue number")
 	}
 
-	req, err := http.NewRequest("GET",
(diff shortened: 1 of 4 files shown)


server/built_in_tools.go
```diff
@@ -165,7 +166,14 @@ func (p *Plugin) toolGetGithubIssue(context ai.ConversationContext, argsGetter a
 		return "invalid parameters to function", errors.New("invalid issue number")
 	}
 
-	req, err := http.NewRequest("GET", fmt.Sprintf("/github/api/v1/issue?owner=%s&repo=%s&number=%d", args.RepoOwner, args.RepoName, args.Number), nil)
+	req, err := http.NewRequest("GET",
```
enzowritescode: `"GET"` should be replaced with `http.MethodGet`

enzowritescode (changes requested):

enzowritescode (approved):

PR #230, opened 2024-08-07
Adapt to RHS scroll direction change

crspeller:
## Description

Adapt to PR in webapp changing RHS scroll direction: https://github.com/mattermost/mattermost/pull/27162

Also requires https://github.com/mattermost/mattermost/pull/27871 so scrolling doesn't break for generating posts.

### DIFF
--- a/webapp/src/components/llmbot_post.tsx
+++ b/webapp/src/components/llmbot_post.tsx
@@ -1,7 +1,7 @@
 import React, {MouseEvent, useEffect, useRef, useState} from 'react';
 import {FormattedMessage} from 'react-intl';
 import {useSelector} from 'react-redux';
-import styled, {css, createGlobalStyle} from 'styled-components';
+import styled from 'styled-components';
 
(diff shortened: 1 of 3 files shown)


webapp/src/components/llmbot_post.tsx
```diff
@@ -10,37 +10,15 @@ import {SendIcon} from '@mattermost/compass-icons/components';
...
-	::before {
-		content: '';
-		position: absolute;
-		width: 110%;
-		height: 110%;
-		left: 50%;
-		top: 50%;
-		transform: translate(-50%, -50%);
-	}`}
+const PostBody = styled.div`
```
jespino, line 21: 0/5 here, I see the value on giving a name to the div itself, but maybe, because it is no longer needed from the styles perspective we can remove this. Anyway, I'm ok leaving it as is, just to give it a name.

crspeller: The name is helpful for debugging and whatnot I think we can leave it.

PR #233, opened 2024-08-19
Anthropic request headers helper

enzowritescode:
## Description
- Add helper function to set common headers for Anthropic requests

### DIFF
--- a/server/ai/anthropic/client.go
+++ b/server/ai/anthropic/client.go
@@ -97,9 +97,7 @@ func (c *Client) MessageCompletionNoStream(completionRequest MessageRequest) (st
 		return "", fmt.Errorf("could not create request: %w", err)
 	}
 
-	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("X-API-Key", c.apiKey)
-	req.Header.Set("anthropic-version", "2023-06-01")
+	c.prepareRequestHeaders(req)
 
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
@@ -135,11 +133,9 @@ func (c *Client) MessageCompletion(completionRequest MessageRequest) (*ai.TextSt
 		return nil, err
 	}
 
-	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("X-API-Key", c.apiKey)
+	c.prepareRequestHeaders(req)
 	req.Header.Set("Accept", "text/event-stream")
 	req.Header.Set("Connection", "keep-alive")
-	req.Header.Set("anthropic-version", "2023-06-01")
 
 	output := make(chan string)
 	errChan := make(chan error)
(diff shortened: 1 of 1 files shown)


server/metrics.go
```diff
@@ -19,8 +19,7 @@ func (p *Plugin) GetMetrics() metrics.Metrics {
 }
 
 func (p *Plugin) metricsMiddleware(c *gin.Context) {
-	metrics := p.GetMetrics()
-	if metrics == nil {
+	if p.GetMetrics() == nil {
```
crspeller: Might be nicer to use the variable below instead of putting the `p.GetMetrics()` into the if

server/ai/anthropic/client.go
```diff
@@ -203,3 +199,9 @@ func (c *Client) MessageCompletion(completionRequest MessageRequest) (*ai.TextSt
 
 	return &ai.TextStreamResult{Stream: output, Err: errChan}, nil
 }
+
+func (c *Client) prepareRequestHeaders(req *http.Request) {
```
crspeller: Not sure this helper is necessary. But if we do want to keep it maybe call it `setAnthropicHeaders` and pull in the streaming ones with a boolean parameter.

crspeller (changes requested):

server/metrics.go
enzowritescode: The `metrics` variable was colliding with the metrics package and the variable was only being used for the if statement.

server/ai/anthropic/client.go
enzowritescode: That sounds good. Yeah it was repeated logic so I figured not repeating it would be better, especially because then there were less hard-coded strings and consolidating into one place would eliminate the argument for making constants to reduce the number of hard-coded strings.

server/metrics.go
enzowritescode: Are you saying like this? This is something I had actually wondered about since I saw `p.GetMetrics()` being called several times

```
func (p *Plugin) metricsMiddleware(c *gin.Context) {
	llmMetrics := p.GetMetrics()
	if llmMetrics == nil {
		c.Next()
		return
	}
	llmMetrics.IncrementHTTPRequests()
	now := time.Now()

	c.Next()

	elapsed := float64(time.Since(now)) / float64(time.Second)

	status := c.Writer.Status()

	if status < 200 || status > 299 {
		llmMetrics.IncrementHTTPErrors()
	}

	endpoint := c.HandlerName()
	llmMetrics.ObserveAPIEndpointDuration(endpoint, c.Request.Method, strconv.Itoa(status), elapsed)
}
```

crspeller: Yup like that.

server/ai/anthropic/client.go
enzowritescode: I just remembered my reasoning for originally calling it `prepareRequestHeaders()`.
using "set" typically denotes that you are setting an instance variable on the calling object, whereas in this case I am setting values of something being passed in.

crspeller (approved):

PR #237, opened 2024-08-23
Reduce excess notifications the user gets.

crspeller:
## Description

In a lot of cases the user will get exessive notifications rom the AI bot. The aim of this and the accompanying webapp PR is to reduce those instances.

1. Block ALL notifications client side from AI posts
2. Enable a flag in the thread viewer to automaticly mark AI threads as viewed.


wai52168:
[crspeller](https://github.com/crspeller)

First of all, thank you so much for your great work on this plugin!
I encountered a small issue: when using the Copilot, mobile devices logged in with the same account keep receiving repeated push notifications.
I hope this can be optimized soon. Thank you again for your efforts and for creating such a great project!

image

PR #247, opened 2024-10-03
Adding better access controls.

crspeller:
Adds access controls for LLM Bots. 

More details in the specification here: https://mattermost.atlassian.net/wiki/spaces/FF/pages/2917400646/UX+Spec+Access+Controls
Some elements of the spec have been differed from this PR.


webapp/src/components/system_console/llm_access.tsx
```diff
@@ -0,0 +1,161 @@
...
+	margin-top: 24px;
+	margin-bottom: 24px;
+	display: grid;
+	grid-template-columns: auto 1fr;
+	grid-row-gap: 8px;
+	grid-column-gap: 8px;
+	grid-align-items: center;
+`;
+
+const MainContainer = styled.div`
```
enzowritescode, line 21: Just so I understand, what is the purpose of calling this out as opposed to just using a div in the return statement?

jespino (approved):
LGTM

enzowritescode (approved):

webapp/src/components/system_console/llm_access.tsx
crspeller: It's a bit vestigial, there used to be styles there.

Issue #245, opened 2024-09-20
💡 idea: Support for reading attached files


AnnoyingTechnology:
## Description

It would be nice if the attached files to a message/prompt could be read by the model. 

Currently, only images seem to be exposed to the model/API.

se4n01:
I'm surprised this isn't done by default, chatgpt spews nonsense if you upload a file currently so I think this is a bug as well as a FR.

```
Se4n01
[8:23 AM]

can you transcribe the text in this text file?
test.txtTXT26B
Chat GPT
BOT
[8:23 AM]

I'm sorry, but I am unable to transcribe the text from an image file at the moment. If you can provide the text directly or in a different format, I'd be happy to help with the transcription.
```

AnnoyingTechnology:
@azigler

PR #253, opened 2024-10-14
Add files to LLM context.

crspeller:
# Closes: #245
# Closes: #61

## Description

Adds files to the context of the context of the LLM. Any file that is supported by the server text extraction (including PDFs) should work as long as that text extraction works. For some reason that extraction doesn't work on text files so the ability to extract text files was added here. 
Docs on Text extraction: https://docs.mattermost.com/collaborate/search-for-messages.html

Many files will be over the limit of your LLM or will reach the token limit set in the system console. Files that are too long will cause the prompt to be cut off at the beginning. Set the token limit of your LLM in the system console to make sure this works correctly.

Use a high token limit LLM to use this effectively, all but the latest (llama3.2) on-prem models will have issues with large files.

### DIFF
--- a/server/ai/configuration.go
+++ b/server/ai/configuration.go
@@ -21,6 +21,7 @@ type BotConfig struct {
 	Service            ServiceConfig `json:"service"`
 	EnableVision       bool          `json:"enableVision"`
 	DisableTools       bool          `json:"disableTools"`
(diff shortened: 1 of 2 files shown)


jespino (approved):
LGTM

jespino:
Now that we have this, would be great to expose something like "attachment summarization"

PR #259, opened 2024-11-05
Add withUserId configuration for OpenAI

### LINKED ISSUE #258: 💡 idea: Sending `user_id` as a parameter to the AI model
## Description

In the OpenAI API there is [a `user` parameter](https://platform.openai.com/docs/api-reference/chat#chat-create-user) when creating a chat. It would be great to send the Mattermost user_id as the user parameter from Mattermost. 

* Providing the `user_id` to AI model is a mandatory requirement in some industries.  
 
* Parameterizing `user_id` should be configurable from the System console.

hun-a:
# Closes #258 

## Description

Hello!

I'm thoroughly enjoying the Mattermost AI plugin and appreciate how useful it has been for our team. As I’ve been using it, I thought of a small enhancement that could add value, so I put together a PR to share with you.

The OpenAI Chat Completion API supports a [user parameter](https://platform.openai.com/docs/api-reference/chat/create#chat-create-user), which can help with user monitoring and detecting abuse. I’ve implemented a feature that allows admins to toggle whether or not to send the `Username` when using OpenAI models. This option is accessible in the plugin settings, providing greater flexibility and control.

Additionally, I have a suggestion that might improve the user experience for a more global audience. Currently, in direct messages (DMs), the plugin uses a combination of `Firstname` and `Lastname` to identify users.
https://github.com/mattermost/mattermost-plugin-ai/blob/master/server/ai/prompts/standard_personality_without_locale.tmpl#L9
```
{{if .RequestingUser.FirstName}}
Their full name is {{.RequestingUser.FirstName}} {{.RequestingUser.LastName}}.
{{end}}
```
Modifying this to pass the `Username` instead might feel more relatable for users from different cultural backgrounds, which could further support Mattermost’s positioning as a globally inclusive collaboration tool.

If you find the suggestion to use Username agreeable, I’d be happy to incorporate it into this PR.

Thank you for considering this enhancement!

### DIFF
--- a/server/ai/configuration.go
+++ b/server/ai/configuration.go
@@ -11,6 +11,7 @@ type ServiceConfig struct {
 	Password                string `json:"password"`
 	TokenLimit              int    `json:"tokenLimit"`
 	StreamingTimeoutSeconds int    `json:"streamingTimeoutSeconds"`
+	WithUserId              bool   `json:"withUserId"`
 }
 
 type BotConfig struct {
--- a/server/ai/configuration.go
+++ b/server/ai/configuration.go
@@ -11,7 +11,7 @@ type ServiceConfig struct {
 	Password                string `json:"password"`
 	TokenLimit              int    `json:"tokenLimit"`
 	StreamingTimeoutSeconds int    `json:"streamingTimeoutSeconds"`
-	WithUserId              bool   `json:"withUserId"`
+	SendUserID              bool   `json:"sendUserID"`
 }
 
 type BotConfig struct {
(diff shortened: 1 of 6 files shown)


crspeller (approved):
This is great! Thanks for the PR! Made a few tweaks and fixed the linting. 

For the username thing, it is actually supplied the the LLM in the previous line: https://github.com/mattermost/mattermost-plugin-ai/blob/master/server/ai/prompts/standard_personality_without_locale.tmpl#L8
Maybe we could use the Mattermost server setting around this to change based on what that setting is set to rather than supplying both? 
Is there a specific behavior you are setting the LLM exhibit that you thing this change will fix?

hun-a:
Hi @crspeller!

Thank you so much for fixing the linting issues and refining the code! 🙏

I initially suggested using `Username` instead because I was aiming for the LLM to prioritize the user's `FirstName` over their full username. However, I overlooked the fact that `Username` is already being supplied in the existing template. Given that, I agree that it makes more sense to stick with the current approach as you suggested.

Thanks again for your help! 😊

PR #269, opened 2024-12-18
Changed the `WebsocketListeners` in the webapp from a `Map` to an array.

### LINKED ISSUE #157: 🐛 bug: When replying in a thread, the response from OpenAI is not updated on the main screen in the stream.
## Description

In the Mattermost AI plugin, when replying in a thread, the right side thread side shows a stream of responses from OpenAI, but the main screen does not update

- Mattermost Version: 9.6.0
- Database Schema Version: 119
- Build Number: 8083694714
- Database: postgres
- AI Copilot: 0.6.2
   - Default Model: gpt-4-0125-preview
   - User restrictions (experimental): false
   - Debug: false

Mattermost_AI_Replies_Not_Updating_in_Main_Screen

## Steps to reproduce

1. Install AI Copilot Plugin
2. Setup AI Copilot with OpenAI API Key
3. Ask AI a question
4. Open thread in reply and ask AI a question

code1492:
# Closes #157

## Description

Hello!

We’ve recently adopted Mattermost, and I received a report about an issue similar to the one below, so I decided to address it:
https://github.com/mattermost/mattermost-plugin-ai/issues/157

The WebsocketListeners was defined as a Map, which seemed to prevent both the main screen and the RHS (right-hand side: thread view) from using WebsocketListeners simultaneously. I updated it by changing the Map to an array.

It appears to be working well, so I’ve submitted a pull request. If there are any issues, please feel free to leave a comment :)

### DIFF
--- a/webapp/src/components/llmbot_post.tsx
+++ b/webapp/src/components/llmbot_post.tsx
@@ -93,8 +93,8 @@ export interface PostUpdateWebsocketMessage {
 
 interface Props {
     post: any;
-    websocketRegister: (postID: string, handler: (msg: WebSocketMessage<PostUpdateWebsocketMessage>) => void) => void;
-    websocketUnregister: (postID: string) => void;
+    websocketRegister: (postID: string, listenerID: string, handler: (msg: WebSocketMessage<PostUpdateWebsocketMessage>) => void) => void;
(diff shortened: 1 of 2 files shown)


crspeller (approved):
Nice! This seems to work well. 
Thanks for the contribution!

PR #272, opened 2025-01-08
Move to official anthropic golang client. Add inital support for images and tool use. 

crspeller:
## Description

- Added anthropic-sdk-go dependency
- Removed custom implementation in favor of official SDK
- Added support for multi-modal conversations with images
- Implemented tool/function calling support with recursive resolution


jespino (approved):
LGTM

jespino:
I don't see anything outstanding to block this, but the linter is complaining.

PR #276, opened 2025-01-13
Make max_tokens configurable and remove pre-set values for top_p, temperature etc

BenCookie95:
## Description
Make max_tokens configurable and remove pre-set values for top_p, temperature etc. Top_p was using the LLM defaults anyway so there was no point in setting them explicitly.

Default values were added [here](https://github.com/mattermost/mattermost-plugin-ai/commit/2234a1d19af4bb95ced3347992ae1f3b43ca9cd0) if you're trying to remember a reason for it.


BenCookie95:
These e2e failures look a bit random, are these usually flaky?

server/ai/configuration.go
```diff
@@ -12,6 +12,7 @@ type ServiceConfig struct {
 	TokenLimit              int    `json:"tokenLimit"`
 	StreamingTimeoutSeconds int    `json:"streamingTimeoutSeconds"`
 	SendUserID              bool   `json:"sendUserID"`
+	MaxTokens               int    `json:"maxTokens"`
```
crspeller: Maybe we should make these names more clear. What is the difference between MaxTokens and TokenLimit? 

Maybe `InputTokenLimit` and `OutputTokenLimit`? What do you think? 
Then we can update everywhere else to be consistent with that wording.

webapp/src/components/system_console/bot.tsx
```diff
@@ -221,6 +230,16 @@ const ServiceItem = (props: ServiceItemProps) => {
     const intl = useIntl();
     const hasAPIKey = type !== 'asksage';
     const isOpenAIType = type === 'openai' || type === 'openaicompatible' || type === 'azure';
+
+    const getDefaultMaxTokens = useCallback(() => {
```
crspeller: Not sure we need to bother with useCallback here. I think it won't do anything here?

server/ai/anthropic/anthropic.go
```diff
@@ -146,10 +148,15 @@ func conversationToMessages(posts []ai.Post) (string, []anthropicSDK.MessagePara
...
 
 func (a *Anthropic) GetDefaultConfig() ai.LLMConfig {
-	return ai.LLMConfig{
-		Model:              a.defaultModel,
-		MaxGeneratedTokens: DefaultMaxTokens,
+	config := ai.LLMConfig{
+		Model: a.defaultModel,
 	}
+	if a.maxTokens == 0 {
+		config.MaxGeneratedTokens = DefaultMaxTokens
```
crspeller, line 155: Let's bump this default to `8192` since we are here. The newer claude models support that.

crspeller (changes requested):

crspeller:
@BenCookie95 If you merge master the e2e tests should pass.

server/ai/configuration.go
BenCookie95: Are you just wanting to rename the struct fields and leave the json field names the same? Renaming the JSON field for `tokenLimit` would require a migration.

I have to be honest, I'm having a hard time understanding what `TokenLimit` is but from what I can see is `TokenLimit` the `InputTokenLimit` and `MaxTokens` is the `OutputTokenLimit`? I do think both these fields need a description underneath them in system console

crspeller: @BenCookie95 yes let's rename the variables for clarity. We can leave the json for now so we don't have to do a migration. Add a comment explaining. 

Yes you have the mapping correct, the names are quite poor which is why I think we should fix them.

BenCookie95: Done, https://github.com/mattermost/mattermost-plugin-ai/pull/276/commits/061adc828924443f6dd2c2856215a568016cc284 for the name changes

crspeller (approved):

PR #278, opened 2025-01-16
Improved configuration validation

enzowritescode:
## Description

- Improve configuration validation based on aider feedback
- Add additional validation for asksage

### DIFF
--- a/server/ai/configuration.go
+++ b/server/ai/configuration.go
@@ -49,10 +49,28 @@ type BotConfig struct {
 }
 
 func (c *BotConfig) IsValid() bool {
-	isInvalid := c.Name == "" ||
-		c.DisplayName == "" ||
-		c.Service.Type == "" ||
-		((c.Service.Type == "openaicompatible" || c.Service.Type == "azure") && c.Service.APIURL == "") ||
-		(c.Service.Type != "asksage" && c.Service.Type != "openaicompatible" && c.Service.Type != "azure" && c.Service.APIKey == "")
-	return !isInvalid
+	// Basic validation
+	if c.Name == "" || c.DisplayName == "" || c.Service.Type == "" {
+		return false
+	}
+
+	// Validate access levels are within bounds
+	if c.ChannelAccessLevel < ChannelAccessLevelAll || c.ChannelAccessLevel > ChannelAccessLevelNone {
+		return false
+	}
+	if c.UserAccessLevel < UserAccessLevelAll || c.UserAccessLevel > UserAccessLevelNone {
+		return false
+	}
+
+	// Service-specific validation
+	switch c.Service.Type {
+	case "openai", "anthropic":
+		return c.Service.APIKey != ""
+	case "openaicompatible", "azure":
+		return c.Service.APIURL != ""
+	case "asksage":
+		return c.Service.Username != "" && c.Service.Password != ""
+	default:
+		return false
+	}
 }


server/ai/configuration_test.go
```diff
@@ -0,0 +1,262 @@
+package ai
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBotConfig_IsValid(t *testing.T) {
```
crspeller, line 9: There isn't a success case. So a function that always fails will pass. (Which is the current case)

server/ai/configuration.go
```diff
@@ -49,10 +57,30 @@ type BotConfig struct {
...
-		((c.Service.Type == "openaicompatible" || c.Service.Type == "azure") && c.Service.APIURL == "") ||
-		(c.Service.Type != "asksage" && c.Service.Type != "openaicompatible" && c.Service.Type != "azure" && c.Service.APIKey == "")
-	return !isInvalid
+	// Basic validation
+	if c.Name == "" || c.DisplayName == "" || c.Service.Type == "" {
+		return false
+	}
+
+	// Validate access levels are within bounds
+	if c.ChannelAccessLevel < ChannelAccessLevelAll || c.ChannelAccessLevel > ChannelAccessLevelNone {
```
crspeller, line 66: This is backwards. `UserAccessLevelAll = 0` and `ChannelAccessLevelNone = 3`

server/ai/configuration.go
```diff
@@ -49,10 +57,30 @@ type BotConfig struct {
...
+		return false
+	}
+	if c.UserAccessLevel < UserAccessLevelAll || c.UserAccessLevel > UserAccessLevelNone {
```
crspeller, line 69: Same here

server/ai/configuration.go
```diff
@@ -49,10 +57,30 @@ type BotConfig struct {
 }
 
 func (c *BotConfig) IsValid() bool {
```
crspeller, line 59: It might be nice to have this return an error with a specific failure. (I bet Aider could make this change pretty quick)

server/ai/configuration.go
```diff
@@ -49,10 +57,30 @@ type BotConfig struct {
...
+		return false
+	}
+	if c.UserAccessLevel < UserAccessLevelAll || c.UserAccessLevel > UserAccessLevelNone {
+		return false
+	}
+
+	// Service-specific validation
+	switch c.Service.Type {
+	case ServiceTypeOpenAI:
+		return c.Service.APIKey != "" && c.Service.OrgID != ""
```
crspeller: The OrgID is optional.

crspeller (changes requested):
I like the improvements! Just a couple fixes.

server/ai/configuration.go
enzowritescode: I'm not sure I understand your comment.

`ChannelAccessLevelAll = 0` and `ChannelAccessLevelNone = 3`. The code checks that the value passed to it is not less than 0 and not greater than 3. Is that not correct?

server/ai/configuration.go
enzowritescode: `UserAccessLevelAll = 0` and `UserAccessLevelNone = 3`. The code checks that the value passed to it is not less than 0 and not greater than 3. Is that not correct?

crspeller: Yes I read those wrong. Don't do code reviews while jet-legged kids 🛫

PR #283, opened 2025-01-24
Fix version bump script

crspeller:
The version bump script doesn't work if there are other tags on the repository.


jespino (approved):
LGTM

PR #301, opened 2025-03-26
Add experimental embedding search

crspeller:
Adding an experimental embedding search implementation using pgvector and OpenAI compatible endpoints.


server/api.go
```diff
@@ -122,6 +129,11 @@ type AIBotInfo struct {
 	UserIDs            []string               `json:"userIDs"`
 }
 
+type AIBotsResponse struct {
+	Bots          []AIBotInfo `json:"bots"`
+	SearchEnabled bool        `json:"searchEnabled"`
```
jespino, line 136: I don't know if it makes sense, but maybe we want to make search enabled a setting for the bot itself. That way you can decide what AIs access the search and what is doesn't, increasing the granularity of the control and permissions over the functionality. This bot can search, and is available to this users, this bot is available to everybody but it can't search. Maybe makes no sense, but it is just a though. Feel free to ignore this comment :)

crspeller: Agree that it might be useful. Same as being able to control what specific tools are available to which bots. I think a UI for determining which tools each bot can access is inevitable. Going to defer this from this PR for now and add this as a control for when we implement a unified UI for all tools.

server/api_admin.go
```diff
@@ -7,11 +7,138 @@ import (
...
+	}
+	newSearch, err := p.initSearch()
+	if err != nil {
+		c.AbortWithError(http.StatusInternalServerError, fmt.Errorf("failed to re-initialize search: %w", err))
+		return
+	}
+	p.search = newSearch
+
+	for {
+		query := `SELECT
```
jespino: nit: should this query be moved to the search store to keep the same level of abstraction for the whole code block?

server/api_admin.go
```diff
@@ -7,11 +7,138 @@ import (
...
+	"fmt"
 
 	"github.com/gin-gonic/gin"
+	"github.com/mattermost/mattermost-plugin-ai/server/embeddings"
 	"github.com/mattermost/mattermost/server/public/model"
 )
 
+const defaultBatchSize = 100
+
+func (p *Plugin) handleReindexPosts(c *gin.Context) {
```
jespino, line 19: Can this lead to an API timeout? Should we use something like a job here?

server/api_admin.go
crspeller: Actually no, because it doesn't actually relate to the embedding search. It's getting the posts from the MM server in order to index them and is generic to any `EmbeddingSearch` service.

I could put it int the server/store.go file, but actually I am kinda fond of having everything in one place. Not sure on that one though.

server/api_admin.go
crspeller: Hmmm... yes you are correct. Plugins don't have access to the job server though. I will think about how to address this.

crspeller: @jespino Added an implementation for a job server to resolve this.

webapp/src/components/system_console/embedding_search_config.tsx
```diff
@@ -0,0 +1,623 @@
...
+interface JobStatusType {
+    status: string; // 'running' | 'completed' | 'failed' | 'canceled' | 'no_job'
+    error?: string;
+    started_at: string; // ISO string from server's time.Time
+    completed_at?: string;
+    processed_rows: number;
+    total_rows: number;
+}
+
+const EmbeddingSearchPanel = ({value, onChange}: Props) => {
```
jespino: nit: This component looks like a good candidate to be splitter in smaller components.

webapp/src/components/search_sources.tsx
```diff
@@ -0,0 +1,165 @@
...
+    userId: string;
+    content: string;
+    score: number;
+}
+
+interface SourceItemProps {
+    source: Source;
+}
+
+const SearchSource = ({source, index}: SourceItemProps & {index: number}) => {
```
jespino, line 111: nit: probably we can extract this subcomponent to another file to keep the separation of the styles too.

webapp/src/components/dot_menu.tsx
```diff
@@ -52,12 +52,12 @@ export const DropdownMenu = styled.div`
     box-shadow: 0px 8px 24px rgba(0, 0, 0, 0.12);
     border-radius: 4px;
 
-    z-index: 12;
+    z-index: 1000;
```
jespino: Are we abusing z-index: 1000;?

server/api_search.go
```diff
@@ -0,0 +1,339 @@
...
+	for _, result := range searchResults {
+		// Get channel name
+		var channelName string
+		channel, chErr := p.pluginAPI.Channel.Get(result.Document.ChannelID)
+		if chErr != nil {
+			p.pluginAPI.Log.Warn("Failed to get channel", "error", chErr, "channelID", result.Document.ChannelID)
+			channelName = "Unknown Channel"
+		} else {
+			if channel.Type == model.ChannelTypeDirect || channel.Type == model.ChannelTypeGroup {
+				channelName = "Direct Message"
```
jespino, line 60: Makes sense to make any distinction between direct and group? I mean, maybe the channelName should be "Group Message" in case of that being a group message.

server/api_search.go
```diff
@@ -0,0 +1,339 @@
...
+
+		// Get username
+		var username string
+		user, userErr := p.pluginAPI.User.Get(result.Document.UserID)
+		if userErr != nil {
+			p.pluginAPI.Log.Warn("Failed to get user", "error", userErr, "userID", result.Document.UserID)
+			username = "Unknown User"
+		} else {
+			username = user.Username
+			// If we have the user's first and last name, use those instead
```
jespino: To be as compatible as possible with Mattermost style here, we should pass the username, the firstname, the lastname and the nick name, and the interface should be the one deciding what to show. I haven't seen that yet in the code, so maybe this comment makes not sense.

server/post_indexing.go
```diff
@@ -0,0 +1,36 @@
...
+		return false
+	}
+
+	// Skip posts from bots
+	if p.IsAnyBot(post.UserId) {
+		return false
+	}
+
+	// Skip non-regular posts
+	if post.Type != model.PostTypeDefault {
```
jespino, line 21: I'm not 100% sure about this, but we can keep going this way and figure out later if we want to change it. My concern is about plugin post types, for example.

jespino (approved):
I added some comments, but I don't see anything concerning that should stop this from being merged.

webapp/src/components/system_console/embedding_search_config.tsx
crspeller: Split up and fixed.

webapp/src/components/search_sources.tsx
crspeller: I think it's fine for now like that.

webapp/src/components/dot_menu.tsx
crspeller: Was not actually necessary anymore.

server/api_search.go
crspeller: Done.

server/api_search.go
crspeller: Actually going to remove this for now. Probably need to improve this behavior later.

server/post_indexing.go
crspeller: I am not sure either. For now I am restricting it, but I think this might change.

PR #306, opened 2025-04-02
Fix the bot creation performance. Only create DMs when nessisary

crspeller:


webapp/src/components/rhs/rhs_new_tab.tsx
```diff
@@ -83,14 +87,48 @@ const setEditorText = (text: string) => {
...
+
+    // If botChannelId is empty, we need to create a direct channel
+    useEffect(() => {
+        const createDirectChannel = async () => {
+            if (!botChannelId && !creatingChannel && activeBot) {
+                setCreatingChannel(true);
+                const botId = activeBot.id;
+
+                // This will as a side effect create the direct channel for us
+                const newChannelID = await getBotDirectChannel(currentUserId, botId);
```
nickmisasi: If this fails, I think we'll be left with an infinite spinner. Could you add a catch, and handle that case? Even just a generic "an error has occurred please try again"?

nickmisasi (changes requested):

jespino (approved):
LGTM except for the comment from Nick.

nickmisasi (approved):
LGTM. Thanks!

PR #307, opened 2025-04-03
Add inter-plugin communication endpoint

jespino:
## Description

This enables plugins to add AI requests, this only includes the non-streaming endpoint, we need to generate a streaming endpoint to cover more plugin use cases. Probably we will need to expand this based on other plugins requirements.


server/plugin.go
```diff
@@ -93,6 +95,16 @@ func resolveffmpegPath() string {
 
 func (p *Plugin) OnActivate() error {
 	p.pluginAPI = pluginapi.NewClient(p.API, p.Driver)
+	p.pluginSecret = p.getConfiguration().InterPluginSecretKey
```
crspeller: This is not necessary. You can just check if the header `Mattermost-Plugin-ID` is set. Then you know the request is coming from a plugin (and which one it is). 

Ref: https://github.com/mattermost/mattermost/blob/a9f09cadc2d5f121cdefbc0b9c255d55c944f327/server/channels/app/plugin_requests.go#L82

server/api.go
```diff
@@ -182,3 +197,104 @@ func (p *Plugin) handleGetAIBots(c *gin.Context) {
...
+	// Get the bot by username or use the first available bot
+	bot := p.GetBotByUsernameOrFirst(req.BotUsername)
+	if bot == nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": fmt.Sprintf("failed to get bot: %s", req.BotUsername)})
+		return
+	}
+
+	userID := req.RequesterUserID
+	if userID == "" {
+		userID = bot.mmBot.UserId
```
crspeller: This fallback seems dangerous.

server/api.go
```diff
@@ -182,3 +197,104 @@ func (p *Plugin) handleGetAIBots(c *gin.Context) {
...
+			{
+				Role:    llm.PostRoleUser,
+				Message: userPrompt,
+			},
+		},
+		Context: context,
+	}
+
+	// Apply any custom parameters for the model
+	options := []llm.LanguageModelOption{}
```
crspeller: Not sure I like how this works. Maybe we build this object on the client somehow?

crspeller (changes requested):
This only works for simple requests with only a single response. What about more advanced flows? Streaming? 
If we don't want to add those, we should call the API something appropriate for it. "
Let's also version this api in case we need to change it.

server/api.go
```diff
@@ -182,3 +197,104 @@ func (p *Plugin) handleGetAIBots(c *gin.Context) {
 
 	c.JSON(http.StatusOK, response)
 }
+
+func (p *Plugin) handleInterPluginCompletion(c *gin.Context) {
```
crspeller: This seems like it should be in a separate file.

server/api.go
jespino: Removed

jespino:
PTAL

interpluginclient/client.go
```diff
@@ -0,0 +1,124 @@
...
+	"net/http"
+	"time"
+
+	"github.com/mattermost/mattermost/server/public/plugin"
+	"github.com/pkg/errors"
+)
+
+const (
+	// DefaultTimeout is the default timeout for all requests to the AI plugin
+	DefaultTimeout = 30 * time.Second
```
crspeller: Too low. Long responses and function calls can take way longer than this. I use 10 minutes in the plugin: https://github.com/mattermost/mattermost-plugin-ai/blob/b590abb965a3ff98a0097792958669a4e49280e8/server/plugin.go#L108
(There is also a separate streaming timeout between streaming responses.)

server/configuration.go
```diff
@@ -19,6 +19,7 @@ type Config struct {
 	EnableLLMTrace           bool                             `json:"enableLLMTrace"`
 	AllowedUpstreamHostnames string                           `json:"allowedUpstreamHostnames"`
 	EmbeddingSearchConfig    embeddings.EmbeddingSearchConfig `json:"embeddingSearchConfig"`
+	InterPluginSecretKey     string                           `json:"interPluginSecretKey"`
```
crspeller: Remove this, unused.

crspeller (changes requested):
Just a couple small fixes.

interpluginclient/client.go
jespino: Done

server/configuration.go
jespino: Done

crspeller (approved):

PR #313, opened 2025-04-07
MM-63635 - modify the register function to be able to open outside ch…

pvev:
MM-63635 - modify the register function to be able to open outside channel context

## Description
In the webapp, during the app bar rendering process for the plugins there are validations to restrict which plugins must be only associated to channel context because those only make sense when run in a channel. Other plugins should be able to run even outside the context of a channel. By registering the plugin using the registerAppBarComponent the plugin component will contain the `rhsComponent` so the webapp will be able to open the RHS plugin in any context like in threads, drafts or scheduled messages.

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -75,8 +75,23 @@ export default class Plugin {
 
         let rhs: any = null;
         if (isRHSCompatable()) {
-            rhs = registry.registerRightHandSidebarComponent(RHS, RHSTitle);
-            setOpenRHSAction(rhs.showRHSPlugin);
+            if (!registry.registerAppBarComponent) {
+                return;
+            }
+
+            // Register with the app bar
+            const appBarComponent = registry.registerAppBarComponent({
+                iconUrl: aiIcon,
+                tooltipText: 'Copilot',
+                supportedProductIds: null,
+                rhsComponent: RHS,
+                rhsTitle: RHSTitle,
+            });
+
+            if (typeof appBarComponent === 'object' && appBarComponent.rhsComponent) {
+                rhs = appBarComponent.rhsComponent;
+                setOpenRHSAction(rhs.showRHSPlugin);
+            }
         }
 
         let currentUserId = store.getState().entities.users.currentUserId;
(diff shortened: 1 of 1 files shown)


webapp/src/index.tsx
```diff
@@ -137,7 +152,8 @@ export default class Plugin {
         }
 
         registry.registerAdminConsoleCustomSetting('Config', Config);
-        if (rhs) {
+
+        if (rhs && !registry.registerAppBarComponent) {
```
pvev, line 156: added this for backward compatibility, not sure if actually needed.

webapp/src/index.tsx
```diff
@@ -75,8 +75,23 @@ export default class Plugin {
 
         let rhs: any = null;
         if (isRHSCompatable()) {
-            rhs = registry.registerRightHandSidebarComponent(RHS, RHSTitle);
-            setOpenRHSAction(rhs.showRHSPlugin);
+            if (!registry.registerAppBarComponent) {
```
crspeller, line 78: We probably should fall back to the old method instead of returning here. (Unless they are on the same level of compatibility then we can just check in isRHSCompatable)

Side note: I really need to get back in to the code here and figure out what we actually need to support for backwards compatibility here.

webapp/src/index.tsx
crspeller: We would need to actually have the backwards compatibility code up there, but yes it would be nice for now to have it.

crspeller (changes requested):
Also seeing a bug where if I open the threads view while the RHS is open, the RHS appears to be open behind the threads view. Clicking on the button will close it (appearing to do nothing) then pressing it again opens it properly. Not sure if this is a bug with this implementation or on the webapp side.

pvev:
> Also seeing a bug where if I open the threads view while the RHS is open, the RHS appears to be open behind the threads view. Clicking on the button will close it (appearing to do nothing) then pressing it again opens it properly. Not sure if this is a bug with this implementation or on the webapp side.

hey @crspeller this is more for the webapp, not sure if a bug or a feature, but any of the other RHS plugins, like i.e Github when RHS is open and user moves to threads or drafts, the RHS plugin gets closed.

On the other hand, for your comment:

> We probably should fall back to the old method instead of returning here. (Unless they are on the same level of compatibility then we can just check in isRHSCompatable)

can you please clarify what you mean? thanks.

PR #316, opened 2025-04-21
Introduce proposal for Azure content moderation

davidkrauser:
## Description

This introduces a design proposal to introduce content moderation via Azure.

I don't know if we want to merge the proposal, but this PR gives us an easy way to discuss the details of the document. Let me know if there's a better way we could review this :slightly_smiling_face:

### DIFF
--- a/docs/design/moderation.md
+++ b/docs/design/moderation.md
@@ -1,0 +1,138 @@
+# Content Moderation Design Proposal
+
+This document outlines a proposal for implementing content moderation in the Mattermost AI Plugin using Azure AI Content Safety APIs. Here we explore how we might filter inappropriate content in both user inputs to LLMs and LLM-generated responses to protect users from potentially harmful, offensive, or inappropriate content.
+
+## API Selection
+
+### Azure Content Moderator (Deprecated)
+
+Azure Content Moderator is deprecated as of February 2024 and will be retired by February 2027. Microsoft recommends migrating to Azure AI Content Safety, which offers more advanced AI features and enhanced performance. Due to this deprecation, we should avoid using the older Content Moderator APIs for any new implementation.
+
+### Azure AI Content Safety (Recommended)
+
+Azure AI Content Safety is a comprehensive solution designed to detect harmful user-generated and AI-generated content in applications and services. It is suitable for many scenarios such as online marketplaces, gaming companies, social messaging platforms, enterprise media companies, and K-12 education solution providers.
+
+The Azure AI Content Safety APIs provide comprehensive content moderation capabilities:
+- Text analysis for harmful content (hate speech, sexual content, violence, self-harm)
+- Image analysis for inappropriate visual content
+- Multiple severity levels (0-Safe, 2-Low, 4-Medium, 6-High)
+- Support for blocklists and custom categories
+
+The proposed design below utilizes the AI Content Safety APIs.
+
+## Proposed Architecture
+
+The proposed moderation system would consist of the following components:
+
+1. **Moderation Interface**: A generic `Moderator` interface that defines methods for text and image moderation, allowing for different implementation providers.
+2. **Azure Implementation**: An implementation of the `Moderator` interface that integrates with Azure AI Content Safety APIs.
+3. **Integration Points**: Code that integrates moderation into the message processing flow, both for user input and LLM responses.
+
+### Service Availability Considerations:
+- The proposed design would use a "fail-closed" approach for safety - if the moderation service encounters an error, content would be blocked by default.
+- This ensures potentially problematic content doesn't slip through due to technical issues.
+- However, this means if moderation APIs are unavailable, all LLM interactions would be disabled by default.
+
+**Open Questions:**
+- Should there be an option for fail-open in certain trusted environments?
+- How can we implement graceful degradation instead of complete service shutdown?
+
+## Proposed Moderation Flow
+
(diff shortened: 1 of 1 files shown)


docs/design/moderation.md
```diff
@@ -0,0 +1,138 @@
...
+   - This happens in `streamResultToPost()` in `post_processing.go`.
+
+3. **Image Moderation**:
+   - Images attached to messages would be moderated before being sent to vision-capable LLMs.
+   - Images that fail moderation would not be included in the context sent to the LLM.
+   - This happens in `PostToAIPost()` in `post_processing.go`.
+
+### Review Mechanism Considerations:
+- If content exceeds the configured threshold, it would be automatically rejected without any human review process.
+- This binary approach (allow/reject) may be too rigid for some use cases.
```
wiggin77: For now, we would not implement a review queue. However rejected content should be audit logged.

docs/design/moderation.md
```diff
@@ -0,0 +1,138 @@
...
+  - Allow (below threshold)
+  - Warn user but allow (borderline)
+  - Block but allow override with explanation (moderately above threshold)
+  - Block completely (significantly above threshold)
+- **Feedback Loop**: We could create a mechanism for users to contest moderation decisions, providing a learning loop for the system and reducing false positives over time.
+
+### User Feedback Considerations:
+- Users receive generic messages when their content is flagged, without specifics about why.
+
+**Open Questions:**
```
wiggin77: Let's focus on a minimum viable product for the first iteration of this.  We can prioritize enhancements after we get customer feedback.

docs/design/moderation.md
```diff
@@ -0,0 +1,138 @@
...
+  - 4: Medium severity (moderate)
+  - 6: High severity (severe)
+- Configuration control over which severity levels trigger moderation actions
+
+### Threshold Configuration Considerations:
+- This proposal uses category-specific thresholds (hate, sexual, violence, selfHarm) based on Azure Content Safety's severity levels.
+- The default configuration would block content at medium severity (4) or higher.
+
+**Open Questions:**
+- Should thresholds be configurable per channel or per team?
```
wiggin77: Not at this time.

docs/design/moderation.md
```diff
@@ -0,0 +1,138 @@
...
+- How can administrators test and tune thresholds effectively?
+- Should we implement different actions based on different severity levels (e.g., warn vs. block)?
+
+**Potential Enhancements:**
+- **Blocklist Integration**: We could leverage Azure Content Safety's blocklist feature to allow administrators to create custom lists of forbidden terms or patterns.
+- **Channel/Team-Specific Settings**: We could enable different moderation policies for different contexts - stricter in public channels, more lenient in private teams with trusted users.
+- **Jailbreak Detection (Prompt Shield)**: We could utilize Azure's specialized Text Jailbreak Detection API for identifying prompt injection attacks. This API specifically detects prompts designed to manipulate or "jailbreak" an LLM's safety guardrails. The API returns a simple boolean result indicating whether a jailbreak attempt was detected.
+
+### Logging and Audit Considerations:
+- All moderation decisions should be logged
```
wiggin77: Main logging is fine, provided the actual content is not included.  Audit logging is mandatory.

docs/design/moderation.md
```diff
@@ -0,0 +1,138 @@
...
+
+**Open Questions:**
+- What additional metadata should be logged for moderation decisions?
+- Should we implement an admin dashboard for monitoring moderation activity?
```
wiggin77: Not at this time.

docs/design/moderation.md
```diff
@@ -0,0 +1,147 @@
...
+
+The proposed implementation would include three primary moderation points:
+
+1. **User Input Moderation**:
+   - When a user sends a message to an AI bot, the message would be checked by the moderation service.
+   - If flagged, the user would receive a notification and the message would not be processed by the LLM.
+   - This happens in `processUserRequestToBot()` in `conversations.go`.
+
+2. **LLM Response Moderation**:
+   - When an LLM generates a response, the complete response would be checked before being displayed to users.
```
jespino: We can do this but it would degrade the user experience, because we stream the response to the user while it is generating. So, there is two ways of doing it, waiting until the whole thing is generated, check it, and decide to sent it to the user (that would make the UX awful), or generating a small buffer and check while is generating, for example every 10 words, or something like that, and stream the words after checking, but that would delay the reply and also would increase drastically the amount of request to the Azure API.

docs/design/moderation.md
```diff
@@ -0,0 +1,147 @@
...
+   - If flagged, users would see a generic message instead of the problematic content.
+   - This happens in `streamResultToPost()` in `post_processing.go`.
+
+3. **Image Moderation**:
```
jespino: I think in this would be also needed to include moderation in the images generated by the AI. I would expect that eventually we support image generation inside mattermost (now is not supported).

docs/design/moderation.md
davidkrauser: We talked about this briefly offline, and we're going to disable streaming when we need to moderate the AI content. We're also going to implement moderation in a separate plugin. The doc has been updated

docs/design/moderation.md
davidkrauser: That's good to keep in mind :+1:

PR #318, opened 2025-04-30
Adding experimental MCP implementation

crspeller:
Integrates the Model Context Protocol (MCP) as a Client and adds a tool approval system. The implementation creates user-specific connections to external MCP servers and requires explicit approval before AI tools can execute operations. The updates improve LLM providers with streaming tool calls allowing for the user confirmation flow.

This is an experimental initial implementation and is hidden in the system console when disabled.


server/mcp/mcp.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// Package mcp provides a client for the Model Control Protocol (MCP) that allows
```
crspeller, line 4: Read this comment first for explanation on why the mcp package is organized like this. Ask any questions here so I can integrate the answers into the comment.

server/mcp/user_client.go
```diff
@@ -0,0 +1,284 @@
...
+
+	"github.com/invopop/jsonschema"
+	"github.com/mark3labs/mcp-go/client"
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mattermost/mattermost-plugin-ai/server/llm"
+	"github.com/mattermost/mattermost/server/public/pluginapi"
+	orderedmap "github.com/wk8/go-ordered-map/v2"
+)
+
+const MMUserIDHeader = "X-Mattermost-UserID"
```
crspeller, line 21: This would allow someone to write an MCP server that trusted the Mattermost auth.

server/anthropic/anthropic.go
```diff
@@ -144,6 +141,42 @@ func conversationToMessages(posts []llm.Post) (string, []anthropicSDK.MessagePar
...
+								Text: anthropicSDK.F(tool.Result),
+							},
+						}),
+					}
+					resultBlocks = append(resultBlocks, toolResultBlock)
+				}
+			}
+
+			if len(resultBlocks) > 0 {
+				flushCurrentMessage()
```
jespino, line 174: Not related to this PR, but this pattern is hard to follow for me, I think I would prefer something like `flushMessage(currentRole, currentMessage)` for each call, making it very explicit. I fill this is kinda a "global variable" style. I don't know if there is any reason for doing it in this way.

server/anthropic/anthropic.go
```diff
@@ -282,15 +310,12 @@ func (a *Anthropic) ChatCompletion(request llm.CompletionRequest, opts ...llm.La
...
 
 	go func() {
-		defer close(output)
-		defer close(errChan)
+		defer close(eventStream)
 
-		if err := a.streamChatWithTools(initialState); err != nil {
-			errChan <- err
-		}
+		_ = a.streamChatWithTools(initialState)
```
jespino: I think we can remove the return here for this function entirely, because it is returning the errors through the stream. The only problem is there is one error that is not being returned through the channel. I don't know if that is an error or is intentional, but in any case the error is not managed. Other option would be extract here the error here, and send to the channel here.

server/built_in_tools.go
```diff
@@ -391,6 +391,79 @@ func (p *Plugin) toolSearchServer(llmContext *llm.Context, argsGetter llm.ToolAr
 	return formatted, nil
 }
 
+/*type CreateChannelArgs struct {
```
jespino: Why this is commented out? Makes sense to directly remove this?

server/anthropic/anthropic.go
crspeller: Agree, needs a refactor there. Will do in a follow up.

server/anthropic/anthropic.go
crspeller: Good catch.

server/built_in_tools.go
crspeller: removed.

server/mcp/user_client.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to list tools: %w", err)
+	}
+
+	// Store the tools for this server
+	for _, tool := range result.Tools {
+		serverClient.tools[tool.Name] = tool
+
+		// Check for tool name conflicts across servers
+		if existingTool, exists := c.toolDefs[tool.Name]; exists {
```
jespino, line 138: I'm not 100% about this. maybe we should be able to have not-conflicting version of this, something like adding a the serverID as prefix to the tool if there is a conflict. That way, you can use different tools (based on the description) even if they conflict in the name.

server/mcp/user_client.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+
+		if err := c.connectToServer(context.Background(), serverID, serverConfig); err != nil {
+			c.log.Error("Failed to connect to MCP server", "userID", c.userID, "serverID", serverID, "error", err)
+			continue
+		}
+	}
+
+	// If no servers were successfully connected, return error
+	if len(c.clients) == 0 {
```
jespino, line 72: There is a case where you run "ConnectToAllServers" and the servers list is empty that this could happen, and I don't think it makes sense. The other thing is, this function doesn't error even if most of the connectToServer failed. What I understand is more forgiving, and still it is shown in the logs, but I'm not sure about that behavior. I think I like the, "best-effort" approach that you are following here, but I want to raise the concern about it.

server/openai/openai.go
```diff
@@ -351,7 +358,7 @@ func (s *OpenAI) streamResultToChannels(request openaiClient.ChatCompletionReque
 			}
 
 			// Add the tool calls to the request
-			request.Messages = append(request.Messages, openaiClient.ChatCompletionMessage{
+			/*request.Messages = append(request.Messages, openaiClient.ChatCompletionMessage{
```
jespino: Looks like we should remove this entirely instead of commenting it out

webapp/src/components/llmbot_post.tsx
```diff
@@ -117,28 +136,70 @@ export const LLMBotPost = (props: Props) => {
...
+
+            return () => {
+                if (props.websocketUnregister) {
+                    props.websocketUnregister(props.post.id, listenerID);
+                }
+            };
+        }
+
+        return () => {/* no cleanup */};
+    }, [props.post.id]);
```
jespino, line 202: I guess is not a problem, but should be the `props.websocketRegister`, and `props.websocketUnregister` part of the dependencies of this effect?

webapp/src/components/system_console/mcp_servers.tsx
```diff
@@ -0,0 +1,551 @@
...
+                    <PlusIcon size={14}/>
+                    <FormattedMessage defaultMessage='Add Header'/>
+                </AddHeaderButton>
+            </HeadersSection>
+        </ServerContainer>
+    );
+};
+
+// Main component for MCP servers configuration
+const MCPServers = ({value, onChange}: Props) => {
```
jespino, line 201: Make sense to split this file in two?

webapp/src/components/tool_approval_set.tsx
```diff
@@ -0,0 +1,109 @@
...
+        }
+    };
+
+    const handleReject = async (toolID: string) => {
+        try {
+            // Update UI immediately to prevent double-clicking
+            setProcessingTools((prev) => [...prev, toolID]);
+
+            // Call the API endpoint
+            await doToolCall(props.postID, []);
```
jespino: I don't know if that currently can happen but if you have multiple tools requesting for permission at the same time, all gets rejected. Also, if you have multiple of them, and you accept 1 of them, you are rejecting the others. I'm not sure if that is the case, but that is my understanding from the  code.

server/mcp/user_client.go
crspeller: We probably should do something here. The issue is the tool names are the prompt engineering of the tools they are kinda important so we don't want to mess them up unless we have to. 

Not going to resolve this in this initial PR, but I will create a ticket for this.

server/mcp/user_client.go
crspeller: Will add a check for 0 and not error out. 
I don't really see a better approach here. I think the way to resolve this is a UI for admins to see the status of MCP servers.

jespino (changes requested):
Some small things to check. But in general looks good to me.

webapp/src/components/llmbot_post.tsx
crspeller: I don't think it's an issue. I tend to not bother with the function dependencies and be explicit about what I think the useEffect should update on. 2/5

webapp/src/components/system_console/mcp_servers.tsx
crspeller: Agreed. Will do it in one of the follow up refactoring PRs.

webapp/src/components/tool_approval_set.tsx
crspeller: Fixed.

webapp/src/components/llmbot_post.tsx
jespino: Ok, for useEffect make sense

jespino (approved):
LGTM

PR #320, opened 2025-05-06
Add disableStreaming setting to support moderation of LLM responses

davidkrauser:
## Overview
This change enables content moderation of AI responses by delaying their display until fully generated. External plugins can now inspect and potentially modify responses before users see them, which is critical for organizations with compliance requirements or content policies.

More discussion about this effort can be found here: https://github.com/mattermost/mattermost-plugin-ai/pull/316#discussion_r2054724301

### Implementation Note

This PR introduces a new listener for the `post_edited` websocket event. Without this,  the UI will not update the post with the final contents after the call to `p.pluginAPI.Post.UpdatePost()` in `post_processing.go`.

## Demo
Here is a video demo of this new functionality in action. Note that with streaming disabled, the message does not appear until it is fully generated.

[streaming-disabled-demo.webm](https://github.com/user-attachments/assets/719f41ce-8fe9-4c68-821b-1fc8a18b296f)

### DIFF
--- a/e2e/helpers/plugincontainer.ts
+++ b/e2e/helpers/plugincontainer.ts
@@ -18,6 +18,7 @@ const RunContainer = async (): Promise<MattermostContainer> => {
 		  "allowPrivateChannels": true,
 		  "disableFunctionCalls": false,
 		  "enableLLMTrace": true,
+		  "disableStreaming": false,
 		  "enableUserRestrictions": false,
 		  "defaultBotName": "mock",
 		  "bots": [
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -17,6 +17,7 @@ type Config struct {
 	DefaultBotName           string                           `json:"defaultBotName"`
 	TranscriptGenerator      string                           `json:"transcriptBackend"`
 	EnableLLMTrace           bool                             `json:"enableLLMTrace"`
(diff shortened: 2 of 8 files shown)


crspeller (changes requested):
Breaking functionality doesn't seem like the right approach here.

If I am understanding correctly this is for Azure AI content moderation. However you don't need to use the moderation API to have this implemented in Azure. It is built right into the Azure OpenAI service and you can apply it to any deployment.

Unless I am missing something here?

davidkrauser:
@crspeller wrote:
> If I am understanding correctly this is for Azure AI content ~moderation~ safety.

I talked with the team, and we want to moderate _all_ content on an instance, not just the responses from the Azure OpenAI bot. Additionally, we want the ability to moderate any LLM response, regardless of backend.

A separate moderation plugin works nicely for that use case.

To work with an external moderation plugin, this PR makes two changes:
1. Adds an option to disable streaming. This isn't _strictly_ necessary, but provides a mechanism to delay showing the user any output before it has been vetted by the moderation plugin.
2. Adds an additional websocket event listener on the client side to watch for post changes. This is necessary. Otherwise, any edits to the post will not appear without a browser refresh.

WDYT?

PR #321, opened 2025-05-06
Refactor Plugin Architecture for Improved Testability and Evaluation

crspeller:
### Description
First phase of refactoring to make the plugin more testable, especially for evals. Restructuring the monolithic server package into dedicated packages with clearer responsibilities.

- Created new agents package to replace the monolithic server implementation
- Extracted some packages from server.
- Added mocking infrastructure with mockery
- Reorganized files to better reflect what they actually do
- Added groundwork for future evals
- The server package only contains the plugin now. Everything else is extracted (although the agents package looks suspiciously like the former plugin package) 

### Future work

- Add more testing and proper evals
- Clean up the initialization mess
- Make the configuration make a bit more sense within the Agents package. 
- Continue refactoring agents package into separate packages like the new react package
- Not 100% sure on the `agents` name. The main package will likely live on with common stuff. I think it works because its like "react agent" or "thread summarization agent". Maybe a more forward looking name. (Especially since we might rename the plugin to Mattermost Agents)

### Review notes
The PR looks bigger than it is because of the large amount of moved and split files. Most files can be ignored.
Important parts to look at:
- New package/folder structure while considering the goal is to turn most functionality into packages like `react`. Appreciate comments on this. 
- agents/react package for an idea of the goal here. 
- evals package for how evals will run (Checks will be implemented with signatures like `func LLMRubric(e *EvalT, result, rubric string) (float32, error)`)


api/api_test.go
```diff
@@ -1,16 +1,17 @@
 // Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-package main
+package api
 
+/*
```
jespino: Is this commented out for any reason? if the tests are not working, maybe we should put a skip on them instead of commenting them out.

jespino (approved):
LGTM

PR #325, opened 2025-05-13
Support new Enterprise Advanced sku

marianunez:
## Description

Added support for Enterprise Advanced sku by:

- Updating plugin api to get latest support for Enterprise Advanced through the latest server public dependency.
- Added support for Enterprise Advanced sku in webapp checks. Removed legacy E10/E20 skus.

### Ticket
https://mattermost.atlassian.net/browse/MM-63646


webapp/src/license.tsx
```diff
@@ -5,27 +5,25 @@ import {useSelector} from 'react-redux';
...
 const professional = 'professional';
 const enterprise = 'enterprise';
+const enterpriseAdvanced = 'advanced';
 
 // isValidSkuShortName returns whether the SKU short name is one of the known strings;
-// namely: E10 or professional, or E20 or enterprise
+// namely: professional, enterprise or enterprise advanced.
 const isValidSkuShortName = (license: Record<string, string>) => {
     switch (license?.SkuShortName) {
-    case e10:
```
jespino, line 17: Does this break backward compatibility? I don't know what is the latest version that we support from Mattermost or the latest license that we support from Mattermost.

crspeller (approved):
Looks like we broke backwards compatibility on the sever too so I guess it's fine. There are the other fallbacks anyway to catch older licenses.

PR #326, opened 2025-05-19
MM-62940 Require endpoints that don't have parameters in the body to have empty body

crspeller:
## Description

Enforces endpoints that don't take body parameters to error when a body is provided.

### DIFF
--- a/api/api_post.go
+++ b/api/api_post.go
@@ -52,6 +52,11 @@ func (a *API) handleReact(c *gin.Context) {
 	channel := c.MustGet(ContextChannelKey).(*model.Channel)
 	bot := c.MustGet(ContextBotKey).(*agents.Bot)
 
+	if err := a.enforceEmptyBody(c); err != nil {
+		c.AbortWithError(http.StatusBadRequest, err)
+		return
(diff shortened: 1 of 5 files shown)


api/api_post.go
```diff
@@ -52,6 +52,11 @@ func (a *API) handleReact(c *gin.Context) {
 	channel := c.MustGet(ContextChannelKey).(*model.Channel)
 	bot := c.MustGet(ContextBotKey).(*agents.Bot)
 
+	if err := a.enforceEmptyBody(c); err != nil {
+		c.AbortWithError(http.StatusBadRequest, err)
+		return
+	}
+
```
esarafianou, line 59: @crspeller `handleThreadAnalysis` is the only API that we don't add any check with this PR. Is there a reason for that?

esarafianou (approved):
@crspeller Looks good, just a comment on why 1 endpoint is without the check in the post API?

api/api_post.go
crspeller: Yes. It doesn't actually need it because its body is non-empty.

PR #329, opened 2025-05-22
Add experimental backend.

crspeller:


pavelzeman:
Thanks for putting this back Chris, so that we can continue the conversation with Nic.

Two outstanding questions we need answers for:
* does AskSage really not support streaming (and why)?
* does AskSage have a way to track token usage (so that we don't have to approximate things).

Once merged - will this be available on Corpus, so Nick can sign in and try it out?

crspeller:
@pavelzeman Also no tools support.

Yes it will, but I will have to configure it first.

PR #331, opened 2025-05-25
Avoid issues on Darwin bundling the plugin

danilvalov:
The same as https://github.com/mattermost/mattermost-plugin-starter-template/pull/218


crspeller (approved):
Thanks!

PR #337, opened 2025-06-06
Update user_guide.md

cwarnermm:
Removed GitHub-Markdown formatted notes that the MystParser can't natively interpret for single-source inclusion in docs.


cwarnermm:
@crspeller - That's the first thing I tried and I wasn't seeing the results I was expecting. Given that the docs are also reducing complexity and noise by reducing the number of callouts like this overall, I elected to incorporate the note content on this page into the body text.

PR #350, opened 2025-06-23
WIP: Mattermost MCP server initial setup

BenCookie95:
## Description

- Transport abstraction: Clean interface allowing stdio (current) and HTTP (planned) transport modes
- Authentication providers: Pluggable authentication supporting PAT (implemented) and OAuth (framework ready)
- mcp-go integration: Uses mcp-go library for protocol compliance via abstracted interface
- Dual mode operation: Standalone server + future plugin integration using same codebase

### DIFF
--- a/mcpserver/cmd/main.go
+++ b/mcpserver/cmd/main.go
@@ -1,0 +1,204 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/mattermost/mattermost-plugin-ai/mcpserver"
+	"github.com/mattermost/mattermost/server/public/shared/mlog"
+)
+
+const (
+	defaultTimeout = 30 * time.Second
+)
+
+func main() {
+	// Parse command line flags
+	var (
+		serverURL    = flag.String("server-url", "", "Mattermost server URL (required)")
+		token        = flag.String("token", "", "Personal Access Token")
+		transport    = flag.String("transport", "stdio", "Transport type (stdio, http) - stdio is default")
+		httpPort     = flag.Int("port", 8080, "HTTP port for http transport (default: 8080)")
+		debug        = flag.Bool("debug", false, "Enable debug logging")
+		logFile      = flag.String("logfile", "", "Path to log file (logs to file in addition to stderr)")
+		clientID     = flag.String("oauth-client-id", "", "OAuth client ID (required for OAuth auth)")
+		clientSecret = flag.String("oauth-client-secret", "", "OAuth client secret (required for OAuth auth)")
+		redirectURL  = flag.String("oauth-redirect-url", "", "OAuth redirect URL (required for OAuth auth)")
+		version      = flag.Bool("version", false, "Show version information")
+	)
+	flag.Parse()
+
+	if *version {
+		fmt.Fprintf(os.Stderr, "Mattermost MCP Server v0.1.0\n")
+		os.Exit(0)
+	}
+
+	// Set up mlog logger
+	logger, err := mlog.NewLogger()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to create logger: %v\n", err)
+		os.Exit(1)
+	}
+
+	cfg := make(mlog.LoggerConfiguration)
+	cfg["console"] = mlog.TargetCfg{
+		Type:          "console",
+		Levels:        mlog.StdAll,
+		Format:        "plain",
+		FormatOptions: json.RawMessage(`{"enable_color": false}`),
+		Options:       json.RawMessage(`{"out": "stderr"}`),
+		MaxQueueSize:  1000,
+	}
+
+	if *debug {
+		cfg["console"] = mlog.TargetCfg{
+			Type:          "console",
+			Levels:        []mlog.Level{mlog.LvlDebug, mlog.LvlInfo, mlog.LvlWarn, mlog.LvlError},
+			Format:        "plain",
+			FormatOptions: json.RawMessage(`{"enable_color": false}`),
+			Options:       json.RawMessage(`{"out": "stderr"}`),
+			MaxQueueSize:  1000,
+		}
+	}
+
+	// Add file logging if logfile flag is provided
+	if *logFile != "" {
+		cfg["file"] = mlog.TargetCfg{
+			Type:         "file",
+			Levels:       mlog.StdAll,
+			Format:       "json",
+			Options:      json.RawMessage(fmt.Sprintf(`{"compress": false, "filename": "%s"}`, *logFile)),
+			MaxQueueSize: 1000,
+		}
+	}
+
+	err = logger.ConfigureTargets(cfg, nil)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to configure logger: %v\n", err)
+		os.Exit(1)
+	}
+	logger.RedirectStdLog(mlog.LvlStdLog)
+
+	// Check required parameters
+	if *serverURL == "" {
+		// Try environment variable
+		*serverURL = os.Getenv("MM_SERVER_URL")
+		if *serverURL == "" {
+			logger.Error("server URL is required (use -server-url or MM_SERVER_URL environment variable)")
+			os.Exit(1)
+		}
+	}
+
+	// Determine authentication mode
+	hasTokenAuth := *token != "" || os.Getenv("MM_ACCESS_TOKEN") != ""
+	hasOAuthAuth := *clientID != "" || *clientSecret != "" || *redirectURL != ""
+
+	if !hasTokenAuth && !hasOAuthAuth {
+		logger.Error("authentication is required: either PAT auth (-token) or OAuth auth (-oauth-client-id, -oauth-client-secret, -oauth-redirect-url)")
+		os.Exit(1)
+	}
+
+	if hasTokenAuth && hasOAuthAuth {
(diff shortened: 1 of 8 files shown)


mcpserver/cmd/main.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+const (
+	defaultTimeout = 30 * time.Second
+)
+
+func main() {
+	// Parse command line flags
+	var (
+		serverURL = flag.String("server-url", "", "Mattermost server URL (required)")
```
crspeller: Probably better to switch this to using cobra. Easier to maintain. (easy claude code fix)

mcpserver/config.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// Transport type (currently only stdio is supported)
+	Transport string `json:"transport"`
+
+	// Development mode enables additional tools for setting up test data
+	DevMode bool `json:"dev_mode"`
+}
+
+// AuthenticationProvider handles authentication for MCP requests
+type AuthenticationProvider interface {
```
crspeller: This doesn't feel like it belongs in config.
It also feels like it's not generic it's for tokens specifically. Presumably this is where you mean to hook in OAuth Later?

mcpserver/server.go
```diff
@@ -0,0 +1,465 @@
...
+// with Mattermost-specific tools and authentication
+type MattermostMCPServer struct {
+	mcpServer    *server.MCPServer
+	authProvider AuthenticationProvider
+	logger       mlog.LoggerIFace
+	config       Config
+}
+
+// NewMattermostMCPServer creates a new Mattermost MCP server with the specified configuration
+func NewMattermostMCPServer(config Config, authProvider AuthenticationProvider, logger mlog.LoggerIFace) (*MattermostMCPServer, error) {
```
crspeller: Maybe we should pass in the transport here too.

mcpserver/server.go
```diff
@@ -0,0 +1,465 @@
...
+	// Create the mcp-go server directly
+	mcpServer := server.NewMCPServer(
+		"mattermost-mcp-server",
+		"0.1.0",
+		server.WithToolCapabilities(true), // Enable tool list changed notifications
```
crspeller: Why true? Our server's tools won't change.

mcpserver/server.go
```diff
@@ -0,0 +1,465 @@
...
+	default:
+		return fmt.Errorf("unsupported transport type: %s", s.config.Transport)
+	}
+}
+
+// serveStdio starts the server using stdio transport
+func (s *MattermostMCPServer) serveStdio() error {
+	// Configure error logger to use our mlog logger if available
+	var errorLogger *log.Logger
+	if s.logger != nil {
```
crspeller: We can probably assume we will always pass this. Maybe we just accept the standard log interface to the constructor?

mcpserver/server.go
```diff
@@ -0,0 +1,465 @@
...
+		return result, nil
+	}
+}
+
+// getAuthenticatedClient gets an authenticated client for the request
+func (s *MattermostMCPServer) getAuthenticatedClient(ctx context.Context) (*model.Client4, string, error) {
+	// For OAuth mode, token must come from request context (set by HTTP transport)
+	// For PAT mode, token can come from context or fall back to config
+	var token string
+	if ctxToken, ok := ctx.Value(TokenKey).(string); ok && ctxToken != "" {
```
crspeller: What's this about? As far as I can we we never set this.

mcpserver/server.go
```diff
@@ -0,0 +1,465 @@
...
+
+	if token == "" {
+		return nil, "", fmt.Errorf("no authentication token available - ensure token is provided via context for OAuth or config for PAT")
+	}
+
+	// Create client directly - no validation needed since Mattermost APIs will validate
+	client := model.NewAPIv4Client(s.config.ServerURL)
+	client.SetToken(token)
+
+	return client, "", nil // userID not needed - Mattermost gets it from session
```
crspeller: Ok, but then why pass back "" at all?

mcpserver/server.go
```diff
@@ -0,0 +1,465 @@
...
+		// Fall back to config token for PAT mode
+		token = s.config.PersonalAccessToken
+	}
+
+	if token == "" {
+		return nil, "", fmt.Errorf("no authentication token available - ensure token is provided via context for OAuth or config for PAT")
+	}
+
+	// Create client directly - no validation needed since Mattermost APIs will validate
+	client := model.NewAPIv4Client(s.config.ServerURL)
```
crspeller: Doing this feels like maybe the job of that `AuthenticationProvider` interface. Maybe returning an authenticated APIv4Client is what that interface should do?

mcpserver/server.go
```diff
@@ -0,0 +1,465 @@
...
+						Type: "text",
+						Text: "Error: " + err.Error(),
+					},
+				},
+				IsError: true,
+			}, nil
+		}
+
+		// No need for user context - Mattermost gets userID from session
+		ctxWithUser := ctx
```
crspeller: AI artifact?

mcpserver/tools.go
```diff
@@ -0,0 +1,1036 @@
...
+	"strings"
+	"time"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/shared/mlog"
+)
+
+// MattermostToolProvider implements Mattermost operations for MCP tools
+type MattermostToolProvider struct {
```
crspeller: Not really a provider if there isn't a generic interface. Do we need this abstraction?

mcpserver/tools.go
```diff
@@ -0,0 +1,1036 @@
...
+	authProvider AuthenticationProvider
+	logger       mlog.LoggerIFace
+}
+
+// NewMattermostToolProvider creates a new Mattermost tool provider
+func NewMattermostToolProvider(authProvider AuthenticationProvider, logger mlog.LoggerIFace) *MattermostToolProvider {
+	return &MattermostToolProvider{
+		authProvider: authProvider,
```
crspeller: I don't think the tool executors need and auth provider. It's unused anyway.

mcpserver/server.go
```diff
@@ -0,0 +1,465 @@
...
+		toolProvider := NewMattermostToolProvider(s.authProvider, s.logger)
+
+		// Create dev tool provider for development tools
+		devToolProvider := NewDevToolProvider(s.authProvider, s.logger, s.config.ServerURL)
+
+		// Execute the tool using our existing implementation
+		var result *mcp.CallToolResult
+		switch toolName {
+		case "read_post":
+			result, err = toolProvider.readPost(ctxWithUser, client, request.Params.Arguments)
```
crspeller: We create providers above then pass the same arguments to every function. Maybe have a look at the way the ToolStore in the llm package works? Maybe you can even reuse that.

mcpserver/server.go
```diff
@@ -0,0 +1,465 @@
...
+	// This will be implemented when OAuth support is added
+	s.logger.Info("HTTP transport requested but not yet implemented")
+	s.logger.Info("Future implementation will support OAuth authentication and StreamableHTTP")
+	return fmt.Errorf("HTTP transport not yet implemented - will be added for OAuth support")
+}
+
+// registerMattermostTools registers all Mattermost tools with the MCP server
+func (s *MattermostMCPServer) registerMattermostTools() {
+	// Register read_post tool
+	readPostTool := mcp.NewTool("read_post",
```
crspeller: It might be nice if the registration and the implementation of the tool where in the same location. You can have a look at the mmtools package but it only solves the arguments part with struct tags.

mcpserver/server.go
BenCookie95: I think this got left over from some oauth boilerplate I later removed. I'll remove it for now to avoid confusion

mcpserver/server.go
BenCookie95: It's part of the `Config` struct

mcpserver/server.go
BenCookie95: Yeah I'll make that change

mcpserver/server.go
BenCookie95: I think putting the tools in it's own package in the MCP server makes sense. I like how the mmtools package has it split into logical files like user.go, search.go. My current implementation isn't very sustainable

mcpserver/server.go
BenCookie95: Yeah, also the user isn't in the context anymore after a refactor. I'll remove this

mcpserver/config.go
BenCookie95: Yeah this should be in auth.go. Token should refer to any authentication token whether it's for stdio or someone using oauth but extracting their auth token should be abstracted away into the functions themselves so that it's ready for oauth later. I'll fix that up

mcpserver/server.go
BenCookie95: I don't think I can use the ToolStore, I'd have to update the context to include the client which I don't think belongs there. I can copy the ToolStore pattern though and also use the argsGetter approach

mcpserver/server.go
BenCookie95: Yeah I'll remove it, I had the GetMe call inside here originally which allowed me to get the userID but I should have just removed this after we removed that call

mcpserver/server.go
BenCookie95: Yeah it already has GetMattermostClient which should probably be renamed to `GetAuthenticatedMattermostClient`. This `getAuthenticatedClient` function will be removed

mcpserver/server.go
BenCookie95: I decided to make separate constructors for each server type; stdio and oauth. I think it's a cleaner approach

PR #357, opened 2025-06-30
feat: Add custom HTTP headers support per LLM service instance

### LINKED ISSUE #343: 💡 idea: Custom headers support for models API
## Description

When setting up models, it should make sense to add a field that will configure additional headers.

These headers should then be passed to the models API with each call.

vpecinka:
Implements custom HTTP headers functionality as requested in issue #343. Each LLM service instance (bot configuration) can now have independent custom headers that are sent with every API request to the provider.

## Backend Changes:
- Added customHeaders field to ServiceConfig and all provider configs
- Implemented HTTP client wrapper to inject custom headers into requests
- Updated config transformation to preserve customHeaders across saves
- Added comprehensive unit and integration tests
- Supports all providers: OpenAI, Anthropic, ASage, OpenAI Compatible

## Frontend Changes:
- Added CustomHeadersItem React component for header management
- Integrated custom headers editor into System Console bot configuration
- Fixed focus issues with dynamic header key editing
- Updated TypeScript types to include customHeaders field

## Key Features:
- Per-service instance custom headers (independent between bots)
- Headers preserved during config save/load operations
- User-friendly UI for adding/editing/removing headers
- Comprehensive validation and error handling
- Full test coverage including multiple instance scenarios

## Use Cases:
- Authentication tokens specific to different API endpoints
- Request tracking and routing headers
- Organization-specific headers for multi-tenant setups
- Custom metadata for monitoring and analytics

Closes #343

# Closes #

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -32,6 +32,48 @@ type messageState struct {
 	context  *llm.Context
 }
 
+// customHeadersTransport wraps an http.RoundTripper to add custom headers to every request
+type customHeadersTransport struct {
+	base    http.RoundTripper
+	headers map[string]string
+}
+
+func (t *customHeadersTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Clone the request to avoid modifying the original
+	newReq := req.Clone(req.Context())
+
+	// Add custom headers
+	for key, value := range t.headers {
+		newReq.Header.Set(key, value)
+	}
+
+	return t.base.RoundTrip(newReq)
+}
+
+// wrapHTTPClientWithCustomHeaders wraps an http.Client to add custom headers to all requests
+func wrapHTTPClientWithCustomHeaders(baseClient *http.Client, customHeaders map[string]string) *http.Client {
+	if len(customHeaders) == 0 {
+		return baseClient
+	}
+
+	transport := baseClient.Transport
+	if transport == nil {
+		transport = http.DefaultTransport
+	}
+
+	wrappedClient := &http.Client{
+		Transport: &customHeadersTransport{
+			base:    transport,
+			headers: customHeaders,
+		},
+		CheckRedirect: baseClient.CheckRedirect,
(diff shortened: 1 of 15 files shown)


anthropic/anthropic.go
```diff
@@ -32,6 +32,48 @@ type messageState struct {
 	context  *llm.Context
 }
 
+// customHeadersTransport wraps an http.RoundTripper to add custom headers to every request
+type customHeadersTransport struct {
```
crspeller: We can probably just wrap the passed HTTP client for all back-ends rather than doing it for each back-end separately.

docs/custom_headers.md
```diff
@@ -0,0 +1,101 @@
+# Custom HTTP Headers Support
```
crspeller: This seems like an excessive amount of docs. Maybe just a simple document of the option in the system console?

examples/custom_headers_example.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/mattermost/mattermost-plugin-ai/config"
+	"github.com/mattermost/mattermost-plugin-ai/llm"
+	"github.com/mattermost/mattermost-plugin-ai/openai"
+)
+
+// Example of how to use custom headers with the LLM providers
+func exampleCustomHeaders() {
```
crspeller: We probably don't need these examples.

crspeller (changes requested):
Thanks for the contribution! I think we just need to clean it up a little.

anthropic/anthropic.go
vpecinka: fixed in c82d50c447266ffaea42673333063ada8f0f7f31

docs/custom_headers.md
vpecinka: ok, removed excessive docs in 931b47a699f49eeeb313a3d473cc0200db676597

PR #367, opened 2025-07-29
Streamable HTTP and SSE MCP server

BenCookie95:
## Description

- Added support for streamable http and backwards compatibility with sse
- Adjusted the team and channel tools to use the search endpoints
- Add OAuth resource metadata url

PRs for Mattermost OAuth updates will be linked below when they are ready

### DIFF
--- a/mcpserver/tools/channels.go
+++ b/mcpserver/tools/channels.go
@@ -103,6 +103,11 @@ func (p *MattermostToolProvider) toolReadChannel(mcpContext *MCPToolContext, arg
 		return "invalid parameters to function", fmt.Errorf("failed to get arguments for tool read_channel: %w", err)
 	}
 
+	// Validate channel ID
+	if !model.IsValidId(args.ChannelID) {
+		return "invalid channel_id format", fmt.Errorf("channel_id must be a valid ID")
+	}
+
 	// Set defaults and validate
 	if args.Limit == 0 {
 		args.Limit = 20
@@ -185,8 +190,8 @@ func (p *MattermostToolProvider) toolCreateChannel(mcpContext *MCPToolContext, a
 	if args.Type == "" {
 		return "type is required", fmt.Errorf("type cannot be empty")
 	}
-	if args.TeamID == "" {
-		return "team_id is required", fmt.Errorf("team_id cannot be empty")
+	if !model.IsValidId(args.TeamID) {
+		return "invalid team_id format", fmt.Errorf("team_id must be a valid ID")
 	}
 
 	// Validate channel type
@@ -234,6 +239,11 @@ func (p *MattermostToolProvider) toolGetChannelInfo(mcpContext *MCPToolContext, 
 	client := mcpContext.Client
 	ctx := context.Background()
 
+	// Validate team ID if provided
+	if args.TeamID != "" && !model.IsValidId(args.TeamID) {
+		return "invalid team_id format", fmt.Errorf("team_id must be a valid ID")
+	}
+
 	var channel *model.Channel
 
 	var lastError error
@@ -241,6 +251,10 @@ func (p *MattermostToolProvider) toolGetChannelInfo(mcpContext *MCPToolContext, 
 	// Try different lookup methods based on provided parameters
 	switch {
 	case args.ChannelID != "":
+		// Validate channel ID format
+		if !model.IsValidId(args.ChannelID) {
+			return "invalid channel_id format", fmt.Errorf("channel_id must be a valid ID")
+		}
 		// Direct ID lookup - fastest method
 		channel, _, err = client.GetChannel(ctx, args.ChannelID, "")
 		if err != nil {
@@ -314,8 +328,8 @@ func (p *MattermostToolProvider) toolGetChannelMembers(mcpContext *MCPToolContex
 	}
 
 	// Validate required fields
-	if args.ChannelID == "" {
-		return "channel_id is required", fmt.Errorf("channel_id cannot be empty")
+	if !model.IsValidId(args.ChannelID) {
+		return "invalid channel_id format", fmt.Errorf("channel_id must be a valid ID")
 	}
 
 	// Set defaults and validate
@@ -391,11 +405,11 @@ func (p *MattermostToolProvider) toolAddUserToChannel(mcpContext *MCPToolContext
 	}
 
 	// Validate required fields
-	if args.UserID == "" {
-		return "user_id is required", fmt.Errorf("user_id cannot be empty")
+	if !model.IsValidId(args.UserID) {
+		return "invalid user_id format", fmt.Errorf("user_id must be a valid ID")
 	}
-	if args.ChannelID == "" {
-		return "channel_id is required", fmt.Errorf("channel_id cannot be empty")
+	if !model.IsValidId(args.ChannelID) {
+		return "invalid channel_id format", fmt.Errorf("channel_id must be a valid ID")
 	}
 
 	// Get client from context
--- a/mcpserver/tools/posts.go
+++ b/mcpserver/tools/posts.go
@@ -88,6 +88,11 @@ func (p *MattermostToolProvider) toolReadPost(mcpContext *MCPToolContext, argsGe
 		return "invalid parameters to function", fmt.Errorf("failed to get arguments for tool read_post: %w", err)
 	}
 
+	// Validate post ID
+	if !model.IsValidId(args.PostID) {
+		return "invalid post_id format", fmt.Errorf("post_id must be a valid ID")
+	}
+
 	// Set default for include_thread
 	if !args.IncludeThread {
 		// Since bool defaults to false, we need to check if it was explicitly set
@@ -170,11 +175,15 @@ func (p *MattermostToolProvider) toolCreatePost(mcpContext *MCPToolContext, args
 	}
 
 	// Validate required fields
-	if args.ChannelID == "" {
-		return "channel_id is required", fmt.Errorf("channel_id cannot be empty")
+	if !model.IsValidId(args.ChannelID) {
+		return "invalid channel_id format", fmt.Errorf("channel_id must be a valid ID")
 	}
 	if args.Message == "" {
 		return "message is required", fmt.Errorf("message cannot be empty")
+	}
+	// Validate root ID if provided (for replies)
+	if args.RootID != "" && !model.IsValidId(args.RootID) {
+		return "invalid root_id format", fmt.Errorf("root_id must be a valid ID")
 	}
 
 	// Get client from context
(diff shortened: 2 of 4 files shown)


mcpserver/tools/channels.go
```diff
@@ -273,8 +273,50 @@ func (p *MattermostToolProvider) toolGetChannelInfo(mcpContext *MCPToolContext,
...
+		}
+
+		channels, _, searchErr := client.SearchAllChannelsForUser(ctx, searchTerm)
+		if searchErr != nil {
+			return "failed to search channels", fmt.Errorf("error searching channels: %w", searchErr)
+		}
+
+		// Find exact match by display name or name
+		for _, ch := range channels {
+			if ch.DisplayName == searchTerm || ch.Name == searchTerm {
```
BenCookie95: This would be an excellent first use case for elicitation where we can ask the user to choose between a number of results. Leaving it like this until we have support for it in the client side of agents

mcpserver/tools/channels.go
```diff
@@ -273,8 +273,50 @@ func (p *MattermostToolProvider) toolGetChannelInfo(mcpContext *MCPToolContext,
 		if err != nil {
 			return "channel not found by name", fmt.Errorf("error fetching channel by name: %w", err)
 		}
+	case args.ChannelDisplayName != "" || args.ChannelName != "":
```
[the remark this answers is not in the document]
BenCookie95: This is just wrong unless I'm losing it

mcpserver/README.md
```diff
@@ -194,6 +200,58 @@ To use with Claude Desktop, add the server to your MCP configuration:
...
+**Basic HTTP server setup:**
+```bash
+./bin/mattermost-mcp-server \
+  --transport http \
+  --server-url https://your-mattermost.com \
+  --http-port 8080 \
+  --debug
+```
+
+**HTTP server with custom port (recommended for production):**
```
crspeller: A little verbose to have two examples here.

mcpserver/README.md
```diff
@@ -194,6 +200,58 @@ To use with Claude Desktop, add the server to your MCP configuration:
...
+./bin/mattermost-mcp-server --transport http --http-port 8080 --site-url https://mcp.yourcompany.com
+```
+
+**Available endpoints:**
+- `/mcp`: Main MCP communication endpoint (SSE and streamable HTTP)
+- `/sse`: Server-Sent Events endpoint (backwards compatibility)
+- `/message`: Message endpoint for SSE transport (backwards compatibility)
+- `/.well-known/oauth-protected-resource`: OAuth metadata endpoint
+
+**Security considerations:**
```
crspeller: Probably unnecessary info.

mcpserver/http_server.go
```diff
@@ -0,0 +1,425 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/mark3labs/mcp-go/server"
```
crspeller, line 16: We should probably switch over to https://github.com/modelcontextprotocol/go-sdk soon. I already switched the client.

mcpserver/auth/provider.go
```diff
@@ -63,42 +66,93 @@ func (p *TokenAuthenticationProvider) GetAuthenticatedMattermostClient(ctx conte
...
+	client := model.NewAPIv4Client(p.mmServerURL)
+	client.SetOAuthToken(token)
+
+	return client, nil
+}
+
+// parseAndValidateOAuthToken extracts and validates OAuth token from context, returning token and user info
+func (p *OAuthAuthenticationProvider) parseAndValidateOAuthToken(ctx context.Context) (string, error) {
+	// Extract HTTP request from context
+	httpReq, ok := ctx.Value(HTTPRequestContextKey).(*http.Request)
```
crspeller: Not a fan of passing stuff though the context like this. I think idiomatically it's only supposed to be used for secondary stuff like logging or metrics. Could we just pass what we need to `GetAuthenticatedMattermostClient` or `NewOAuthAuthenticationProvider`?

mcpserver/http_server.go
```diff
@@ -0,0 +1,425 @@
...
+		"0.1.0",
+		server.WithToolCapabilities(false),
+		server.WithLogging(),
+		server.WithRecovery(),
+	)
+
+	// Register tools
+	mattermostServer.registerTools()
+
+	// Create SSE server using the mcp-go library with authentication
```
crspeller: Not sure I understand these comments.

mcpserver/http_server.go
```diff
@@ -0,0 +1,425 @@
...
+// createHTTPContextFunc creates a context function for Streamable HTTP authentication
+func (s *MattermostHTTPMCPServer) createHTTPContextFunc() server.HTTPContextFunc {
+	return func(ctx context.Context, r *http.Request) context.Context {
+		// Add HTTP request to context so OAuth provider can access it
+		return context.WithValue(ctx, auth.HTTPRequestContextKey, r)
+	}
+}
+
+// setupHTTPServerWithOAuth sets up the HTTP server with OAuth metadata endpoints and MCP routing
+func (s *MattermostHTTPMCPServer) setupHTTPServerWithOAuth(baseURL, addr string) error {
```
crspeller: Not sure this helper makes things more clear. Maybe inline it?

mcpserver/http_server.go
```diff
@@ -0,0 +1,425 @@
...
+		ReadTimeout:  30 * time.Second,
+		WriteTimeout: 30 * time.Second,
+	}
+
+	return nil
+}
+
+// GetTestHandler returns the HTTP handler for testing purposes
+func (s *MattermostHTTPMCPServer) GetTestHandler() http.Handler {
+	if s.httpServer != nil {
```
crspeller: Don't understand why we have to do this.

mcpserver/http_server.go
```diff
@@ -0,0 +1,425 @@
...
+	"github.com/mattermost/mattermost/server/public/shared/mlog"
+)
+
+// MattermostHTTPMCPServer wraps MattermostMCPServer for HTTP transport
+type MattermostHTTPMCPServer struct {
+	*MattermostMCPServer
+	config               HTTPConfig
+	sseServer            *server.SSEServer
+	streamableHTTPServer *server.StreamableHTTPServer
+	httpMux              *http.ServeMux
```
crspeller: Do we need to store this? I feel like it should be dynamically created then stored in the httpServer below.

mcpserver/http_server.go
```diff
@@ -0,0 +1,425 @@
...
+		}
+
+		// Set the authenticated context for downstream handlers
+		r = r.WithContext(ctx)
+		next(w, r)
+	}
+}
+
+// setupMCPRoutes sets up the MCP communication routes
+func (s *MattermostHTTPMCPServer) setupMCPRoutes() {
```
crspeller: I always like to have all my routes initialized in one place for smaller systems as it tends to be a way that new devs can figure out what the system does / how it works. Maybe combine this with the other routes like in https://github.com/mattermost/mattermost-plugin-agents/blob/master/api/api.go#L111
(Not in ServeHTTP though, yes I know :P)

mcpserver/http_server.go
```diff
@@ -0,0 +1,425 @@
...
+			w.WriteHeader(http.StatusOK)
+			return
+		}
+
+		next.ServeHTTP(w, r)
+	})
+}
+
+// loggingMiddleware creates a standard HTTP logging middleware
+func (s *MattermostHTTPMCPServer) loggingMiddleware(next http.Handler) http.Handler {
```
crspeller, line 333: Could use https://github.com/gin-gonic/gin like the plugin does to avoid some of the boilerplate like this. But the default stuff is probably fine 0/5

mcpserver/http_server.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+
+	s.logger.Warn("origin not in allowed list",
+		mlog.String("origin", origin),
+		mlog.Any("allowed_origins", allowedOrigins))
+	return false
+}
+
+// normalizeURL normalizes a URL by removing default ports and converting to lowercase
+func normalizeURL(u *url.URL) string {
```
jupenur, line 250: This strips square brackets from IPv6 hosts and won't work. I'd steer away from `fmt.Sprintf` altogether since the `net` and `url` packages have better alternatives.

mcpserver/http_server.go
BenCookie95: I started with gin to try match the plugin but I ended up switching to net/http. I cannot remember the issues I had with it because it was like 3 weeks ago. Either way I think this is fine

mcpserver/http_server.go
BenCookie95: Consolidated into a function called setupRoutes

mcpserver/http_server.go
BenCookie95: Should we wait until v1.0.0 is released? the README says it'll be sometime this month

crspeller: Actually it says in a couple of days they want to make a 1.0 release. We can wait until then. (separate PR)

mcpserver/auth/provider.go
BenCookie95: Having the http request in there wasn't a great idea but I do need a way to pass some sort of user context down the chain to the tools so they can easily create a Mattermost Client. I'm just passing the auth token into context now. I also moved the header parsing out of the auth layer and into the `http_server.go` file.

mcpserver/tools/file_utils.go
```diff
@@ -13,17 +13,23 @@ import (
...
-// fetchFileData fetches file data from a file path or URL and returns it as []byte
-func fetchFileData(filespec string) ([]byte, error) {
+// fetchFileDataForLocal fetches file data from a file path or URL (local access only)
+func fetchFileDataForLocal(filespec string, accessMode types.AccessMode) ([]byte, error) {
 	if filespec == "" {
 		return nil, fmt.Errorf("empty filespec provided")
 	}
 
+	// URLs are only allowed for local access mode
+	if accessMode != types.AccessModeLocal {
```
BenCookie95, line 27: I also added an accessMode check for each of these functions just to ensure they aren't used in the wrong place in future.

BenCookie95:
> [handleFileAttachments](https://github.com/mattermost/mattermost-plugin-agents/blob/52521e4ad01c097e21a6135b8d17e467037d2465/mcpserver/tools/util.go#L135) needs to be reworked before exposing the server over a network because currently it assumes attachments are from the local filesystem.

@jupenur I added an `access` struct tag for field-level access control during JSON deserialization and schema generation. Fields tagged with `access:"local"` (like file attachments) are restricted to local MCP connections (stdio), while `access:"remote"` fields are only available for remote connections. `NewJSONSchemaForAccessMode()` generates access-filtered schemas so LLMs won't be fed fields unavailable to their transport mode, and `validateAccessRestrictions()` prevents clients from making POST requests with restricted fields included. I also made sure the functions that are handling the local file uploads are locked down to the `local` access mode so that they aren't accidentally misused in future.

crspeller (approved):
Looks good pending the other changes above.

jupenur (approved):

esarafianou:
@BenCookie95 no need for my review since Juho has approved. Feel free to merge :)

Issue #368, opened 2025-07-31
💡 idea: Custom prompt in specific languages


zatkoma:
## Description

Seems like [prompts](https://github.com/mattermost/mattermost-plugin-agents/tree/master/prompts) are stored in english and seems there is [localization in prompts](https://github.com/mattermost/mattermost-plugin-agents/blob/master/prompts/locale.tmpl). That should makes sense as LLMs are usually multi–lingual but we should see in some tests that LLM's performs better if you communicate to them by same language in which you have instructions.

Now if someone wants to try this – then they have to make own copy of plugin and test that – but will be helpful if will be possible to customize prompts. Also from my experience custom instructions aren't usually enough.

vpecinka:
#369

PR #369, opened 2025-07-31
Add Multi-Language Support for AI Bot Prompts

vpecinka:
related to #368

Summary

Implements per-bot language selection for system prompts with fallback to English when translations are missing.
 - Type: Feature
 - Breaking Changes: None
 - Language Support: English (default), Czech (initial translation)
 

Changes
 - Backend: Added BotLanguage context field and language-aware prompt loading with fallback
 - Frontend: Added language dropdown in bot configuration (English/Czech)
 - Prompts: Organized templates by language subdirectories
 - Infrastructure: Language validation and embed system for prompt template loading

Features
 - Per-bot language configuration (no global default)
 - Automatic fallback to English for missing translations
 - Support for expanding to additional languages
 - Maintains backward compatibility

Testing
 - Verified prompt loading and fallback behavior
 - Confirmed UI language selection persists correctly
 - Tested with existing English prompts and new Czech translations

Future Work
 - Additional language translations (Spanish, French, German, etc.)
 - Automatic language detection based on user locale
 - Translation validation tooling


crspeller (changes requested):
Interesting. Is this driven by some behavior issues or some improvement seen when the system prompts language matches the uesrs? Just wondering the reasoning behind the PR.

As far as I have seen the language of the system prompt doesn't usually effect the quality of the output for any specific language. As far as I am aware, the system prompts of the major providers are not translated. 
Can you link to any resources that show doing this translations to increase performance? It's a lot of maintenance to commit to without knowing.

Also if we are to merge this I think we would want to automate the translations to all the supported languages.

zatkoma:
Hi guys, sorry for the late response. 🌴 Unfortunately, @crspeller, we don't have any public data on this. We primarily use in this plugin our own large language model which is post-trained on Czech tasks/instructions as our language is a bit niche. 

Since we are making our model we should see based anotations that LLMs performs better when the instructions are written in Czech, which is why we want to create translated prompts to our language. 

I totally understand that this may be niche, but customising the prompts is a sensible approach for smaller languages. We probably don't need automatic translations to all languages – if people use state-of-the-art models, that "system prompts" will work fine – but for smaller models makes sense to customize and dig-deep into prompts.

PR #370, opened 2025-08-08
Organise the file upload for mcp

BenCookie95:
## Description
We needed a proper place to put files that we want to upload for profile pictures or post attachments. I also decided it would be a good idea to move the binary inside the mcpserver directory since the data folder is also there.

We are a bit more stringent about the "deployment" structure now. We expect the binary to be inside the bin directory and there will be a data directory at the same level as bin.

Ticket from Eva: https://mattermost.atlassian.net/browse/MM-64998


mcpserver/README.md
```diff
@@ -200,7 +205,51 @@ Development mode (`--dev` flag) enables additional tools for setting up realisti
...
+│   │   └── mattermost-mcp-server  # MCP server binary
+│   ├── data/                      # File storage directory
+│   │   ├── documents/             # Example: organized folders
+│   │   ├── images/
+│   │   └── README.txt             # Example: files
+│   └── ...                        # Source code
+└── bin/                           # Other project binaries
+```
+
+### File Path Usage
```
crspeller: Not sure this is useful to the user as those are instructions for the LLM?

mcpserver/tools/file_utils.go
```diff
@@ -16,6 +16,38 @@ import (
 	"github.com/mattermost/mattermost/server/public/model"
 )
 
+// GetDataDirectoryInternal is the internal function that can be overridden in tests
+var GetDataDirectoryInternal = getDataDirectory
```
crspeller, line 20: Maybe we can pass the data directory to `fetchFileData` instead of this override?

mcpserver/tools/file_utils.go
```diff
@@ -44,7 +76,49 @@ func fetchFileData(filespec string) ([]byte, error) {
...
+	}
+
+	// If rel starts with ".." or is "..", the path is outside data directory
+	if strings.HasPrefix(rel, ".."+string(filepath.Separator)) || rel == ".." {
+		return nil, fmt.Errorf("path outside data directory: %s", filespec)
+	}
+
+	filePath := absRequestedPath
+
+	file, err := os.Open(filePath)
```
crspeller: We can probably simplify and make this code more secure with the new https://pkg.go.dev/os#Root in the os library. There is a blog post about it here: https://go.dev/blog/osroot

mcpserver/README.md
```diff
@@ -200,7 +205,51 @@ Development mode (`--dev` flag) enables additional tools for setting up realisti
...
+- **`create_user`**: Set profile images using the `profile_image` parameter (dev mode)
+- **`create_team`**: Set team icons using the `team_icon` parameter (dev mode)
+
+### File Management
+
+1. **Place files in the data directory**: Copy your files to `mcpserver/data/` or subdirectories within it
+2. **Reference with relative paths**: Use paths relative to the data directory (e.g., `report.pdf` for `mcpserver/data/report.pdf`)
+3. **Organize as needed**: Create subdirectories for better organization (e.g., `documents/`, `images/`)
+
+**Note**: The data directory is automatically created when the server starts and is excluded from git commits for security.
```
crspeller: Not sure what this means.

crspeller (changes requested):
I think this PR is probably overkill because we don't ever intend this functionality to be in the remote server right? If it ever does it will be though some kind of MCP file upload functionality not local files right?

I still think it's worth doing properly, security overkill is better than the alternative :)

mcpserver/README.md
BenCookie95: I do think it's useful for the user because they will have to point the llm to where the file is. We don't want them telling the llm the file is in data/file.txt because that would end up in data/data/file.txt

BenCookie95: We already specify that under the File Management heading though so maybe I can remove this section

mcpserver/README.md
BenCookie95: Probably not relevant to the mcp user, more of a dev note but I can remove this

mcpserver/tools/file_utils.go
BenCookie95: Thanks, I'll take a look

mcpserver/README.md
crspeller: Right, I think we just need to make that super clear somehow in here.

mcpserver/tools/file_utils.go
BenCookie95: I decided to leave this here, I thought it was messier have each caller pass the data directory to `fetchFileDataForLocal`

crspeller (approved):

PR #381, opened 2025-09-16
[MM-65675] Port OpenAI Integration to Official SDK

nickmisasi:
Closes: https://github.com/mattermost/mattermost-plugin-agents/pull/380

## Description

Ports the OpenAI implementation from sashabaranov/go-openai to openai/openai-go. 

Adds tests

Includes a few fixes as well:
- Dimensions are now passed explicitly to the embeddings provider creation, since they weren't before
- newEmbeddingsProvider calls the correct `newEmbeddings` function when in official OpenAI mode
- Removes MaxTokens in favour of MaxCompletionTokens

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -415,10 +415,8 @@ func (s *OpenAI) streamResultToChannels(params openai.ChatCompletionNewParams, l
 		case "":
 			// Not done yet, keep going
 		default:
-			if choice.FinishReason != "" {
-				fmt.Printf("Unknown finish reason: %s", choice.FinishReason)
-				return
-			}
+			fmt.Printf("Unknown finish reason: %s", choice.FinishReason)
+			return
 		}
 	}
 


openai/openai.go
```diff
@@ -486,16 +519,24 @@ func (s *OpenAI) ChatCompletionNoStream(request llm.CompletionRequest, opts ...l
 }
 
 func (s *OpenAI) Transcribe(file io.Reader) (*subtitles.Subtitles, error) {
-	resp, err := s.client.CreateTranscription(context.Background(), openaiClient.AudioRequest{
-		Model:    openaiClient.Whisper1,
-		Reader:   file,
-		FilePath: "input.mp3",
-		Format:   openaiClient.AudioResponseFormatVTT,
-	})
+	// Read the file into a buffer since the new API expects a filename
```
crspeller: I don't understand. Why is this buffer required?

openai/openai.go
```diff
@@ -444,36 +449,64 @@ func (s *OpenAI) createConfig(opts []llm.LanguageModelOption) llm.LanguageModelC
...
+	}
 }
 
 func (s *OpenAI) ChatCompletion(request llm.CompletionRequest, opts ...llm.LanguageModelOption) (*llm.TextStreamResult, error) {
-	openAIRequest := s.completionRequestFromConfig(s.createConfig(opts))
-	openAIRequest = modifyCompletionRequestWithRequest(openAIRequest, request)
-	openAIRequest.Stream = true
+	params := s.completionRequestFromConfig(s.createConfig(opts))
+	params = modifyCompletionRequestWithRequest(params, request)
+	// Streaming is enabled via the NewStreaming method, not a params field
```
crspeller: Probably unnecessary comment.

openai/openai.go
```diff
@@ -48,187 +51,224 @@ const (
...
+		params.Tools = toolsToOpenAITools(internalRequest.Context.Tools.GetTools())
 	}
-	return openAIRequest
+	return params
 }
 
-func toolsToOpenAITools(tools []llm.Tool) []openaiClient.Tool {
-	result := make([]openaiClient.Tool, 0, len(tools))
+// schemaToFunctionParameters converts a jsonschema.Schema to shared.FunctionParameters
+func schemaToFunctionParameters(schema *jsonschema.Schema) shared.FunctionParameters {
```
crspeller, line 152: Dealing with these function parameters is always so annoying!

crspeller (changes requested):
Nice! Just a couple things

openai/openai.go
```diff
@@ -48,187 +51,224 @@ const (
...
-			return clientConfig
-		},
-	)
+	// Azure configuration
+	opts := []option.RequestOption{
+		option.WithAPIKey(config.APIKey),
+		option.WithHTTPClient(httpClient),
+		option.WithBaseURL(strings.TrimSuffix(config.APIURL, "/")),
+		// For Azure, we need to use a different base URL pattern
+		// Azure expects: https://{your-resource-name}.openai.azure.com/openai/deployments/{deployment-id}
```
davidkrauser: **question:** is this comment in the right place?

openai/openai.go
```diff
@@ -48,187 +51,224 @@ const (
...
 
 func NewAzure(config Config, httpClient *http.Client) *OpenAI {
-	return newOpenAI(config, httpClient,
-		func(apiKey string) openaiClient.ClientConfig {
-			clientConfig := openaiClient.DefaultAzureConfig(apiKey, strings.TrimSuffix(config.APIURL, "/"))
-			clientConfig.APIVersion = "2024-06-01"
-			return clientConfig
-		},
-	)
+	// Azure configuration
```
davidkrauser: **nitpick:** we should remove this comment - it doesn't add any extra information in this context.

openai/openai.go
```diff
@@ -404,25 +391,43 @@ func (s *OpenAI) streamResultToChannels(request openaiClient.ChatCompletionReque
...
 			return
+		case "":
+			// Not done yet, keep going
 		default:
-			fmt.Printf("Unknown finish reason: %s", response.Choices[0].FinishReason)
-			return
+			if choice.FinishReason != "" {
+				fmt.Printf("Unknown finish reason: %s", choice.FinishReason)
+				return
+			}
```
davidkrauser: **question:** if the finish reason _is not_ set, we still want to return, right?

openai/openai.go
```diff
@@ -48,187 +51,224 @@ const (
...
-		},
-	)
 }
 
 // NewCompatibleEmbeddings creates a new OpenAI client configured only for embeddings functionality
 func NewCompatibleEmbeddings(config Config, httpClient *http.Client) *OpenAI {
 	if config.EmbeddingModel == "" {
-		config.EmbeddingModel = string(openaiClient.LargeEmbedding3)
-		config.EmbeddingDimentions = 3072
+		config.EmbeddingModel = openai.EmbeddingModelTextEmbedding3Large
```
davidkrauser, line 125: **suggestion:** do we want to make this configurable in the future?

nickmisasi: It is configurable, this is just setting a default if it's not configured

openai/openai.go
nickmisasi: If it's not set (ie, empty string) it means keep going - that's handled by the cause above this default one.

openai/openai.go
nickmisasi: Changed

openai/openai.go
davidkrauser: ah - I missed that. Why do we need the if check here if we know it is not an empty string?

nickmisasi: Removed

PR #382, opened 2025-09-16
Update PR template

nickmisasi:
## Description

Update PR template to be in-line with PE.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,5 +1,65 @@
...
+
+For an easier comparison of UI changes a table (template below) can be used.
+
+|  before  |  after  |
+|----|----|
+| <insert before screenshot here> | <insert after screenshot here> |
+
+-->
+
+#### Release Note
```
crspeller, line 33: Question: Does this automation work for plugin repos?
If not then maybe drop this section.

nickmisasi: it at least works in the cloud plugin repo. We might need to configure it first via webhooks though

PR #383, opened 2025-09-17
[MM-64686] Log LLM token usage tracking

davidkrauser:
#### Summary
The changes in this branch implement LLM token usage tracking that can be used to track the costs of AI interactions within the Mattermost agents plugin. This addresses the need to monitor and measure resource consumption across different language models. The implementation captures detailed token usage metrics including input tokens, output tokens, model types, and associated bot identities.

The structured logging format allows organizations to import these logs into their existing reporting and aggregation tools for comprehensive cost analysis and budget management.

#### Supported Providers
This change only supports Anthropic and OpenAI backends.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64686

#### Output Details
Token usage is logged to `token_usage.log` with content like the following:
```
{"timestamp":"2025-09-17 16:16:45.479 -04:00","level":"info","msg":"Token Usage","user_id":"iac1dy16n7bi7ycwhg5wfg4dac","team_id":"6y8ixw7tr3yztjgk3hip87bjuh","bot_username":"matty","input_tokens":416,"output_tokens":3,"total_tokens":419}
{"timestamp":"2025-09-17 16:16:48.957 -04:00","level":"info","msg":"Token Usage","user_id":"iac1dy16n7bi7ycwhg5wfg4dac","team_id":"6y8ixw7tr3yztjgk3hip87bjuh","bot_username":"matty","input_tokens":34,"output_tokens":5,"total_tokens":39}
```

### DIFF
--- a/llm/token_tracking.go
+++ b/llm/token_tracking.go
@@ -1,0 +1,135 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package llm
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/mattermost/mattermost/server/public/shared/mlog"
+)
+
+// TokenTrackingWrapper wraps a LanguageModel to track token usage in the database
+type TokenTrackingWrapper struct {
+	wrapped     LanguageModel
+	botUsername string
+	tokenLogger *mlog.Logger
+}
+
+// NewTokenLoggingWrapper creates a new wrapper that tracks token usage
+func NewTokenLoggingWrapper(wrapped LanguageModel, botUsername string) (*TokenTrackingWrapper, error) {
+	tokenLogger, err := createTokenLogger()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create token tracking wrapper: %w", err)
+	}
+	return &TokenTrackingWrapper{
+		wrapped:     wrapped,
+		botUsername: botUsername,
+		tokenLogger: tokenLogger,
+	}, nil
+}
+
+// createTokenLogger creates a dedicated logger for token usage metrics
+func createTokenLogger() (*mlog.Logger, error) {
+	logger, err := mlog.NewLogger()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create token logger: %w", err)
+	}
+
+	jsonTargetCfg := mlog.TargetCfg{
+		Type:   "file",
+		Format: "json",
+		Levels: []mlog.Level{mlog.LvlInfo, mlog.LvlDebug},
+	}
+	jsonFileOptions := map[string]interface{}{
+		"filename": "token_usage.log",
+	}
+	jsonOptions, err := json.Marshal(jsonFileOptions)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal json file options: %w", err)
+	}
+	jsonTargetCfg.Options = json.RawMessage(jsonOptions)
+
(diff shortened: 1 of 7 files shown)


llm/token_tracking.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/mattermost/mattermost/server/public/shared/mlog"
+)
+
+// TokenTrackingWrapper wraps a LanguageModel to track token usage in the database
```
crspeller: To a log file.

crspeller (changes requested):
Just one minor nit and the merge conflicts.

llm/token_tracking.go
davidkrauser: nice catch - I'll fix that :+1:

davidkrauser: Done on commit: aefdda8

llm/token_tracking.go
```diff
@@ -0,0 +1,131 @@
...
+
+	err = logger.ConfigureTargets(map[string]mlog.TargetCfg{
+		"token_usage": jsonTargetCfg,
+	}, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to configure token logger targets: %w", err)
+	}
+
+	return logger, nil
+}
```
nickmisasi, line 61: Do we need to add support for our AdvancedLoggingJSON custom log formatting? https://docs.mattermost.com/administration-guide/manage/logging.html#mattermost-logging

It's great to have the functionality in the first place, but, we've learned with other features (like compliance export) that users may expect this data in a format different from our hard coded one, and thus end up being unable to use the feature anyways

davidkrauser: > users may expect this data in a format different from our hard coded one

That's a good point - so in a47ec47, I replaced the custom one-off logger instance with the standard Mattermost logger. With that change, the admin can now specify custom configuration and targets for these logs (and won't be locked down to a specific file).

davidkrauser: I reverted the change - the plugin API doesn't allow for custom log levels that we can filter on. We'll need to add support for that to the plugin API if we want to enable more configurable logging for this.

I think we should, but will need to be a later PR (since that will require core Mattermost changes). WDYT?

nickmisasi: Agreed that we both should do it, and that we should not do it right now :) 

Will this eventual effort potentially provide us with some backing to add proper audit logging to plugins as well?

davidkrauser: @nickmisasi we've already got audit logging support over the plugin api: https://github.com/mattermost/mattermost/pull/31204 :slightly_smiling_face:

davidkrauser: Just pushed one more commit related to this - 064efef. This will ensure we roll these logs and compress them. Since this is on by default and isn't configurable, I don't want the logs to blow up disk space - so I'm following the default Mattermost configuration when it comes to rolling and compression.

@crspeller After thinking about this discussion more - I don't think we should mark this feature "done" until we can make the logging configurable via the advanced logging features in Mattermost. That will require a small change to Mattermost to enable setting a custom log level in the plugin api when logging. That log level will allow users to filter for these logs specifically (by that log level), and to configure options for those logs specifically (e.g. redirecting to a logging aggregation service, etc.). WDYT?

davidkrauser: We talked about this offline - and we're fine with the current behavior. The only configuration option we're adding for now is the ability to enable/disable logging entirely.

PR #387, opened 2025-09-25
[MM-65676] Migrate to OpenAI Responses API, Support for Reasoning Models

nickmisasi:
#### Summary
This PR adds support for the Responses API via OpenAI and OpenAI compatible LLMs. This model is in addition to support for the Completions API - configurable through the System Console. Completions API will still remain the default for now.

The Responses API gives us a lot of benefits that aren't present in the Completions API. 
1. Native tools - we can access Web Search, Code Interpreter, etc "for free"
2. Reasoning Summary - As part of the response stream, OpenAI sends reasoning a reasoning summary for us to display to the user. This 
#### Screenshots

Thinking panel collapsed, while reasoning is streaming:
image
Thinking panel expanded, while reasoning is streaming:
image
Thinking panel expanded, reasoning finished:
image
Thinking panel collapsed, reasoning finished:
image

#### Release Note

```release-note
Added support for OpenAI's Responses API
Added support for web search via OpenAI native tools (EXPERIMENTAL) - configurable on/off and requires Responses API use.
Added UX support for reasoning models
```

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -493,7 +493,6 @@ func (s *OpenAI) streamResponsesAPIToChannels(params openai.ChatCompletionNewPar
 	// Convert ChatCompletionNewParams to ResponseNewParams
 	responseParams := s.convertToResponseParams(params, llmContext)
 
-
 	// Create a streaming request
 	stream := s.client.Responses.NewStreaming(ctx, responseParams)
 	defer stream.Close()
@@ -559,7 +558,6 @@ func (s *OpenAI) streamResponsesAPIToChannels(params openai.ChatCompletionNewPar
 
 		// Ping the watchdog when we receive a response
 		watchdog <- struct{}{}
-
 
 		// Process event types
 


nickmisasi:
@asaadmahmood still working on that - if you're ok with the overall look and feel based on the screenshots, I'll finish cleaning up the PR and formally open it for review, hopefully by that point I have Spinwicks going

streaming/streaming.go
```diff
@@ -245,6 +263,27 @@ func (p *MMPostStreamService) StreamToPost(ctx context.Context, stream *llm.Text
 				}
 				p.sendPostStreamingUpdateEvent(post, post.Message)
 				return
+			case llm.EventTypeReasoning:
+				// Handle reasoning summary chunk - accumulate and stream
+				if reasoningChunk, ok := event.Value.(string); ok {
+					reasoningBuffer.WriteString(reasoningChunk)
```
crspeller, line 275: I think we need to buffer in the actual message because we want when the stop button is pressed to show the reasoning not lose it. 

(Also on streaming error, but that isn't handled well for the regular messages at the moment.)

nickmisasi: As it was the stop issue wasn't a problem because it wouldn't actually show that button until the reasoning phase was over. I've pushed an update to handle stop button and error, but also a few tweaks to the front-end so the stop button shows (along with the regenerate button if stopped during reasoning)

crspeller (approved):

openai/openai.go
```diff
@@ -450,6 +464,473 @@ func (s *OpenAI) streamResultToChannels(params openai.ChatCompletionNewParams, l
...
+		}
+	}
+
+	for stream.Next() {
+		event := stream.Current()
+
+		// Ping the watchdog when we receive a response
+		watchdog <- struct{}{}
+
+		// Debug: Log all event types to see what's coming from the API
```
BenCookie95: There are a few leftover comments from when you were debugging

BenCookie95 (approved):

openai/openai.go
nickmisasi: @claude can you clean up the leftover debug comments ben mentioned?

PR #390, opened 2025-09-30
[MM-65728] Block notifications from immediate bot replies in threads

davidkrauser:
#### Summary
Currently when you message an agent, the agent replies to your post in a thread and always triggers an immediate notification. With this PR, I've introduced a mechanism that blocks notifications related to threaded bot replies.This will silence the immediate notifications you get when interacting with agents.

For web and desktop notifications, we block all threaded bot replies that are created within 1 second of the parent post. Note that this mechanism will block notifications for all bot posts with this behavior. **So if a bot replies immediately to a top-level user post and creates a thread, you will never get a web/desktop notification for that post.** We could introduce an API to provide all known agent bot names to the webapp and filter on that list, but I'm not sure it's worth the extra complexity.

For push and email notifications, we block all threaded bot replies that are created by known AI Agent bots (with no time restriction). **This means you will never receive a push or email notification from an AI Agent bot if that post is created within a thread.**

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65728

#### Release Note
```release-note
Silence notifications when starting a new threaded conversation with an agent.
```


davidkrauser:
Talked with @crspeller, and we want this to block mobile and email notifications, too. Moving this to a draft while I add support for that.

server/post_cache.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+func (pc *postCache) getPost(api plugin.API, postID string) (*model.Post, error) {
+	if postID == "" {
+		return nil, fmt.Errorf("post ID cannot be empty")
+	}
+
+	// Check cache first (with lock)
+	pc.cacheLock.Lock()
```
crspeller: Should we use an RW mutex here? (Ignore if we remove the cache, read other comments first)

server/main.go
```diff
@@ -257,18 +276,76 @@ func (p *Plugin) MessageHasBeenUpdated(c *plugin.Context, newPost, oldPost *mode
...
+		return false
+	}
+
+	// Check if cache is initialized
+	if p.postCache == nil {
+		return false
+	}
+
+	// Get the post to check if it's from a bot
+	post, err := p.postCache.getPost(p.API, postID)
```
crspeller: It looks like we have `SenderId` in the `model.PushNotification`. Could we use that along with the already cached userIDs of the bots to only debounce our own bots and avoid the complex caching? https://github.com/mattermost/mattermost-plugin-agents/blob/00eb17aab573666ba9db21b3797eb5d58d053f23/bots/bots.go

davidkrauser: That's a good suggestion - I was still in the mindset of how we're filtering webapp-side and overlooked that :+1: 

My only worry is the case where we have AI Agent bots sending messages that notify a lot of people. For example, if one of these bots @-channel's in a group with many users, we will need to retrieve the associated post for each individual notification. Without some sort of cache, that will be a lot of RPC calls.

WDYT? Is this something we should be concerned about?

server/post_cache.go
davidkrauser: An RW mutex is a good suggestion. I'll make that change if we decide to keep the cache :+1:

server/main.go
crspeller: Right, because we still need to retrieve the parent post to calculate the timing. But since we are limiting ourselves now to only our own Agents bots, perhaps for simplicity we can just block all without respecting timings? 

I know we might have some situations in the future where we might want notifications, but I don't think there are any at the moment.

crspeller (changes requested):

server/main.go
davidkrauser: Unfortunately, we can't block all notifications generated by these bot accounts, otherwise they wouldn't be able to notify anyone (e.g. via direct message or mention). We also need to get the post to check if it is part of a thread (vs. a top-level message that we want notifications for). Do we want to support that use case?

crspeller: The `RootId` is also part of the `model.PushNotification` and the `model.EmailNotification` so we can check it there without retrieving any posts. For now I think we can just block all notifications regardless of timing.

davidkrauser: Done :+1:
I've also updated the PR description to document the new behavior.

server/main.go
```diff
@@ -264,3 +266,45 @@ func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Req
...
+		return pushNotification, ""
+	}
+
+	if p.shouldBlockBotReplyNotification(pushNotification.SenderId, pushNotification.RootId) {
+		return nil, "notification blocked: bot reply in thread"
+	}
+	return pushNotification, ""
+}
+
+func (p *Plugin) shouldBlockBotReplyNotification(senderID, rootID string) bool {
```
BenCookie95, line 299: This will block all bots across all plugins right? Should we check if the bot is part of the agents config?

davidkrauser: No - this function has a check to ensure we only block push and email notifications from bots that were created by this plugin:

https://github.com/mattermost/mattermost-plugin-agents/blob/72334d0562bf43b36361ea3b71a1ce3c4e70832e/server/main.go#L303-L306

On the webapp side, though, I don't have a list of these bots. So we block **all** desktop and web notifications if posted by **any** bot within 1 second of the parent. There're more details in the PR description:

> For web and desktop notifications, we block all threaded bot replies that are created within 1 second of the parent post. Note that this mechanism will block notifications for all bot posts with this behavior. **So if a bot replies immediately to a top-level user post and creates a thread, you will never get a web/desktop notification for that post.** We could introduce an API to provide all known agent bot names to the webapp and filter on that list, but I'm not sure it's worth the extra complexity.

BenCookie95: Ah ok, thank you!

.github/workflows/ci.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-go@v4
         with:
-          go-version: '1.23'
+          go-version: '1.24'
```
nickmisasi: nit: before you merge, can we change this to `go-version-file: go.mod` (see: https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file)

PR #391, opened 2025-10-02
In-memory transport for MCP

BenCookie95:
#### Summary
Added support for an in-memory Mattermost MCP server.

**Notes:**
- Using the user's existing session to create an MCP session. Not creating a new one right now, I had mostly finished this PR by the time we decided on the other avenue. I'll create a ticket to address it after.
- Integrated the embedded server into our existing MCP client logic. The `Client` struct now has a field to hold the embedded server configuration. `UserClients.clients` holds both the remote server and embedded server connections.
- When creating a connection to the embedded mcp server for a user, we have to pass the sessionID and sessionResolver into context at connection time. They live in context for the duration of the embedded session. I could have included the session token in the in the tool call metadata but putting it into context kept it more in line with how it's working currently on the mcpserver side. Having the sessionID in context is fine, we could put the actual token but I was just worried about tokens being dumped into logs if the context got logged somewhere.
- `llm.Context` also contains a sessionID and session resolver now.
- The admin api doesn't pass a session ID down to the embedded MCP server when doing tool discovery in the system console (the discovery endpoint obviously already checks for a valid session). Only reason for that is that we aren't passing the plugin context down to the api handlers. The embedded MCP server doesn't require a session for tool discovery right now anyway.
- I slightly changed the wording in system console to distinguish remote MCPs from our embedded Mattermost MCP.
- I had to change the logging in the mcpserver to use a new Logger interface that both logging systems could satisfy. I then created PluginAPILoggerAdapter that wraps pluginapi.LogService to implement the types.Logger interface. This is so the embedded MCP server logs are output in the same location as the plugin and in the same format. The previous setup allowed me to output to the same location but the formatting was off and the mcp server logs were being flushed into one single log message.

#### Release Note

```release-note
Add support for in-memory transport for the Agents plugin to utilize Mattermost MCP
```

### DIFF
--- a/webapp/src/components/system_console/mcp_servers.tsx
+++ b/webapp/src/components/system_console/mcp_servers.tsx
@@ -241,7 +241,7 @@ const MCPServers = ({mcpConfig, onChange}: Props) => {
         enabled: mcpConfig?.enabled || false,
         servers: Array.isArray(mcpConfig?.servers) ? mcpConfig.servers : [],
         embeddedServer: mcpConfig?.embeddedServer || {
-            enabled: false,
+            enabled: !mcpConfig?.enabled,
         },
         idleTimeoutMinutes: mcpConfig?.idleTimeoutMinutes || 30,
     };


BenCookie95:
@crspeller @nickmisasi I have a bunch of tests coming in a separate PR. It's another 2000 lines so I didn't want to overload this one with more code

mcp/client.go
```diff
@@ -33,6 +51,90 @@ type ServerConfig struct {
...
+	transport, err := c.server.CreateClientTransport(userID, sessionID, c.pluginAPI)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create in-memory transport: %w", err)
+	}
+
+	// Create MCP client
+	mcpClient := mcp.NewClient(
+		&mcp.Implementation{
+			Name:    "mattermost-agents-embedded",
+			Version: "1.0",
```
nickmisasi, line 86: Should we put this as 0.1 or something? 1.0 should be saved for a stable final release?

webapp/src/components/system_console/mcp_servers.tsx
```diff
@@ -333,12 +341,23 @@ const MCPServers = ({mcpConfig, onChange}: Props) => {
...
+                                        label={intl.formatMessage({defaultMessage: 'Enable Embedded Server'})}
+                                        value={config.embeddedServer.enabled}
+                                        onChange={(enabled) => onChange({
+                                            ...config,
+                                            embeddedServer: {
+                                                ...config.embeddedServer,
+                                                enabled,
+                                            },
+                                        })}
+                                        helpText={intl.formatMessage({defaultMessage: 'Enable the built-in Mattermost MCP server that provides tools for accessing channels, users, and posts without external OAuth setup.'})}
```
nickmisasi, line 354: We may want to better clarify what the embedded server is vs other options and what ramifications are of enabling it if any

api/api_admin.go
```diff
@@ -142,19 +142,49 @@ func (a *API) handleGetMCPTools(c *gin.Context) {
...
-	if !mcpConfig.Enabled || len(mcpConfig.Servers) == 0 {
+	// If MCP is not enabled, return empty response
+	if !mcpConfig.Enabled {
 		c.JSON(http.StatusOK, MCPToolsResponse{
 			Servers: []MCPServerInfo{},
 		})
 		return
 	}
 
+	// Calculate total capacity (remote servers + embedded server if enabled)
```
crspeller: Probably excessive memory saving. Maybe just len(mcpConfig.Servers)+1 without the check?

api/api_admin.go
```diff
@@ -187,9 +217,30 @@ func (a *API) handleGetMCPTools(c *gin.Context) {
 	c.JSON(http.StatusOK, response)
 }
 
-// discoverServerTools connects to a single MCP server and discovers its tools
-func (a *API) discoverServerTools(ctx context.Context, requestingAdminID string, serverConfig mcp.ServerConfig) ([]MCPToolInfo, error) {
-	toolInfos, err := mcp.DiscoverServerTools(ctx, requestingAdminID, serverConfig, a.pluginAPI.Log, a.mcpClientManager.GetOAuthManager())
+// discoverRemoteServerTools connects to a single remote MCP server and discovers its tools
+func (a *API) discoverRemoteServerTools(ctx context.Context, requestingAdminID string, serverConfig mcp.ServerConfig) ([]MCPToolInfo, error) {
```
crspeller: Why `requestingAdminID` and not `requestingUserId`?

mcpserver/types/logger.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package types
+
+// Logger is a minimal logging interface that can be satisfied by both
+// standalone (mlog) and embedded (pluginapi) logging implementations
+type Logger interface {
```
crspeller: Not a big fan of a `types` package. Maybe just put this where it's needed?

webapp/src/components/system_console/mcp_servers.tsx
```diff
@@ -333,12 +341,23 @@ const MCPServers = ({mcpConfig, onChange}: Props) => {
                                         }}
                                         helptext={intl.formatMessage({defaultMessage: 'How long to keep an inactive user connection open before closing it automatically. Lower values save resources, higher values improve response times.'})}
                                     />
+                                    <BooleanItem
+                                        label={intl.formatMessage({defaultMessage: 'Enable Embedded Server'})}
```
crspeller, line 345: Let's invert this so that it is on by default when you enable MCP.

mcp/mcp.go
```diff
@@ -29,22 +30,30 @@ type Errors struct {
 	Errors         []error             // Generic errors (connection, config, etc.)
 }
 
+// EmbeddedServerConfig contains configuration for the embedded MCP server
+type EmbeddedServerConfig struct {
+	Enabled bool `json:"enabled"`
+
+	// Future configuration options can be added here
```
crspeller: Nit: Not helpful comment.

server/main.go
```diff
@@ -144,7 +145,19 @@ func (p *Plugin) OnActivate() error {
 	manifestID := manifest.Id
 	oauthCallbackURL := fmt.Sprintf("%s/plugins/%s/oauth/callback", *siteURL, manifestID)
 
-	mcpClientManager := mcp.NewClientManager(p.configuration.MCP(), pluginAPI.Log, pluginAPI, mcp.NewOAuthManager(mmClient, oauthCallbackURL))
+	// Create embedded MCP server if enabled
+	if p.configuration.MCP().EmbeddedServer.Enabled {
```
crspeller, line 151: This needs a `p.configuration.RegisterUpdateListener` otherwise it will require a server restart to update the enabled setting.

crspeller (changes requested):
Looking good! Just a few things.

mcp/client.go
BenCookie95: Actually I think this is wrong, I think it's the MCP version you are support. We do the same thing when connecting to SSE servers. I think the version refers to the MCP version right? What version are we supporting? Is it the oldest one `2024-11-05`?

BenCookie95: @crspeller What's the min version of MCP we are supporting

webapp/src/components/system_console/mcp_servers.tsx
BenCookie95: I guess my only thought with this is that if someone upgrades their agents plugin and MCP is enabled, this will get auto enabled too and the admin might not be aware. Any thoughts on that?

mcpserver/types/logger.go
BenCookie95: I removed the types package and put the logger in it's package as it's needed in a lot of inter connected places. AccessMode is now part of the tools package

webapp/src/components/system_console/mcp_servers.tsx
crspeller: Let's have it so that when someone on a new server enables MCP that the embedded server is enabled by default. For existing MCP enabled servers we can keep it off by default. So this is just changing the system console behavior.

mcp/client.go
crspeller: Not sure, but checking the docs this does actually appear to be the version of the server not the MCP version.

llm/context.go
```diff
@@ -27,6 +27,11 @@ type Context struct {
 	// User that is making the request
 	RequestingUser *model.User
 
+	// Session information for authentication
+	SessionID string
+	// SessionResolver provides secure access to session token when needed
+	SessionResolver func() (string, error)
+
```
jupenur, line 34: This is a very odd place to store this information. These are never meant to be included in the LLM prompt, right?

BenCookie95: At the time it was the easiest way to get this down the chain to the places I needed it but it's removed in the other PR where we are creating the session token explicitly for the MCP user

PR #395, opened 2025-10-06
[MM-65976] Add web search annotation support 

nickmisasi:
#### Summary
Adds support for web search annotations 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65976

#### Screenshots
Anthropic
image
OpenAI
image

#### Release Note

```release-note
Support for annotations from native search tools
```

### DIFF
--- a/webapp/src/components/post_text.tsx
+++ b/webapp/src/components/post_text.tsx
@@ -9,11 +9,10 @@ import {GlobalState} from '@mattermost/types/store';
 import {Channel} from '@mattermost/types/channels';
 import {Team} from '@mattermost/types/teams';
 
-import {insertAnnotationMarkers, replaceCitationMarkers} from './citations/citation_processor';
-import {Annotation} from './citations/types';
-
 import manifest from '@/manifest';
 
+import {insertAnnotationMarkers, replaceCitationMarkers} from './citations/citation_processor';
+import {Annotation} from './citations/types';
 
 export type ChannelNamesMap = {
     [name: string]: {


anthropic/anthropic.go
```diff
@@ -325,6 +344,67 @@ func (a *Anthropic) streamChatWithTools(state messageState) {
 	}
 }
 
+// extractAnnotations extracts citations from Anthropic's message content blocks
+func (a *Anthropic) extractAnnotations(message anthropicSDK.Message) []llm.Annotation {
```
BenCookie95, line 348: Should we unit test this with some mock data?

openai/openai.go
```diff
@@ -502,6 +503,13 @@ func (s *OpenAI) streamResponsesAPIToChannels(params openai.ChatCompletionNewPar
 	var currentToolIndex int
 	var reasoningSummaryBuffer strings.Builder
 	var reasoningComplete bool // Track if we've sent the complete reasoning
+	// var citationAcc = newCitationAccumulator()
```
BenCookie95:
```suggestion
```

webapp/src/components/post_text.tsx
```diff
@@ -83,21 +87,32 @@ const PostText = (props: Props) => {
...
+        formatText(processedMessage, markdownOptions),
         messageHtmlToComponentOptions,
     );
 
     if (!text) {
         return <TextContainer showCursor={props.showCursor}>{<p/>}</TextContainer>;
     }
 
+    // Post-process the rendered JSX to replace citation markers with React components
+    const processedText = props.annotations && props.annotations.length > 0 ?
```
BenCookie95, line 109: Would it be worth putting this in a useCallback?

BenCookie95 (changes requested):
Just a few small things

webapp/src/components/post_text.tsx
nickmisasi: I don't think we'd see a material decrease because the annotations if present will be non-zero at first mount of the component, and then the text that's being streamed will cause the need to re-run the `replacecitationmarkers` since the citation can be placed inline anywhere in the response

anthropic/anthropic.go
nickmisasi: added

BenCookie95 (approved):

llm/annotations.go
```diff
@@ -0,0 +1,24 @@
...
+
+package llm
+
+// AnnotationType represents different types of annotations
+type AnnotationType string
+
+const (
+	// AnnotationTypeURLCitation represents a web search citation
+	AnnotationTypeURLCitation AnnotationType = "url_citation"
+	// Future annotation types can be added here (e.g., "tool_search", "vector_search")
```
crspeller: Unhelpful comment.

webapp/src/components/llmbot_post.tsx
```diff
@@ -281,6 +297,18 @@ export const LLMBotPost = (props: Props) => {
...
+            const persistedAnnotations = props.post.props?.annotations || '';
+            if (persistedAnnotations) {
+                try {
+                    setAnnotations(JSON.parse(persistedAnnotations));
+                } catch (error) {
+                    setAnnotations([]);
+                }
+            } else {
+                setAnnotations([]);
+            }
```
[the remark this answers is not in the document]
crspeller: Not a bad idea.

llm/annotations.go
nickmisasi:
```suggestion
```

PR #403, opened 2025-10-08
Mark MCP posts as AI generated

BenCookie95:
#### Summary
We need to mark posts as AI generated when they are coming from an LLM. For the embedded server we want to pass down the ID of the bot that was used when calling the tool. For other server types we just need to get the current UserID. I added a way to disable this for http/sse and stdio. They have use cases for data generation where you won't want to mark posts as AI generated.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65988

#### Screenshots
Screenshot 2025-10-16 at 10 44 07 AM

#### Release Note

```release-note
Mark MCP posts as AI generated
```


llmcontext/llm_context.go
```diff
@@ -210,5 +210,9 @@ func (b *Builder) WithLLMContextBot(bot *bots.Bot) llm.ContextOption {
 		c.BotUsername = bot.GetConfig().Name
 		c.BotModel = bot.GetConfig().Service.DefaultModel
 		c.CustomInstructions = bot.GetConfig().CustomInstructions
+		// Set the bot user ID for AI-generated content tracking
+		if bot.GetMMBot() != nil {
+			c.BotUserID = bot.GetMMBot().UserId
+		}
```
crspeller:
```suggestion
		if mmbot := bot.GetMMBot(); mmbot != nil {
			c.BotUserID = mmbot.UserId
		}
```

mcpserver/config.go
```diff
@@ -3,12 +3,11 @@
...
-// ServerConfig interface defines common configuration methods for all server types
-type ServerConfig interface {
-	GetMMServerURL() string
-	GetMMInternalServerURL() string
-	GetDevMode() bool
-}
+import "github.com/mattermost/mattermost-plugin-ai/mcpserver/types"
+
+// ServerConfig is an alias for the interface in the types package
+// This avoids circular dependencies while maintaining backward compatibility
```
crspeller: How? I don't understand. Shouldn't this only need to be here?

BenCookie95:
> So these would get marked as generated by AI, with the current user as the AI provider? Will that be strange from a UX perspective?

No in the UI it would just say `AI generated` when you hover over the AI star icon instead of `Message posted by @{username}`.  We don't know what they were using and I don't think there's a reliable way to tell.

> Would you mind expanding on this in the description a bit? Which use-cases would this be for?

This would be for when we generate data for demo servers. We don't want all the posts being marked as AI generated

mcpserver/config.go
BenCookie95: Missed that, removed it. `ServerConfig` is defined in the types package, I can't put it in mcpserver/config.go because that file is part of the mcpserver package and it already imports the tools package. The tools package needs the `ServerConfig` as well

PR #404, opened 2025-10-09
Built in web search

nickmisasi:
#### Release Note

```release-note

```


config/config.go
```diff
@@ -26,6 +26,30 @@ type Config struct {
...
 	MCP                      mcp.Config                       `json:"mcp"`
+	WebSearch                WebSearchConfig                  `json:"webSearch"`
+}
+
+type WebSearchConfig struct {
+	Enabled         bool                  `json:"enabled"`
+	Provider        string                `json:"provider"`
+	Google          WebSearchGoogleConfig `json:"google"`
+	Brave           WebSearchBraveConfig  `json:"brave"`
+	DomainBlacklist []string              `json:"domainBlacklist"`
```
esarafianou: nit: `DomainDenylist` - The industry is moving away from the whitelist/blacklist terminology to allowlist/denylist

docs/admin_guide.md
```diff
@@ -103,6 +103,100 @@ Text input in the custom instructions field is included in the prompt for every
...
+
+Google Custom Search provides access to Google's search index but has several important limitations:
+
+- **Not a first-party integration**: Mattermost uses Google's Custom Search API, which doesn't provide the same quality of results as searching directly on google.com
+- **Relies on web scraping**: After receiving search results, Mattermost must scrape web pages to extract content for the agent. Many websites block automated scraping or return limited content to bots
+- **Rate limits**: Google Custom Search has strict daily quota limits
+
+Due to these limitations, Google Custom Search may not always provide optimal results for agent queries.
+
+##### Brave Search (Recommended)
```
esarafianou, line 133: nit: It'd be better to lead with the recommended option and have Google custom search second.

webapp/src/components/system_console/web_search/web_search_panel.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    helpText={intl.formatMessage({defaultMessage: 'Allow agents to call Mattermost\'s built-in web search tool. If your LLM already provides native web search support, leave this disabled.'})}
+                />
+                <SelectionItem
+                    label={intl.formatMessage({defaultMessage: 'Provider'})}
+                    value={value.provider}
+                    onChange={(e) => handleUpdate({provider: e.target.value})}
+                    disabled={!value.enabled}
+                >
+                    <SelectionItemOption value='google'>{'Google Custom Search'}</SelectionItemOption>
+                    <SelectionItemOption value='brave'>{'Brave Search'}</SelectionItemOption>
```
esarafianou, line 78: I think we should have a helptext to highlight Bleve ToS: 

_Ensure you subscribe to Brave's Pro AI plan when using this feature. Using Brave's regular Search API (non-AI tier) violates Brave's Terms of Service and may result in account suspension._

websearch/google.go
```diff
@@ -0,0 +1,112 @@
...
+	values.Set("key", g.apiKey)
+	values.Set("cx", g.searchEngineID)
+	values.Set("q", query)
+	values.Set("num", strconv.Itoa(limit))
+	req.URL.RawQuery = values.Encode()
+	req.Header.Set("Accept", "application/json")
+
+	client := g.httpClient
+	if client == nil {
+		client = http.DefaultClient
```
esarafianou: Considering that `untrustedHTTPClient` is passed as the client to `NewWebSearchService`, the `client == nil` check should never really trigger. Instead of falling back to `defaultClient`, which lacks SSRF protections,  can we return an error?

websearch/brave.go
```diff
@@ -0,0 +1,278 @@
...
+	values := url.Values{}
+	values.Set("q", query)
+	values.Set("summary", "1")
+	req.URL.RawQuery = values.Encode()
+	req.Header.Set("X-Subscription-Token", b.apiKey)
+	req.Header.Set("Accept", "application/json")
+
+	client := b.httpClient
+	if client == nil {
+		client = http.DefaultClient
```
esarafianou: Same comment as with google on returning an error instead of using a client without SSRF protections

mmtools/web_search.go
```diff
@@ -0,0 +1,951 @@
...
+	// Security check 2: Check if the domain is blacklisted
+	cfg := s.cfgGetter()
+	if cfg != nil && isBlacklisted(pageURL, cfg.WebSearch.DomainBlacklist) {
+		s.logWarn("source fetch blocked by domain blacklist", "url", pageURL)
+		return "this domain is blocked by the administrator's configuration", errors.New("domain blacklisted")
+	}
+
+	client := s.httpClient
+	if client == nil {
+		client = http.DefaultClient
```
esarafianou: Same comment as in the google search snippet on lack of SSRF protection.

openai/openai.go
```diff
@@ -522,6 +522,7 @@ func (s *OpenAI) streamResponsesAPIToChannels(params openai.ChatCompletionNewPar
 	var currentToolIndex int
 	var reasoningSummaryBuffer strings.Builder
 	var reasoningComplete bool // Track if we've sent the complete reasoning
+	// var citationAcc = newCitationAccumulator()
```
esarafianou: Probably a development leftover?

mmtools/web_search.go
```diff
@@ -0,0 +1,951 @@
...
+	WebSearchAllowedURLsKey = "mm_web_search_allowed_urls"
+	// WebSearchExecutedQueriesKey is the key used to track which queries have been executed
+	WebSearchExecutedQueriesKey = "mm_web_search_executed_queries"
+	// WebSearchCountKey is the key used to track the number of searches executed
+	WebSearchCountKey = "mm_web_search_count"
+
+	minQueryLength       = 3
+	maxWebSearches       = 3
+	WebSearchDescription = "Perform a live web search using Google's Custom Search API. Use this tool to retrieve current information. Keep your search queries generic and concise according to the user's ask. Do not pass this tool URLs. You are limited to 3 searches per conversation - use them wisely. DO NOT repeat a search query you have already executed. In your final answer, cite sources using the exact format !!CITE1!!, !!CITE2!!, etc. These markers will be automatically converted to clickable citation links. Do NOT include URLs directly in your response text. Instead of creating a new search, refer to previous 'Live web search results' - only create a new search if the previous results are not relevant."
```
esarafianou: Shouldn't this prompt be generic and not mention Google's custom search API since we support more web search providers?

conversations/conversations.go
```diff
@@ -144,23 +152,48 @@ func (c *Conversations) ProcessUserRequestWithContext(bot *bots.Bot, postingUser
...
 
+	// If web search context wasn't found, initialize fresh tracking
+	if llmContext.Parameters == nil {
+		llmContext.Parameters = make(map[string]interface{})
+	}
+	if _, hasCount := llmContext.Parameters[mmtools.WebSearchCountKey]; !hasCount {
+		llmContext.Parameters[mmtools.WebSearchCountKey] = 0
+	}
+	if _, hasQueries := llmContext.Parameters[mmtools.WebSearchExecutedQueriesKey]; !hasQueries {
+		llmContext.Parameters[mmtools.WebSearchExecutedQueriesKey] = []string{}
```
esarafianou, line 190: Shouldn't we check if the WebSearch tool is available before initializing with fresh parameters? 0/5, just a recommendation to ensure it's clear in the code

mmtools/web_search.go
```diff
@@ -0,0 +1,951 @@
...
+	if err != nil {
+		s.logError("failed to read source fetch response", "error", err, "url", pageURL)
+		return "unable to read the response", err
+	}
+
+	// Extract content
+	var textContent string
+	parsedURL, parseErr := url.Parse(pageURL)
+	if parseErr == nil {
+		article, readErr := readability.FromReader(bytes.NewReader(body), parsedURL)
```
esarafianou, line 536: Leaving a comment as a reminder to:
1. Check if we could replace go-readability with docconv (used in the monorepo) which implements jusText for HTML. 
2. Otherwise, check if we could a bit more defensive to crashes and/or limit the download size to less than 30MB without impacting usability.

mmtools/web_search.go
```diff
@@ -0,0 +1,951 @@
...
+	builder.WriteString("2. Cite sources using !!CITE#!! format based on the numbered list provided above.\n")
+	if matchedResult != nil {
+		builder.WriteString(fmt.Sprintf("3. Use !!CITE%d!! when citing information from this fetched source.\n", matchedResult.Index))
+	}
+
+	return builder.String()
+}
+
+// wrapUntrustedContent wraps external web content with security warnings to prevent prompt injection
+func wrapUntrustedContent(content string) string {
```
esarafianou: This function is currently used only in tests. If we need this as standalone function for testing, shouldn't we reuse it in the `formatSummarizedContent` above? Not that that it's slighly different compared to what we have in lines 655-670

mmtools/web_search.go
```diff
@@ -0,0 +1,951 @@
...
+	// If there's a pre-formatted answer (e.g., from Brave), include it with special instructions
+	if searchResp.Answer != "" {
+		builder.WriteString("\nSummary:\n")
+		builder.WriteString("The following summary was generated by the search provider and already includes properly formatted citations (!!CITE#!!).\n")
+		builder.WriteString("You can use this summary directly in your response, preserving all !!CITE#!! markers exactly as they appear.\n")
+		builder.WriteString("You may also add additional citations from the sources below if needed.\n\n")
+		builder.WriteString(searchResp.Answer)
+		builder.WriteString("\n\n")
+		builder.WriteString("IMPORTANT: When using the summary above, preserve all !!CITE#!! markers exactly as written. You can also cite sources directly using !!CITE1!!, !!CITE2!!, etc.\n\n")
+	} else {
```
esarafianou: Suggestion for summary prompt injection checks:

```suggestion
		if unsafe, reason := isSummaryUnsafe(searchResp.Answer); unsafe {
			s.logWarn("provider summary rejected", "reason", reason)
			builder.WriteString("Provider summary was discarded due to safety checks. Use the sources below.\n\n")
			searchResp.Answer = ""
		} else {
			builder.WriteString("\nSummary:\n")
			builder.WriteString("The following summary was generated by the search provider and already includes properly formatted citations (!!CITE#!!).\n")
			builder.WriteString("You can use this summary directly in your response, preserving all !!CITE#!! markers exactly as they appear.\n")
			builder.WriteString("You may also add additional citations from the sources below if needed.\n\n")
			builder.WriteString(searchResp.Answer)
			builder.WriteString("\n\n")
			builder.WriteString("IMPORTANT: When using the summary above, preserve all !!CITE#!! markers exactly as written. You can also cite sources directly using !!CITE1!!, !!CITE2!!, etc.\n\n")
		}
	}

	if searchResp.Answer == "" {
```

Where `isSummaryUnsafe` is:

```go
// isSummaryUnsafe performs lightweight heuristic checks to decide if a summary
// should be rejected before it reaches the main agent.
func isSummaryUnsafe(summary string) (bool, string) {
	lower := strings.ToLower(summary)
	indicators := []string{
		"system:", "assistant:", "user:", "developer:", "role:",
		"function_call", "tool_calls", "\"tool_calls\":", "\"function\":", "\"arguments\":", "\"name\":", "\"type\":\"function\"",
		"ignore previous", "ignore all previous", "disregard", "forget previous", "reset instructions",
		"act as", "you are now", "system prompt",
		"run this", "execute this",
	}

	for _, phrase := range indicators {
		if strings.Contains(lower, phrase) {
			return true, "contains directive: " + phrase
		}
	}

	return false, ""
}

```
(open to update the indicators list)

mmtools/web_search.go
```diff
@@ -0,0 +1,951 @@
...
+		return "fetched page contained no readable content", nil
+	}
+
+	// Perform recursive summarization
+	summary, err := s.summarizeContent(bot, textContent)
+	if err != nil {
+		s.logWarn("recursive summarization failed, falling back to raw content with warnings", "error", err)
+		return s.wrapSourceContentWithContext(textContent, matchedResult, llmContext), nil
+	}
+
```
esarafianou, line 564: If we move forward with the `isSummaryUnsafe`, we can add it here too:

```suggestion
	if unsafe, reason := isSummaryUnsafe(summary); unsafe {
		s.logWarn("summarized content rejected", "reason", reason)
		return s.wrapSourceContentWithContext(summary, matchedResult, llmContext), nil
	}
	
```

conversations/conversations.go
nickmisasi: We don't have access to the Configuration struct at this level and I'm not sure it's worth the refactor to save some small overhead of empty parameters

mmtools/web_search.go
nickmisasi: I've added the limit of 30MB. 

In a separate branch, I did the switch to docconv, but the results weren't very good. Routinely the fetched page would be converted into something tiny, missing tons of context from that page. It was really noticeable in search results from places like Reddit, where no matter how much I played around with tuning the different parameters like link density, docconv would always strip out a post's comments. While Reddit isn't always the greatest source, forum-type results are pretty common in google searches, and it'll be important to be able to parse them.

mmtools/web_search.go
nickmisasi: I've removed this function as it was unused in production logic (which uses the more context-aware `wrapSourceContentWithContext` instead) and was only being called by its own test.

mmtools/web_search.go
```diff
@@ -0,0 +1,986 @@
...
+		"act as", "you are now", "system prompt",
+		"run this", "execute this",
+	}
+
+	for _, phrase := range indicators {
+		if strings.Contains(lower, phrase) {
+			return true, "contains directive: " + phrase
+		}
+	}
+
```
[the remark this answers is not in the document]
nickmisasi: @esarafianou I was a bit worried about this one myself. Thoughts?

conversations/conversations.go
esarafianou: Agreed, no need.

mmtools/web_search.go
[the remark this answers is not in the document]
esarafianou: You're right that false positives are possible. The examples provided don't quite illustrate the issue: `system:` won't match "system administrator" (no colon) and `tool_calls` differs from "tool calls" (no underscore). That said, false positives could still occur in other scenarios.

> Consider adding context-aware checks or allowing certain patterns when they appear in technical contexts to reduce false positives while maintaining security.

What copilot suggests would be a significantly larger undertaking.

How about start with the current implementation (I'm open to refining the list) and monitor for false positives in practice. 

Do we have a mechanism to track/audit possible false positives ? That would allow us to make a more informed decision after some usage.

esarafianou (approved):

mmtools/web_search.go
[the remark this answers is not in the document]
nickmisasi: I'm cool with trying this out for now. Recent behaviour from Google's search results seems to have more relevant information in snippets, so I've found the agent not needing to leverage this function a lot now anyways. 

There isn't a way to track/audit in a way that we can get that data back to us. Rudder would have been the way, but it's gone now.

mmtools/web_search.go
```diff
@@ -0,0 +1,971 @@
...
+	var builder strings.Builder
+	builder.WriteString(fmt.Sprintf("Live web search results for \"%s\":\n", query))
+	remainingSearches := maxWebSearches - searchCount
+	builder.WriteString(fmt.Sprintf("(Search %d of %d - %d searches remaining)\n", searchCount, maxWebSearches, remainingSearches))
+
+	// If there's a pre-formatted answer (e.g., from Brave), include it with special instructions
+	if searchResp.Answer != "" {
+		if unsafe, reason := isSummaryUnsafe(searchResp.Answer); unsafe {
```
crspeller: Why have this check? How is this less safe then anything else we are doing here?

mmtools/web_search.go
```diff
@@ -0,0 +1,971 @@
...
+	}
+
+	return builder.String()
+}
+
+// isSummaryUnsafe performs lightweight heuristic checks to decide if a summary
+// should be rejected before it reaches the main agent.
+func isSummaryUnsafe(summary string) (bool, string) {
+	lower := strings.ToLower(summary)
+	indicators := []string{
```
crspeller: I think my thoughts on doing stuff like this is probably more harm then good. Any text that happens to have the phrase "If I were to run this race it would" would be rejected for no reason. 

Detecting prompt injections like this is an unsolved problem that I don't think we want to tackle here.

conversations/tool_handling.go
```diff
@@ -12,10 +12,103 @@ import (
...
 	"github.com/mattermost/mattermost-plugin-ai/llm"
 	"github.com/mattermost/mattermost-plugin-ai/mmapi"
+	"github.com/mattermost/mattermost-plugin-ai/mmtools"
 	"github.com/mattermost/mattermost-plugin-ai/streaming"
 	"github.com/mattermost/mattermost/server/public/model"
 )
 
+// extractWebSearchContext retrieves web search context from the thread
+// The context may be stored on a previous post if multiple tool calls occurred
+func (c *Conversations) extractWebSearchContext(currentPost *model.Post) map[string]interface{} {
```
crspeller, line 22: Maybe we should store this context in the DB. (Not this PR, something for V2 changes?)

prompts/standard_personality_without_locale.tmpl
```diff
@@ -5,6 +5,8 @@ Users may refer to you as {{.BotName}} or mention you with your username @{{.Bot
 
 {{.BotName}} does not know the full extent of what they can or cannot do. When asked about capabilities, {{.BotName}} should only mention what they have been explicitly told they can do (such as tools they can access). IMPORTANT: When asked about capabilities, {{.BotName}} must always refer users to the documentation at docs.mattermost.com for complete information.
 
+{{.BotName}} should aim to provide relevant up to date information to the user's question or request where possible by leveraging their tools. However, {{.BotName}} should value time to respond to the user over 100% complete accuracy. When using the WebSearch tool, {{.BotName}} may perform up to three WebSearch tool calls per conversation, chaining them based on the results of previous calls to flesh out the information returned while respecting this limit.
```
crspeller, line 8: Maybe this should be part of the tool description as it most of the time will be disabled?

mmtools/web_search.go
```diff
@@ -0,0 +1,971 @@
...
+	flat := make([]WebSearchResult, 0)
+	for _, value := range values {
+		flat = append(flat, value.Results...)
+	}
+
+	return flat
+}
+
+// DecorateStreamWithAnnotations attaches annotation events based on search results to the provided stream.
+func DecorateStreamWithAnnotations(result *llm.TextStreamResult, searchData []WebSearchContextValue, logger WebSearchLog) *llm.TextStreamResult {
```
crspeller, line 786: This seriously complicates streaming here. Is there some other approach we could take? What about handling this client side?

nickmisasi: The primary use for this is around multi-turn conversations with many web searches, or when WebSearch causes WebSearchFetchSource to be called. Say you started with 

"What are the benefits of React 19"

It responds citing 3 sources. Those sources are only stored on the props for that post response. If you have a follow up question 

"How does that contrast to Vue?"

You need to re-incorporate the citations from the first question, so that the LLM can still reference them, and use the correct indices when it places its citation markers.

There would be ways that the front-end could process this, but it'd mean 
1. We'd have to cascade that processing out to both Webapp and Mobile
2. There could be races, delays, or more possibility for failures (network, etc) to persist to the backend which drops/breaks citations

This emittance of the EventTypeAnnotation fits the same pattern used by the native search for OpenAI and Anthropic

I think ideally in a future iteration (v2?) we can start to store context at the thread level, rather than post props level, I think it would open a lot of opportunities for simplification.

prompts/standard_personality_without_locale.tmpl
nickmisasi: Having these instructions in the Tool descriptions wasn't enough to steer the agent, I was only able to get it to actually follow these directions by having them in the system prompt. It was especially problematic with OpenAI that would think itself into 10000 web search calls of varying wording with the same results

PR #409, opened 2025-10-14
Add config migration for service/bot separation

davidkrauser:
This commit introduces the infrastructure needed to support separating LLM service configurations from bot definitions, allowing services to be defined once and referenced by multiple bots.

Changes include:
- Update BotConfig structure to support both embedded services (deprecated) and service references via ServiceID field
- Add migration framework with cluster mutex locking to handle multi-instance deployments safely
- Implement service extraction and deduplication logic that will enable migrating embedded service configs to top-level service definitions
- Add comprehensive test suite covering various migration scenarios including deduplication, error handling, and edge cases
- Update all bot-related code to work with Service as a pointer type

The service separation migration is prepared but will not activate in current deployments. That will come in a future commit.


llm/configuration.go
```diff
@@ -48,11 +49,17 @@ const (
 )
 
 type BotConfig struct {
-	ID                 string             `json:"id"`
-	Name               string             `json:"name"`
-	DisplayName        string             `json:"displayName"`
-	CustomInstructions string             `json:"customInstructions"`
-	Service            ServiceConfig      `json:"service"`
+	ID                 string `json:"id"`
```
crspeller, line 52: I don't think bots need IDs for the bots on the server side as the Name must be unique. Maybe remove?

server/main.go
```diff
@@ -69,13 +70,32 @@ func (p *Plugin) OnActivate() error {
 		PluginVersion:  manifest.Version, // Manifest imported from manifest.go which is generated by the build process
 	})
 
-	updated, newCfg, err := migrateServicesToBots(p.API, pluginAPI, *p.configuration.Config())
+	currentCfg, changed, err := runAllMigrations(p.API, pluginAPI, *p.configuration.Config())
```
crspeller: nit: `updatedConfig` might be more clear. `current` could mean the currently persisted config which is not this.

davidkrauser: That's fair - I've changed this to: `potentiallyUpdatedConfig, wasUpdated, err := /.../`

llm/configuration.go
davidkrauser: I left this in since we had it before. Can we leave that to a future change to simplify this?

crspeller (approved):

server/migrations.go
```diff
@@ -51,46 +168,84 @@ func migrateServicesToBots(mutexAPI cluster.MutexPluginAPI, pluginAPI *pluginapi
...
+	if _, kvErr := pluginAPI.KV.Set("migrate_services_to_bots_done", true); kvErr != nil {
+		pluginAPI.Log.Error("failed to mark migration as done", "key", "migrate_services_to_bots_done", "error", kvErr)
+	}
+
+	return true, *existingConfig, nil
+}
+
+// runAllMigrations executes all migrations under a single mutex to prevent race conditions
+// in multi-instance deployments. Returns the final configuration and any errors encountered.
+func runAllMigrations(mutexAPI cluster.MutexPluginAPI, pluginAPI *pluginapi.Client, cfg config.Config) (config.Config, bool, error) {
```
BenCookie95: Is `runAllMigrations` the right name for this? Do we anticipate adding more migrations to this function or should we call it something like `runServiceBotSeparationMigrations`

server/migrations.go
```diff
@@ -51,46 +168,84 @@ func migrateServicesToBots(mutexAPI cluster.MutexPluginAPI, pluginAPI *pluginapi
...
+			DisplayName: displayName,
+			ServiceID:   service.ID,
 		})
 	}
 
-	// If there is one bot then give it the standard name
-	if len(existingConfig.Bots) == 1 {
-		existingConfig.Bots[0].Name = "ai"
-		existingConfig.Bots[0].DisplayName = "Agents"
+	if _, kvErr := pluginAPI.KV.Set("migrate_services_to_bots_done", true); kvErr != nil {
```
BenCookie95: Wouldn't this be troublesome because we aren't persisting the config before we set this? If there is a failure persisting the config in `main.go` it won't run this migration again

server/main.go
```diff
@@ -69,13 +70,32 @@ func (p *Plugin) OnActivate() error {
...
 	if err != nil {
-		pluginAPI.Log.Error("failed to migrate services to bots", "error", err)
-		// Don't fail on migration errors
-	}
-	if updated && err == nil {
-		p.configuration.Update(&newCfg)
+		// Log error but don't fail activation - migrations may have already run on another instance
+		pluginAPI.Log.Error("failed to run migrations", "error", err)
+	} else if wasUpdated {
+		// Wrap config in the configuration struct that has the proper nesting
```
BenCookie95: Would it be worth doing the persistence in the `runAllMigrations` function?

server/migrations.go
davidkrauser: We have two migrations in there currently, one that existed prior to my change and a new one. I assume if we add more migrations in the future they could be added here :slightly_smiling_face:

server/migrations.go
davidkrauser: That's a good point. I've adjusted the code to set this _after_ we persist the config.

server/main.go
davidkrauser: That's a good suggestion. Done :+1:

server/main.go
```diff
@@ -72,30 +71,12 @@ func (p *Plugin) OnActivate() error {
...
-		if unmarshalErr := json.Unmarshal(marshalBytes, &out); unmarshalErr != nil {
-			return fmt.Errorf("failed to unmarshal migrated configuration: %w", unmarshalErr)
-		}
-
-		// Persist to database via API
-		if saveErr := pluginAPI.Configuration.SavePluginConfig(out); saveErr != nil {
-			return fmt.Errorf("failed to save migrated configuration: %w", saveErr)
-		}
-
+		return fmt.Errorf("failed to run migrations: %w", err)
```
davidkrauser, line 74: I changed this to an explicit failure if we can't run the migration. If it fails and we end up in a bad state, or if we're just using the old config structure, the plugin won't work at all.

server/main.go
```diff
@@ -69,13 +69,14 @@ func (p *Plugin) OnActivate() error {
 		PluginVersion:  manifest.Version, // Manifest imported from manifest.go which is generated by the build process
 	})
 
-	updated, newCfg, err := migrateServicesToBots(p.API, pluginAPI, *p.configuration.Config())
+	potentiallyUpdatedConfig, wasUpdated, err := runAllMigrations(p.API, pluginAPI, *p.configuration.Config())
```
BenCookie95: Will passing `*p.configuration.Config()` into `runAllMigrations` in a HA scenario cause issues because of the mutex? If cluster1 processes the update while cluster2 is waiting for the lock, cluster2 will have an old config reference right? `runAllMigrations` will  return the old config in cluster2 and cluster2 will keep the old config despite cluster1 having run the migration?

BenCookie95 (changes requested):

server/main.go
davidkrauser: Nice catch! This looks like this was a problem even before my change. I've moved the call to get the config after getting the lock: https://github.com/mattermost/mattermost-plugin-agents/pull/409/commits/10f646d6f737754e37cb19fedb010ace36f158ff

This isn't perfect, though. We could have a scenario where:
- One node migrates the config, persists the config, and sets the done flag, then releases the lock
- Another node gets the lock, but hasn't received notification that the config has changed
- We try to migrate, but the done flag is set. So we don't migrate and try to run with a bad config

Since this is existing behavior and the new migration is disabled in this PR, would you be okay with me fixing this in a follow-up PR?

BenCookie95: I'm not sure I understand the scenario you're describing if we are getting the config after the lock won't it have the most up to date config? and the migrations will just get skipped?

davidkrauser: The lock is released after we call `SavePluginConfig()`, but (if I understand correctly), `SavePluginConfig()` returns _before_ ensuring all nodes have received the updated config. In other words, we've ensured that the config has been saved, and that notifications are going out to the other nodes; however, we don't have guarantees that `OnConfigurationChange()` and the subsequent `LoadPluginConfiguration()` calls have happened.

davidkrauser: To fix this, my feeling is that we need to drop the "done" flags we're putting in the KV store, and just always try to migrate. If the data is already in the correct format, then it becomes almost a no-op.

BenCookie95: Ahh I see, thanks. We should probably solve for that but I'm ok with that being in another PR!

PR #412, opened 2025-10-15
[MM-66149] Improved eval support, run in CI

nickmisasi:
#### Summary

Adds support for running evals against different LLM providers, along with a configurable model parameter for testing. Also enables running of the evals in CI pipelines with results posted as unified PR comments.

CI pipelines now cover Anthropic, Azure, OpenAI, and OpenAI Compatible providers. Results from all providers are aggregated into a single GitHub comment on pull requests.

Expects environment variables:

**Anthropic:**
- `ANTHROPIC_API_KEY`
- `ANTHROPIC_MODEL` (defaults to claude-sonnet-4-20250514)

**OpenAI:**
- `OPENAI_API_KEY`
- `OPENAI_MODEL` (defaults to gpt-4o)

**Azure:**
- `AZURE_OPENAI_API_KEY`
- `AZURE_OPENAI_ENDPOINT`
- `AZURE_OPENAI_MODEL` (defaults to gpt-4o)

**OpenAI Compatible (for local LLMs):**
- `OPENAI_COMPATIBLE_API_URL` (required - e.g., http://localhost:8080/v1)
- `OPENAI_COMPATIBLE_MODEL` (required - e.g., llama-3)
- `OPENAI_COMPATIBLE_API_KEY` (optional - not needed for local LLMs)

You can run the evals in interactive mode with `make evals`. You can run headless with `make evals-ci`. You can generate GitHub-formatted comments with `make evals-comment`.

By default, all LLM providers (openai, anthropic, azure) will be run. You can scope to a specific one with `LLM_PROVIDER=anthropic` or multiple with `LLM_PROVIDER=anthropic,openai` or include OpenAI compatible with `LLM_PROVIDER=openaicompatible`.

**Grader-specific LLM:**
This PR introduces the ability to set a specific LLM provider and model as the grader. This allows us to use the strongest models we have available for grading.
- `GRADER_LLM_PROVIDER` (default `openai`) - can be any of the supported LLM Providers: `openai`, `anthropic`, `azure`, `openaicompatible`
- `GRADER_LLM_MODEL` (default `gpt-5`) - can be any of the supported models for that LLM provider

**CI Pipeline:**
- Eval jobs run in parallel as a matrix for all providers
- Results are aggregated into a single unified PR comment
- Jobs always succeed (never block PRs)
- Each provider's results are displayed in separate sections with overall statistics and detailed failure information

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-66150  
https://mattermost.atlassian.net/browse/MM-66149

#### Screenshots

[Include screenshots of the unified PR comment showing results from multiple providers]

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,8 +8,6 @@
 - Run e2e tests: `make e2e`
 - Run specific e2e test file: `cd e2e && npx playwright test filename.spec.ts --reporter=list`
 - Run prompt evaluations (CI mode, non-interactive): `make evals-ci`
-- Run prompt evaluations and generate GitHub comment: `make evals-comment`
-- Build evalviewer binary: `make evalviewer`
 - Run evals with specific provider: `LLM_PROVIDER=openai make evals-ci` (options: openai, anthropic, azure, openaicompatible, all)
 - Run evals with specific model: `ANTHROPIC_MODEL=claude-3-opus-20240229 make evals-ci`
 - Run evals with multiple providers: `LLM_PROVIDER=openai,anthropic make evals-ci`
--- a/evals/evals.go
+++ b/evals/evals.go
@@ -80,7 +80,9 @@ func createProvider(providerName string, modelOverride string) (llm.LanguageMode
 		provider := anthropic.New(llm.ServiceConfig{
 			APIKey:       apiKey,
 			DefaultModel: model,
-		}, []string{}, httpClient)
+		}, llm.BotConfig{
+			ReasoningEnabled: true,
+		}, httpClient)
 		if provider == nil {
 			return nil, errors.New("failed to create Anthropic provider")
 		}


anthropic/anthropic.go
```diff
@@ -216,11 +217,15 @@ func (a *Anthropic) streamChatWithTools(state messageState) {
...
+	// Anthropic requires a minimum thinking budget of 1024 tokens
+	// If the thinking budget is more than the max_tokens, Anthropic will return an error.
+	if state.config.EnableThinking && thinkingBudget < int64(state.config.MaxGeneratedTokens) {
+		params.Thinking = anthropicSDK.ThinkingConfigParamUnion{
+			OfEnabled: &anthropicSDK.ThinkingConfigEnabledParam{
+				Type:         "enabled",
+				BudgetTokens: thinkingBudget,
+			},
+		}
 	}
```
nickmisasi: Noticed an issue that some actions (particularly the "choose a reaction for me" would override the max_tokens to 25, which would then cause this to break. For now I've added the `EnableThinking` config, with a `WithoutThinking()` option for the LLM to allow for these to work. 

I've opened a ticket here: https://mattermost.atlassian.net/browse/MM-66224 to add a few more configs around reasoning to the front-end like a global disable, explicit reasoning budget/level, etc.

nickmisasi:
@crspeller @davidkrauser The pipelines are working, but there are some eval failures across the different models. There's also an inherent indeterministic output from both the tests and the graders, so these could be flaky over time. 

These pipelines aren't going to hold up merging a PR, so perhaps we open tickets for the failing evals and address those with some prompt engineering in the future?

Makefile
```diff
@@ -304,6 +304,34 @@ mcp-server:
 	@echo Building MCP server...
 	$(GO) build $(GO_BUILD_FLAGS) -o bin/mattermost-mcp-server ./mcpserver/cmd/main.go
 
+## Builds the evalviewer binary.
+.PHONY: evalviewer
+evalviewer:
+	@echo Building evalviewer...
+	cd cmd/evalviewer && $(GO) build $(GO_BUILD_FLAGS) -o ../../bin/evalviewer .
```
davidkrauser: **nit:** do we need the `cd` command here? Or will something like `$(GO) build $(GO_BUILD_FLAGS) -o ./bin/evalviewer ./cmd/evalviewer/...` work?

evals/evals.go
```diff
@@ -30,31 +32,171 @@ type Eval struct {
...
+// getProviderConfig reads environment variables for a specific provider
+func getProviderConfig(providerName string) (ProviderConfig, error) {
+	config := ProviderConfig{
+		Timeout: 20 * time.Second,
+	}
+
+	switch strings.ToLower(providerName) {
+	case "openai":
+		config.APIKey = os.Getenv("OPENAI_API_KEY")
+		config.Model = os.Getenv("OPENAI_MODEL")
```
davidkrauser: **question:** do we want to support open AI compatible APIs for this? I think we'd just need to be able to specify an API URL, right? Would make it possible to use this with local LLMs.

evals/evals.go
```diff
@@ -30,31 +32,171 @@ type Eval struct {
 	runNumber int
 }
 
+// ProviderConfig holds provider-specific configuration
+type ProviderConfig struct {
+	APIKey  string
+	Model   string
+	APIURL  string // For compatible providers like Azure
+	Timeout time.Duration
+}
```
davidkrauser: **suggestion:** do we need this type and the `getProviderConfig()` function? We could just create the providers directly from the environment variables in `createProvider()`.

davidkrauser (commented):
@nickmisasi wrote:
> The pipelines are working, but there are some eval failures across the different models. There's also an inherent indeterministic output from both the tests and the graders, so these could be flaky over time.

I don't love the idea of introducing failing pipelines, even if they  don't block merging. What value does running them against each PR add if they always fail, or if they're unreliable?

evals/evals.go
nickmisasi: Added

anthropic/anthropic.go
```diff
@@ -232,7 +233,7 @@ func (a *Anthropic) streamChatWithTools(state messageState) {
 
 	// Anthropic requires a minimum thinking budget of 1024 tokens
 	// If the thinking budget is more than the max_tokens, Anthropic will return an error.
-	if thinkingBudget < int64(state.config.MaxGeneratedTokens) {
+	if state.config.EnableThinking && thinkingBudget < int64(state.config.MaxGeneratedTokens) {
```
crspeller: Are these changes still relevant?

CLAUDE.md
```diff
@@ -7,6 +7,14 @@
 - Run specific Go test: `go test -v ./server/path/to/package -run TestName`
 - Run e2e tests: `make e2e`
 - Run specific e2e test file: `cd e2e && npx playwright test filename.spec.ts --reporter=list`
+- Run prompt evaluations (interactive TUI): `make evals`
```
crspeller: Probably don't want agents running the TUI and getting stuck.

Makefile
```diff
@@ -304,6 +304,41 @@ mcp-server:
...
+##   OPENAI_MODEL: Model to use for OpenAI (default: gpt-4o)
+##   ANTHROPIC_API_KEY: Anthropic API key
+##   ANTHROPIC_MODEL: Model to use for Anthropic (default: claude-sonnet-4-20250514)
+##   AZURE_OPENAI_API_KEY: Azure OpenAI API key
+##   AZURE_OPENAI_ENDPOINT: Azure OpenAI endpoint URL
+##   AZURE_OPENAI_MODEL: Model to use for Azure OpenAI (default: gpt-4o)
+.PHONY: evals
+evals: evalviewer
+	@echo Running evaluations interactively...
+	./bin/evalviewer run -v ./conversations ./threads ./channels ./react
```
crspeller, line 329: Can we do ./... so we don't have to remember to update this list?

CLAUDE.md
```diff
@@ -7,6 +7,14 @@
...
+- Run prompt evaluations (CI mode, non-interactive): `make evals-ci`
+- Run prompt evaluations and generate GitHub comment: `make evals-comment`
+- Build evalviewer binary: `make evalviewer`
```
crspeller: I don't think these two are relevant to an agent.

CLAUDE.md
```diff
@@ -7,6 +7,14 @@
...
+- Run evals with specific provider: `LLM_PROVIDER=openai make evals` (options: openai, anthropic, azure, openaicompatible, all)
+- Run evals with specific model: `ANTHROPIC_MODEL=claude-3-opus-20240229 make evals`
+- Run evals with multiple providers: `LLM_PROVIDER=openai,anthropic make evals`
+- Run evals with OpenAI compatible API (e.g., local LLMs): `LLM_PROVIDER=openaicompatible OPENAI_COMPATIBLE_API_URL=http://localhost:8080/v1 OPENAI_COMPATIBLE_MODEL=llama-3 make evals`
```
crspeller: Agents should use the non-gui path no?

Makefile
nickmisasi: Given the file structure, this means that the evals run every Go test in the repo. I think we should leave it like this for now

crspeller (approved):

PR #414, opened 2025-10-16
[MM-66171] Anthropic tool calling fixes

nickmisasi:
#### Summary
Anthropic requires [thinking blocks to be transmitted back to them when using Extended Thinking + Tool Use](https://docs.claude.com/en/docs/build-with-claude/extended-thinking#extended-thinking-with-tool-use). This wasn't happening, which meant that tool calls would error with a 400 from the API. 

This PR fixes this issue, first by consuming [signature blocks](https://docs.claude.com/en/docs/build-with-claude/extended-thinking#thinking-encryption) returned from Anthropic, and then passing those, along with the reasoning response from Anthropic into tool use. 

This PR also fixes an issue with thread title requests that were failing because they were passing an empty system prompt to Anthropic - which was throwing a 400 error as well.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66171

#### Release Note

```release-note
None
```

### DIFF
--- a/streaming/streaming.go
+++ b/streaming/streaming.go
@@ -309,14 +309,6 @@ func (p *MMPostStreamService) StreamToPost(ctx context.Context, stream *llm.Text
 						p.mmClient.LogDebug("Added reasoning signature to post props", "post_id", post.Id)
 					}
 					reasoningBuffer.Reset()
-				} else if reasoningText, ok := event.Value.(string); ok {
-					// Backward compatibility: handle old string format
-					p.sendPostStreamingReasoningEvent(post, reasoningText, "reasoning_summary_done")
-					if reasoningText != "" {
-						post.AddProp(ReasoningSummaryProp, reasoningText)
-						p.mmClient.LogDebug("Added reasoning summary to post props (legacy format)", "post_id", post.Id, "reasoning_length", len(reasoningText))
-					}
-					reasoningBuffer.Reset()
 				}
 			case llm.EventTypeToolCalls:
 				// Handle tool call event


anthropic/anthropic.go
```diff
@@ -216,17 +230,22 @@ func (a *Anthropic) streamChatWithTools(state messageState) {
...
+	// Anthropic requires a minimum thinking budget of 1024 tokens
+	// If the thinking budget is more than the max_tokens, Anthropic will return an error.
+	if thinkingBudget < int64(state.config.MaxGeneratedTokens) {
+		params.Thinking = anthropicSDK.ThinkingConfigParamUnion{
+			OfEnabled: &anthropicSDK.ThinkingConfigEnabledParam{
+				Type:         "enabled",
+				BudgetTokens: thinkingBudget,
+			},
+		}
 	}
```
nickmisasi, line 242: I pulled this from https://github.com/mattermost/mattermost-plugin-agents/pull/412/files#diff-0017aaca6b35cba6c5109bba02455a1abbce06817d1a051aaf294dcbd88764ec as it was necessary to fix the post title and react for me flows with reasoning enabled

Left out the "EnableThinking" parameter as it's not necessary, and can be included once the evals PR merges.

devinbinnie: Can you summarize this, perhaps in an .md file?

streaming/streaming.go
```diff
@@ -293,15 +294,27 @@ func (p *MMPostStreamService) StreamToPost(ctx context.Context, stream *llm.Text
...
+					if reasoningData.Text != "" {
+						post.AddProp(ReasoningSummaryProp, reasoningData.Text)
+						p.mmClient.LogDebug("Added reasoning summary to post props", "post_id", post.Id, "reasoning_length", len(reasoningData.Text))
+					}
+					if reasoningData.Signature != "" {
+						post.AddProp(ReasoningSignatureProp, reasoningData.Signature)
+						p.mmClient.LogDebug("Added reasoning signature to post props", "post_id", post.Id)
+					}
+					reasoningBuffer.Reset()
+				} else if reasoningText, ok := event.Value.(string); ok {
```
crspeller: Don't think we need this.

anthropic/anthropic.go
nickmisasi: No sorry

PR #415, opened 2025-10-16
MM-66207: Refactor llmbot_post component and fix loading state UX issues

nickmisasi:
## Overview

This PR refactors the large `llmbot_post.tsx` component into a modular structure and fixes several UX issues with loading state indicators.

### Component Restructure
- **Created `llmbot_post/` folder** with the following modular structure:
  - `index.tsx` - Main component (reduced from 589 to ~371 lines)
  - `types.ts` - All TypeScript interfaces and enums
  - `styles.tsx` - All styled-components definitions
  - `reasoning_display.tsx` - Reasoning summary display component
  - `controls_bar.tsx` - Action buttons component
  - `utils.ts` - Utility functions

### Loading State Behavior

| Phase | "Starting..." | "Thinking..." | Blinking Cursor | Button |
|-------|--------------|---------------|-----------------|--------|
| Initial | ✅ | ❌ | ❌ | Stop Generating |
| Reasoning | ❌ | ✅ | ❌ | Stop Generating |
| Streaming | ❌ | ❌ | ✅ | Stop Generating |
| Complete | ❌ | ❌ | ❌ | Regenerate |

Starting:
image
Thinking:
image
Done:
image


webapp/src/components/llmbot_post/index.tsx
```diff
@@ -4,256 +4,70 @@
 import React, {useEffect, useRef, useState} from 'react';
```
crspeller, line 4: Can we name this file `llmbot_post.tsx`? I find the `index.tsx` pattern annoying for finding files.

webapp/src/components/llmbot_post/permalink_data.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {PermalinkData} from './types';
+
+export function extractPermalinkData(post: any): PermalinkData | null {
```
crspeller, line 15: This might be better organized as a file called `permalink_data.ts` with the type and this utility in it.

webapp/src/components/llmbot_post/types.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
```
crspeller: I don't really like separate types files. Can we put these types in the files that own them?

webapp/src/components/llmbot_post/styles.tsx
```diff
@@ -0,0 +1,174 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import styled from 'styled-components';
+
+export const PostBody = styled.div`
+`;
+
+export const ControlsBar = styled.div`
```
crspeller: I find it easier to put styled components in the files that own them. That's part of why styled components are nice because everything is together. Unless these are truly shared (I don't think they are) 

If there are a bunch I usually put them after the component so it looks nicer.

crspeller (changes requested):
Some broad style comments. Happy to change style with good reason, but this would be a change in how this plugin is generally organized.

webapp/src/components/llmbot_post/llmbot_post.tsx
```diff
@@ -407,6 +233,7 @@ export const LLMBotPost = (props: Props) => {
                     setIsReasoningLoading(false);
                 } else if (data.control === 'cancel') {
                     setGenerating(false);
+                    setPrecontent(false);
                     setStopped(false);
                     setIsReasoningLoading(false);
```
davidkrauser, line 238: These states are hard to reason about since their effects are sprinkled throughout these files. I'm wondering if we should introduce an explicit state-machine to handle this in a future improvement. Would make it easier to understand and change how this works.

PR #417, opened 2025-10-21
LLM Bridge prototyping Build

crspeller:


public/bridgeclient/completion.go
```diff
@@ -0,0 +1,356 @@
...
+// Returns the complete response text or an error.
+//
+// Example:
+//
+//	response, err := client.AgentCompletion("gpt4", client.CompletionRequest{
+//	    Posts: []client.Post{
+//	        {Role: "user", Message: "What is the capital of France?"},
+//	    },
+//	})
+func (c *Client) AgentCompletion(agent string, request CompletionRequest) (string, error) {
```
fmartingr, line 19: Is there a way to target the default agent? Maybe bundling `agent` inside the request and leaving it empty would use the default one?

I wouldn't like `AgentCompletion("", CompletionRequest{})`

PR #420, opened 2025-10-27
Hide AI UI Elements for Unauthorized Users

crspeller:
#### Summary
- Remove apps bar icon when user has no access to any bots
- Hide AI Action button when channel has no accessible bots
- Hide Ask AI button when channel has no accessible bots
- Remove RHS unconfigured/blocked messages for unauthorized users
- Clean up unused imports and variables

This prevents showing "contact admin" messages to users who are intentionally restricted from accessing AI features.

#### Ticket Link
Request on hub: https://hub.mattermost.com/private-core/pl/nepwbjwuw38edpx98o6jxmri3a

### DIFF
--- a/webapp/src/components/rhs/rhs.tsx
+++ b/webapp/src/components/rhs/rhs.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React, {useState, useEffect, useCallback} from 'react';
-import {FormattedMessage, useIntl} from 'react-intl';
+import {useIntl} from 'react-intl';
 import {useDispatch, useSelector} from 'react-redux';
 import styled from 'styled-components';
 


webapp/src/index.tsx
```diff
@@ -152,7 +155,7 @@ export default class Plugin {
 
         registry.registerAdminConsoleCustomSetting('Config', Config);
         if (rhs) {
-            registry.registerChannelHeaderButtonAction(<IconAIContainer src={aiIcon}/>, () => {
+            registry.registerChannelHeaderButtonAction(<ChannelHeaderIcon/>, () => {
```
davidkrauser, line 158: **question:** if the bot list is empty, we'll set a `null` action here. Is that what we want? Would it be better to not register an action at all for that case?

crspeller: `initialize` only runs once. If the configuration updates we want the icon to update.

PR #424, opened 2025-10-30
[MM-66224] Configurable Reasoning/Thinking effort/tokens

nickmisasi:
#### Summary

Added configurable reasoning value at the Agent level. I chose the agent level because this would allow you more granularity - the Agents default bot could be set with no or little reasoning to speed up general use interactions like react for me, summarize, etc, while maintaining other smarter Agents with more advanced reasoning - all from one single LLM Service.

Both OpenAI and Anthropic have a toggle for enable/disable

Anthropic contains a configurable ThinkingBudget with some usage guidelines
OpenAI/Compatible will expose a dropdown menu allowing you to select between Minimal, Low, Medium, and High 

I also increased the server test coverage in a few places where it was lacking. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66224

#### Screenshots
Anthropic:

image

The inline docs upper bound takes the actual max output tokens value set for that service - previous screenshot it was set to 64000, this one I upped to 101000
image
Shows error when exceeding max_tokens
image

OpenAI/Compat:
image

#### Release Note

```release-note
Configurable reasoning on/off and level on a per-agent basis
```


anthropic/anthropic_test.go
```diff
@@ -588,3 +588,278 @@ func TestConversationToMessages(t *testing.T) {
 		})
 	}
 }
+
+func TestNew(t *testing.T) {
```
crspeller: This doesn't test anything.

anthropic/anthropic_test.go
```diff
@@ -588,3 +588,278 @@ func TestConversationToMessages(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			a := New(tt.serviceConfig, tt.botConfig, nil)
+			require.NotNil(t, a)
+			tt.validate(t, a)
+		})
+	}
+}
+
+func TestThinkingBudgetConfiguration(t *testing.T) {
```
crspeller, line 592: Again what does this test?

anthropic/anthropic.go
```diff
@@ -52,7 +53,8 @@ func New(llmService llm.ServiceConfig, enabledNativeTools []string, httpClient *
 		defaultModel:       llmService.DefaultModel,
 		inputTokenLimit:    llmService.InputTokenLimit,
 		outputTokenLimit:   llmService.OutputTokenLimit,
-		enabledNativeTools: enabledNativeTools,
+		enabledNativeTools: botConfig.EnabledNativeTools,
+		botConfig:          botConfig,
```
crspeller: Probably don't want to depend directly on bot config. Rather extract the reasoning enabled/thinking budget prams.

config/config.go
```diff
@@ -164,6 +164,7 @@ func OpenAIConfigFromServiceConfig(serviceConfig llm.ServiceConfig, enabledNativ
 		StreamingTimeout:   streamingTimeout,
 		SendUserID:         serviceConfig.SendUserID,
 		UseResponsesAPI:    serviceConfig.UseResponsesAPI,
-		EnabledNativeTools: enabledNativeTools,
+		EnabledNativeTools: botConfig.EnabledNativeTools,
+		BotConfig:          botConfig,
```
crspeller: Same here probably want extract the individual parameters.

config/config_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mattermost/mattermost-plugin-ai/llm"
+	"github.com/mattermost/mattermost-plugin-ai/openai"
+)
+
+func TestOpenAIConfigFromServiceConfig(t *testing.T) {
```
crspeller: Other than maybe the streaming timeout calculation, not sure this needs a test.

llm/configuration.go
```diff
@@ -66,6 +66,22 @@ type BotConfig struct {
...
+	// Valid values: "minimal", "low", "medium", "high"
+	// Only applicable to OpenAI with ResponsesAPI enabled
+	// Default: "medium"
+	ReasoningEffort string `json:"reasoningEffort"`
+
+	// ThinkingBudget determines the token budget for Anthropic thinking
+	// Must be at least 1024 and cannot exceed the OutputTokenLimit
+	// Only applicable to Anthropic
+	// Default: 1/4 of OutputTokenLimit, capped at 8192
+	ThinkingBudget int `json:"thinkingBudget"`
```
crspeller, line 84: Painful to need both of these. Maybe we could map the Reasoning effort of OpenAI to specific numbers or something? Probably not worth it. Just complaining.

openai/openai.go
```diff
@@ -834,15 +835,32 @@ func (s *OpenAI) convertToResponseParams(params openai.ChatCompletionNewParams,
...
+		case "":
+			// Empty string defaults to medium effort for clarity
+			effort = shared.ReasoningEffortMedium
+		default:
+			effort = shared.ReasoningEffortMedium
+		}
+
+		result.Reasoning = shared.ReasoningParam{
+			Effort: effort,
+			// Request a detailed summary of the reasoning
```
crspeller: Excessive comment.

openai/openai_test.go
```diff
@@ -734,3 +734,103 @@ func TestToolBufferElement(t *testing.T) {
...
+			botConfig: llm.BotConfig{
+				ReasoningEnabled: false,
+				ReasoningEffort:  "high",
+			},
+			shouldSetReasoning: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
crspeller, line 798: Testing a function without calling it. Interesting strategy.

webapp/src/components/system_console/reasoning_config.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+    // Determine if this service supports reasoning
+    const isAnthropic = props.service.type === 'anthropic';
+    const isOpenAIWithResponses = ['openai', 'openaicompatible', 'azure'].includes(props.service.type) && props.service.useResponsesAPI;
+
+    if (!isAnthropic && !isOpenAIWithResponses) {
+        return null;
+    }
+
+    const reasoningEnabled = props.bot.reasoningEnabled ?? true; // Default to enabled
```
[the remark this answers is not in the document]
crspeller: Not a bad idea.

anthropic/anthropic_test.go
nickmisasi: It is a superficial test I agree, but I think we should leave it. It _does_ test for things like if I were to override one of the configs for debug purposes, making sure I revert it before it accidentally sneaks into master.

anthropic/anthropic_test.go
nickmisasi: Refactored the code so that this actually calls the function rather than calculating the thinking itself

anthropic/anthropic.go
nickmisasi: Updated

config/config.go
nickmisasi: UPdated

config/config_test.go
nickmisasi: For the same reason as Anthropic I think this is still valuable. Or at least not completely superfluous. If you strongly disagree I can remove :P

llm/configuration.go
nickmisasi: I think the alternative would look hackier than just having both

anthropic/anthropic_test.go
crspeller: I feel like having this around as an example test for LLMs and people could be a determent. Like it could possibly catch if you override in just the right place, but I don't think it's worth having the extra code around.

crspeller (approved):
Looks good. Not going to block on the superficial tests comment, but I would remove them.

openai/openai_test.go
BenCookie95: bold-strategy-cotton

webapp/src/components/system_console/reasoning_config.tsx
```diff
@@ -0,0 +1,247 @@
...
+                                />
+                                <HelpText>
+                                    {intl.formatMessage({
+                                        defaultMessage: 'Token budget for extended thinking. Higher values allow deeper reasoning but increase response time and cost. Must be between 1024 and {maxTokens}. Leave blank to use default ({defaultBudget}).',
+                                    }, {
+                                        maxTokens: props.maxTokens,
+                                        defaultBudget: getDefaultThinkingBudget(),
+                                    })}
+                                </HelpText>
+                                {typeof props.bot.thinkingBudget === 'number' && props.bot.thinkingBudget > 0 && props.bot.thinkingBudget < 1024 && (
```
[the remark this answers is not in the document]
BenCookie95: Same for 4096 and 8192, might be worth making constants

webapp/src/components/system_console/reasoning_config.tsx
```diff
@@ -0,0 +1,247 @@
...
+        box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075), 0 0 8px rgba(102, 175, 233, 0.75);
+        outline: 0;
+    }
+`;
+
+const ErrorText = styled.div`
+    font-size: 12px;
+    font-weight: 400;
+    line-height: 16px;
+    color: #D24B4E;
```
BenCookie95: Do we have a theme colour for this?

BenCookie95 (approved):

PR #425, opened 2025-10-30
[MM-66080] Run MCP server from the Agents plugin

BenCookie95:
#### Summary
Add MCP endpoint to the agents plugin so you can run the streamable MCP server as part of the plugin. 

System console setting:
Screenshot 2025-11-05 at 2 02 18 PM

This needs the following change in order for streaming to work https://github.com/mattermost/mattermost/pull/34411, should we bump the minimum mm server version of the plugin before releasing?
 
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66080

#### Release Note

```release-note
Add ability to run Mattermost MCP server from the Agents plugin
```

### DIFF
--- a/mcpserver/plugin_handlers.go
+++ b/mcpserver/plugin_handlers.go
@@ -54,13 +54,21 @@ func NewPluginMCPHandlers(siteURL string, logger loggerlib.Logger) (*PluginMCPHa
 		nil, // ServerOptions
 	)
 
+	trackAIGenerated := true
+
+	// Create server config for tool provider
+	config := BaseConfig{
+		MMServerURL:         siteURL,
+		MMInternalServerURL: "",
+		DevMode:             false,
+		TrackAIGenerated:    &trackAIGenerated,
+	}
+
 	// Register tools with remote access mode
 	toolProvider := tools.NewMattermostToolProvider(
 		authProvider,
 		logger,
-		siteURL, // External server URL
-		"",      // Internal server URL (use external)
-		false,   // devMode
+		config,
 		tools.AccessModeRemote,
 	)
 	toolProvider.ProvideTools(mcpServer)


api/api.go
```diff
@@ -116,20 +118,29 @@ func (a *API) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Reques
 	router.Use(a.ginlogger)
 	router.Use(a.metricsMiddleware)
 
-	// LLM Bridge API v1 routes - inter-plugin only
-	llmBridgeRoute := router.Group("/bridge/v1")
-	llmBridgeRoute.Use(a.interPluginAuthorizationRequired)
+	// Store plugin.Context in gin.Context for use in middleware
```
crspeller: Maybe do this on `mcpServerGroup`?

api/middleware_mcp.go
```diff
@@ -0,0 +1,76 @@
...
+	config := a.pluginAPI.Configuration.GetConfig()
+	if config.ServiceSettings.SiteURL == nil || *config.ServiceSettings.SiteURL == "" {
+		a.pluginAPI.Log.Error("Site URL not configured, cannot send OAuth metadata URL")
+		c.AbortWithStatus(http.StatusUnauthorized)
+		return
+	}
+
+	siteURL := *config.ServiceSettings.SiteURL
+	// OAuth metadata is now under the plugin mcp-server path
+	resourceMetadataURL := fmt.Sprintf("%s/plugins/mattermost-ai/mcp-server/.well-known/oauth-protected-resource", siteURL)
```
crspeller, line 40: Question does `.well-known/oauth-protected-resource` work on a subpath like this or does it have to be at the root?

BenCookie95: It works, we tell it where it exists with the WWW-Authenticate header. I double checked the spec before doing it as well, perfectly valid to have this in a subpath

webapp/src/components/system_console/mcp_servers.tsx
```diff
@@ -328,6 +330,12 @@ const MCPServers = ({mcpConfig, onChange}: Props) => {
                                         onChange={(enabled) => onChange({...config, enabled})}
                                         helpText={intl.formatMessage({defaultMessage: 'Enable the Model Context Protocol (MCP) integration to access tools from MCP servers.'})}
                                     />
+                                    <BooleanItem
```
BenCookie95, line 333: This is slightly confusing when it's near the other embedded MCP server setting. Should we move it to another location?

nickmisasi: I think we should change the above to "Enable MCP Client", and the below to "Enable MCP Server". A link to docs will be really important, we should add some differentiating in the docs folder

crspeller: Agree with Nick here. Let's be very explicit. 
We should probably also support running the server without the client enabled.

PR #427, opened 2025-10-31
Tool UI fixes

BenCookie95:
#### Summary
- Match UI design in figma (Tool calls at bottom of document).
  - I deviated from the design slightly for font sizes, icon sizes and button sizes. I tried to mirror what we already had for other agents components.
- Tools are auto collapsed apart from when you have to decide whether to accept/reject. They auto collapse once you select an option
- Pretty print JSON when it's detected
- I fixed a bug with tools where if there was an error in the response of the tool we treated it the same as a network error and never bubbled the issue up to the user.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65734

#### Release Note

```release-note
Tools UX improvements
```

### DIFF
--- a/webapp/src/components/tool_card.tsx
+++ b/webapp/src/components/tool_card.tsx
@@ -170,7 +170,7 @@ const ResponseLabel = styled.div`
     line-height: 20px;
     color: rgba(var(--center-channel-color-rgb), 0.75);
     padding-top: 8px;
-    padding-left: 13px;
+    padding-left: 24px;
 `;
 
 const ResultContainer = styled.div`


webapp/src/websocket.ts
```diff
@@ -13,8 +13,24 @@ type WebsocketListenerObject = {
 }
 type WebsocketListeners = WebsocketListenerObject[]
 
+export type ToolPermissionWebsocketMessage = {
+    user_id: string;
+    root_post_id: string;
+    tool_name: string;
+    permission: 'auto-approve' | 'ask';
```
crspeller: Probably worth a custom type if we are going to repeat this.

webapp/src/components/tool_approval_set.tsx
```diff
@@ -44,10 +49,66 @@ const ToolApprovalSet: React.FC<ToolApprovalSetProps> = (props) => {
     // Track which tools are currently being processed
     const [isSubmitting, setIsSubmitting] = useState(false);
     const [error, setError] = useState('');
-    const [collapsedTools, setCollapsedTools] = useState<string[]>([]);
+    const [autoApprovedTools, setAutoApprovedTools] = useState<string[]>([]);
```
crspeller: Wondering if we should actually store this in redux and create a custom hook like we do for `useBotList`. Not sure if the websocket complexity is worth it. Thoughts?

conversations/tool_handling.go
```diff
@@ -119,3 +162,58 @@ func (c *Conversations) HandleToolCall(userID string, post *model.Post, channel
 
 	return nil
 }
+
+// HandleAutoApprovedToolCall is a simplified handler for auto-approved tool calls
+// This is called by the streaming service when all tools in a call are auto-approved
+func (c *Conversations) HandleAutoApprovedToolCall(postID string, toolIDs []string) {
+	post, err := c.mmClient.GetPost(postID)
```
crspeller: I am confused as to why we need all these extra checks here. If these checks are necessary they should be part of the regualr flow. Auto approved tools should behave in exactly the same way as the user just clicking the approve button right?

mcp/tool_permissions.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	// Check if tool is in the auto-approved list
+	for _, approvedTool := range permissions.AutoApprovedTools {
+		if approvedTool == toolName {
+			return true, nil
+		}
+	}
```
crspeller: nit: You can replace this with slices.ContainsFunc https://pkg.go.dev/slices#ContainsFunc (and a few other places around here)

mcp/tool_permissions.go
```diff
@@ -0,0 +1,133 @@
...
+	key := buildConversationPermissionKey(userID, rootPostID)
+	if err := client.KVSet(key, permissions); err != nil {
+		return fmt.Errorf("failed to store conversation permissions in KV store: %w", err)
+	}
+
+	return nil
+}
+
+// RemoveAutoApproval removes a tool from the auto-approve list for this conversation
+func RemoveAutoApproval(client mmapi.Client, userID, rootPostID, toolName string) error {
```
crspeller: Maybe reduce complexity here with `SetAutoApproval`?

webapp/src/websocket.ts
crspeller: Actually this is a boolean on the server, but a string here. Why?

api/api_post.go
```diff
@@ -304,6 +306,79 @@ func (a *API) handleToolCall(c *gin.Context) {
...
+	})
+}
+
+func (a *API) handleUpdateToolPermission(c *gin.Context) {
+	userID := c.GetHeader("Mattermost-User-Id")
+	post := c.MustGet(ContextPostKey).(*model.Post)
+
+	var data struct {
+		ToolName   string `json:"tool_name" binding:"required"`
+		Permission string `json:"permission" binding:"required,oneof=auto-approve ask"`
```
crspeller: Feels strange that this gets crunched down to a boolean. Why not just accept a boolean?

conversations/tool_handling.go
```diff
@@ -28,6 +29,19 @@ func (c *Conversations) HandleToolCall(userID string, post *model.Post, channel
...
 	}
 
+	// Get the root post ID for conversation-scoped permissions
+	rootPostID := post.RootId
+	if rootPostID == "" {
+		rootPostID = post.Id
+	}
+
+	// Handle permission updates
+	if autoApproveTool != "" {
```
crspeller: Why is this handling permissions updates? Isn't this handle by a separate endpoint?

conversations/tool_handling.go
```diff
@@ -39,21 +53,50 @@ func (c *Conversations) HandleToolCall(userID string, post *model.Post, channel
...
+
 	llmContext := c.contextBuilder.BuildLLMContextUserRequest(
 		bot,
 		user,
 		channel,
 		c.contextBuilder.WithLLMContextDefaultTools(bot, mmapi.IsDMWith(bot.GetMMBot().UserId, channel)),
 	)
 
+	// Check if tool is auto-approved
+	isAutoApproved := func(toolName string) bool {
```
crspeller: This feels weird, just run this once in the loop and use the result?

conversations/tool_handling.go
```diff
@@ -39,21 +53,50 @@ func (c *Conversations) HandleToolCall(userID string, post *model.Post, channel
...
+		return slices.Contains(autoApprovedTools, toolName)
+	}
+
 	for i := range tools {
-		if slices.Contains(acceptedToolIDs, tools[i].ID) {
+		// Mark if tool was auto-approved
```
crspeller: Why?

crspeller (changes requested):
This is more PR than I was expecting!

webapp/src/components/tool_approval_set.tsx
BenCookie95: I did go back and forth on this for a bit but I thought the more complete/future proof solution was a websocket

crspeller: What does the websocket get us over redux? 
If we do decide to keep it we might want to at least wrap the complexity in a custom hook.

conversations/tool_handling.go
BenCookie95: The flow is slightly different for auto approved tools because we have a bit of recursion going on. Imagine you have auto approved all the tools in your conversation. `StreamToPost` calls `HandleAutoApprovedToolCall`, the permission check is a bit redundant in the first instance because it would have already been checked before it makes it's way to `StreamToNewPost` but `HandleAutoApprovedToolCall`->`HandleToolCall`->`LLM.ChatCompletion`->`StreamToNewPost`->`StreamToPost`->`HandleAutoApprovedToolCall` can happen as long as `LLM.ChatCompletion` keeps returning tool calls for us to make. So you could have a situation where you reach your usage restrictions or you're removed from a channel between auto approved tool call events. This could theoretically keep running forever as long as the llm keeps returning tool calls to make. 

This makes me wonder if we need a stop button to stop the llm, right now there is no way to do that from the UI. Claude and other services turn the post button into a stop button while these things are running. Might be worth copying that but it definitely adds extra scope to this PR.

We should also consider a max number of turns so this doesn't run forever.

webapp/src/components/tool_approval_set.tsx
BenCookie95: Having multiple tabs of Mattermost open mostly. We currently load permissions for each conversation when we open the thread which makes it a little redundant but I was thinking if we moved to global tool permissions we would need this websocket event to keep those in sync. I'm not tied to either approach so I'm open to change this if you feel strongly about it!

I also just realised I'm making redundant calls to fetch tools per set when it should be per conversation so I'll fix that.

crspeller: Alright, that's good enough for me. Maybe we can see if we can contain some of the complexity whit a custom hook. The complexity is mostly what I worry about.

streaming/streaming.go
```diff
@@ -339,6 +378,25 @@ func (p *MMPostStreamService) StreamToPost(ctx context.Context, stream *llm.Text
...
+							autoApprovedIDs = append(autoApprovedIDs, tc.ID)
+						} else {
+							allAutoApproved = false
+						}
+					}
+
+					// If all tools are auto-approved, trigger execution via callback
+					if allAutoApproved && len(autoApprovedIDs) > 0 {
+						handler := p.toolCallHandler
+						if handler != nil {
```
crspeller: If this handler is not set it's a fatal error. Panic is probably acceptable here, but we should at least log it.

conversations/tool_handling.go
crspeller: I like the idea of restricting how long this goes and to have the ability to stop it. We should at least make a ticket for this. 

As for the extra permissions checking, it's really no different than if the LLM took a long time to think and the user's permissions changed during that. I don't think we really need to handle this situation. I think in the channel remove case at least the server will handle cutting off the websockets since we use the ChannelID to route them.

crspeller (changes requested):

crspeller:
FYI for those following this PR I removed the accept all feature from this PR and moved it to a separate one to reduce this PR's scope. https://github.com/mattermost/mattermost-plugin-agents/pull/437/

PR #428, opened 2025-11-05
Fix bot configuration migration to handle embedded service configs

davidkrauser:
Previously, bots with embedded service configurations (old format) were incorrectly rejected during validation before migration could run. This caused "Configured bot is not valid" errors on plugin startup when users had legacy configurations with embedded services.

Changes:
- Update BotConfig.IsValid() to accept bots with either serviceID (new) or embedded Service (old format during migration)
- Modify EnsureBots() to handle both config formats by extracting service from embedded config when serviceID is not present
- Remove KV store migration flags - migrations now run based on actual config format detection to ensure idempotency
- Add comprehensive test case based on real-world config with multiple bots sharing identical embedded services
- Fix linting issues: use switch statements instead of if-else chains

The migration now properly:
1. Detects configs with embedded services
2. Extracts and deduplicates services
3. Updates bot configs to reference services by ID
4. Clears embedded service fields after migration

```release-note
NONE
```

### DIFF
--- a/llm/configuration.go
+++ b/llm/configuration.go
@@ -69,13 +69,8 @@ type BotConfig struct {
 }
 
 func (c *BotConfig) IsValid() bool {
-	// Basic validation
-	if c.Name == "" || c.DisplayName == "" {
-		return false
-	}
-
-	// Must have a ServiceID
-	if c.ServiceID == "" {
+	// Basic validation - service validation happens separately
+	if c.Name == "" || c.DisplayName == "" || c.ServiceID == "" {
 		return false
 	}
 


bots/bots.go
```diff
@@ -90,10 +90,25 @@ func (b *MMBots) EnsureBots() error {
...
+				b.pluginAPI.Log.Error("Bot references non-existent service", "bot_name", botCfg.Name, "service_id", botCfg.ServiceID)
+				continue
+			}
+		case botCfg.Service != nil:
+			// Old format: bot has embedded service (should be migrated)
+			b.pluginAPI.Log.Warn("Bot is using deprecated embedded service configuration", "bot_name", botCfg.Name)
+			service = *botCfg.Service
+			ok = true
+		default:
+			b.pluginAPI.Log.Error("Bot has neither serviceID nor embedded service", "bot_name", botCfg.Name)
```
crspeller: Do we need this? Shouldn't the migration always take care of this? 
If we do what case does this cover?

davidkrauser: From what I understand: using `mmctl` to set the config doesn't trigger `OnActivate()` (which is where we run migrations). It only triggers `OnConfigurationChange()`. This missing code was the cause of the specific error reported by @jlandells (he's using `mmctl` to load the config).

davidkrauser: Just tested this using `mmctl` - and I missed an edge case where the service is missing an ID. Just pushed a fix for that, too.

crspeller: Talked offline. Let's try to run the migration `OnConfigurationChange`

llm/configuration.go
```diff
@@ -69,8 +69,13 @@ type BotConfig struct {
...
-	if c.Name == "" || c.DisplayName == "" || c.ServiceID == "" {
+	// Basic validation
+	if c.Name == "" || c.DisplayName == "" {
+		return false
+	}
+
+	// Must have a ServiceID
+	if c.ServiceID == "" {
 		return false
 	}
```
nickmisasi: I think this can be reverted

bots/bots.go
davidkrauser: @crspeller this code is now removed, and we're running the migration in both `OnActivate` and `OnConfigurationChange`. We can't drop it from `OnActivate` because we need a valid (migrated) config to initialize the bots.

nickmisasi (approved):

llm/configuration.go
davidkrauser: Yes, it can :+1:

PR #430, opened 2025-11-05
Native bedrock (Initial implementation)

nickmisasi:
#### Summary
Adds native support for Amazon Bedrock, no longer requiring the use of an OpenAI API gateway.

**What's included**
- Chat
- Streaming Chat
- Support for tool calling/MCP
- Support for vision
- Evals framework support for Bedrock

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66127

#### Release Note

```release-note
Added native support for Amazon Bedrock as an LLM service type
```

### DIFF
--- a/.github/scripts/aggregate-eval-results.js
+++ b/.github/scripts/aggregate-eval-results.js
@@ -10,7 +10,7 @@ async function aggregateAndPostResults(github, context) {
     let combinedComment = '## 🤖 LLM Evaluation Results\n\n';
 
     const resultsDir = 'eval-results';
-    const providers = ['eval-results-openai', 'eval-results-anthropic', 'eval-results-azure', 'eval-results-mistral'];
+    const providers = ['eval-results-openai', 'eval-results-anthropic', 'eval-results-azure', 'eval-results-mistral', 'eval-results-bedrock'];
 
     let hasAnyResults = false;
 


docs/providers.md
```diff
@@ -70,6 +71,186 @@ Obtain an [Anthropic API key](https://console.anthropic.com/settings/keys), then
...
+| Setting | Required | Description |
+|---------|----------|-------------|
+| **AWS Region** | Yes | AWS region where Bedrock is available (e.g., `us-east-1`, `us-west-2`, `eu-central-1`) |
+| **Custom Endpoint URL** | No | Optional custom endpoint for VPC endpoints or proxies (e.g., `https://bedrock-runtime.vpce-xxx.us-east-1.vpce.amazonaws.com`). Leave blank for standard AWS endpoints. |
+| **AWS Access Key ID** | No | IAM user access key ID for long-term credentials. Takes precedence over API Key if both are set. Can also be set via `AWS_ACCESS_KEY_ID` environment variable. |
+| **AWS Secret Access Key** | No | IAM user secret access key. Required if AWS Access Key ID is provided. Can also be set via `AWS_SECRET_ACCESS_KEY` environment variable. |
+| **API Key** | No | Bedrock console API key (base64 encoded, format: `ABSKQm...`). If IAM credentials above are set, they take precedence. Can also use environment variables or IAM roles. |
+| **Default Model** | Yes | The Bedrock model ID to use by default (see Available Models below) |
+
+### Available Models
```
crspeller: Let's not list models here, link to AWS docs instead?

bots/bots.go
```diff
@@ -173,7 +172,7 @@ func (b *MMBots) EnsureBots() error {
 }
 
 func (b *MMBots) getLLM(serviceConfig llm.ServiceConfig, botConfig llm.BotConfig) (llm.LanguageModel, error) {
-	// Create the correct model
+	// Create the correc@t model
```
crspeller: `correc@t` 😅

bedrock/bedrock.go
```diff
@@ -0,0 +1,550 @@
...
+
+type Bedrock struct {
+	client           *bedrockruntime.Client
+	defaultModel     string
+	inputTokenLimit  int
+	outputTokenLimit int
+	region           string
+}
+
+func New(llmService llm.ServiceConfig, awsRegion string, httpClient *http.Client) (*Bedrock, error) {
```
crspeller: Why take the service config and the region if the service config contains the region? (see where this is called from)

bots/bots.go
```diff
@@ -184,6 +183,14 @@ func (b *MMBots) getLLM(serviceConfig llm.ServiceConfig, botConfig llm.BotConfig
 		result = openai.NewAzure(config.OpenAIConfigFromServiceConfig(serviceConfig, botConfig.EnabledNativeTools), b.llmUpstreamHTTPClient)
 	case llm.ServiceTypeAnthropic:
 		result = anthropic.New(serviceConfig, botConfig.EnabledNativeTools, b.llmUpstreamHTTPClient)
+	case llm.ServiceTypeBedrock:
+		// For Bedrock, Region contains the AWS region
+		var err error
+		result, err = bedrock.New(serviceConfig, serviceConfig.Region, b.llmUpstreamHTTPClient)
+		if err != nil {
+			b.pluginAPI.Log.Error("Failed to create Bedrock client", "error", err, "bot_name", botConfig.Name)
```
crspeller: We might be double logging here as the error we return will also log right?

bedrock/bedrock.go
```diff
@@ -0,0 +1,550 @@
...
+	}
+	return converted
+}
+
+func (b *Bedrock) InputTokenLimit() int {
+	if b.inputTokenLimit > 0 {
+		return b.inputTokenLimit
+	}
+
+	// Return model-specific defaults
```
crspeller: Can we hit the AWS API for this? This list is not great with already out of date models.

bedrock/bedrock_test.go
```diff
@@ -0,0 +1,284 @@
...
+			b := &Bedrock{
+				defaultModel:    tt.defaultModel,
+				inputTokenLimit: tt.inputTokenLimit,
+			}
+
+			limit := b.InputTokenLimit()
+			assert.Equal(t, tt.expectedLimit, limit)
+		})
+	}
+}
```
crspeller, line 288: Seems excessive.

bedrock/bedrock_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+func TestCountTokens(t *testing.T) {
+	b := &Bedrock{}
+
+	tests := []struct {
+		name     string
+		text     string
+		minCount int
```
crspeller: Why a range? If we are going to test it we might as well get the right answer :P

bedrock/bedrock_test.go
```diff
@@ -0,0 +1,284 @@
...
+	})
+
+	t.Run("empty tools array", func(t *testing.T) {
+		tools := []llm.Tool{}
+		converted := convertTools(tools)
+		assert.Len(t, converted, 0)
+	})
+}
+
+func TestGetDefaultConfig(t *testing.T) {
```
crspeller, line 269: Excessive

bedrock/bedrock_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := isValidImageType(tt.mimeType)
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func TestConversationToMessages(t *testing.T) {
```
crspeller, line 41: Image content test?

crspeller (changes requested):
I wonder if we should add bedrock to the evals or e2e like we did for Mistral. So we can find out when the API breaks. (Can be this or separate PR)

bedrock/bedrock.go
nickmisasi: Bedrock doesn't expose an API endpoint for model metadata/limits. Simplified to return a sensible default (200k) and rely on users to configure `inputTokenLimit` in service config if needed for their specific model.

PR #431, opened 2025-11-07
Tool description/response improvements and better error handling

BenCookie95:
#### Summary
- We received some feedback that the create post tool needs to show more information to the end user of where the post is going. Previously it would only show a channel ID but it now also passes the channel display name and team name when creating a post so the user can make a better decision.
- Return all results when  get_channel_info returns more than 1 result.
- Don't return an MCP error when there are no results, this allows the llm to try again on it's own. I also added some suggestions to the no channel found response so the llm can try some variations of the display name or just try search by channel name

#### Release Note

```release-note
Tool response improvements
```

### DIFF
--- a/mcp/client_integration_test.go
+++ b/mcp/client_integration_test.go
@@ -102,8 +102,7 @@ func TestClient_CallTool_ReadChannel(t *testing.T) {
 
 	// Validate output content based on what read_channel tool returns
 	assert.NotEmpty(t, result, "Result should not be empty")
-	assert.Contains(t, result, "Found", "Result should contain 'Found' (part of output format)")
-	assert.Contains(t, result, "posts in channel", "Result should mention posts in channel")
+	assert.Contains(t, result, "Found 2 posts", "Result should show 'Found 2 posts' (system messages for user joining)")
 
 	t.Logf("Tool result:\n%s", result)
 }
--- a/mcpserver/tools_integration_test.go
+++ b/mcpserver/tools_integration_test.go
@@ -33,8 +33,10 @@ func TestMCPToolsIntegration(t *testing.T) {
 	t.Run("CreatePostTool", func(t *testing.T) {
 		t.Run("HappyPath", func(t *testing.T) {
 			args := map[string]interface{}{
-				"channel_id": testData.Channel.Id,
-				"message":    "Hello from MCP integration test!",
+				"channel_id":           testData.Channel.Id,
+				"channel_display_name": testData.Channel.DisplayName,
+				"team_display_name":    testData.Team.DisplayName,
+				"message":              "Hello from MCP integration test!",
 			}
 
 			result, err := executeToolWithMCP(t, suite, "create_post", args)


mcpserver/tools/posts.go
```diff
@@ -192,6 +217,32 @@ func (p *MattermostToolProvider) toolCreatePost(mcpContext *MCPToolContext, args
...
 	ctx := context.Background()
 
+	// Validate that the provided display names match the actual channel and team
+	channel, _, err := client.GetChannel(ctx, args.ChannelID, "")
+	if err != nil {
+		return "failed to validate channel", fmt.Errorf("error fetching channel for validation: %w", err)
+	}
+
+	// Check if channel display name matches
+	if channel.DisplayName != args.ChannelDisplayName {
```
BenCookie95, line 237: At first I thought this was a strange validation to add but when you have multiple channels returned from the get_channel_info tool is there a chance that the llm hallucinates and gives the wrong display name with the channelID?

mcpserver/tools/channels.go
```diff
@@ -254,44 +267,99 @@ func (p *MattermostToolProvider) toolGetChannelInfo(mcpContext *MCPToolContext,
...
-		// Direct ID lookup - fastest method
-		channel, _, err = client.GetChannel(ctx, args.ChannelID, "")
+		// Direct ID lookup - fastest method, always returns single result
+		var channel *model.Channel
+		var resp *model.Response
+		channel, resp, err = client.GetChannel(ctx, args.ChannelID, "")
 		if err != nil {
-			return "channel not found by ID", fmt.Errorf("error fetching channel by ID: %w", err)
+			// Check if it's a 404 (not found) - return success with message
+			if resp != nil && resp.StatusCode == 404 {
```
nickmisasi:
```suggestion
			if resp != nil && resp.StatusCode == http.StatusNotFound {
```

mcpserver/tools/channels.go
```diff
@@ -254,44 +267,99 @@ func (p *MattermostToolProvider) toolGetChannelInfo(mcpContext *MCPToolContext,
...
+			case args.ChannelDisplayName != "":
+				notFoundMsg.WriteString(fmt.Sprintf("No channels found with display name '%s'.", args.ChannelDisplayName))
+			default:
+				notFoundMsg.WriteString(fmt.Sprintf("No channels found with name '%s'.", args.ChannelName))
+			}
+
+			// Search scope
+			if args.TeamID != "" {
+				notFoundMsg.WriteString(" (searched within specified team)")
+			} else {
```
nickmisasi, line 324: Might be worth including the team name here, or at least the ID since we already have it

mcpserver/tools/posts.go
crspeller: It's also a gate for smarter LLMs that won't do that.

mcpserver/tools/channels.go
```diff
@@ -254,44 +267,99 @@ func (p *MattermostToolProvider) toolGetChannelInfo(mcpContext *MCPToolContext,
...
+			if args.ChannelDisplayName != "" {
+				// Tried display name - this is likely a case sensitivity issue
+				notFoundMsg.WriteString("1. Call get_channel_info again with different capitalization (channel_display_name is CASE-SENSITIVE)\n")
+				notFoundMsg.WriteString("2. Call get_channel_info with channel_name parameter instead (NOT case-sensitive)\n")
+			}
+
+			if args.ChannelName != "" && args.ChannelDisplayName == "" {
+				// Tried channel name only - suggest trying display_name
+				notFoundMsg.WriteString("1. Call get_channel_info with channel_display_name parameter instead (try different capitalizations as it's CASE-SENSITIVE)\n")
+			}
```
nickmisasi, line 339: not really saying we need to change it, but I worry this will cause the LLM to loop.

mcpserver/tools/channels.go
BenCookie95: Added

PR #438, opened 2025-11-12
[MM-66579] Inform LLMs about DM-only tool availability in channels

nickmisasi:
#### Summary

**Problem:** When users requested DM-only tool capabilities (like `create_post`) in public channels, the LLM would respond with "I don't have that capability" without explaining why or offering alternatives.

**Solution:** LLMs now receive information about disabled tools via system messages, enabling them to inform users that these capabilities are available in Direct Messages or via the Agents tab.

**Technical Implementation:**
- All tools (built-in MM tools + MCP tools) are now added to the `ToolStore` regardless of channel type for LLM awareness
- Security is enforced at execution time via `WithToolsDisabled()` option, which prevents tools from being sent to the LLM API in non-DM channels
- A system message is appended listing disabled tools with the instruction to respond concisely when users request these capabilities
- Refactored duplicate `buildDisabledToolsMessage()` functions into shared `llm.BuildDisabledToolsMessage()`

**Architecture Changes:**
- Removed `isDM` parameter from `MMToolProvider.GetTools()` and related methods
- Tools are filtered at execution time (via `WithToolsDisabled()`) rather than at definition time
- MCP tools now always added to ToolStore in `llmcontext.Builder.getToolsStoreForUser()`

#### Ticket Link

Jira https://mattermost.atlassian.net/browse/MM-66579

#### Screenshots

image

#### Release Note

```release-note
Improved AI bot responses when users request DM-only tool capabilities in channels. The bot now explains that these features are available in Direct Messages or via the Agents tab, rather than simply stating it lacks the capability.

### CODE REVIEW

File: llm/tools.go [hunk 1]
```diff
@@ -128,6 +128,39 @@ func (s *ToolStore) GetTools() []Tool {
...
+			Name:        tool.Name,
+			Description: tool.Description,
+		})
+	}
+	return result
+}
+
+// BuildDisabledToolsMessage creates a system message informing the LLM about tools
+// that are available in DM or Agents tab but not in the current context (e.g., a channel).
+func BuildDisabledToolsMessage(toolsInfo []ToolInfo) string {
```
crspeller: This feels like it should be done though the templates. I don't want to split up the prompt more than we have to. 

We can pass if tools are disabled and what they are to the template system and do this logic there.

anthropic/anthropic.go
```diff
@@ -208,11 +208,22 @@ func (a *Anthropic) streamChatWithTools(state messageState) {
 		params.Tools = convertTools(state.tools)
 	}
 
+	// Build system message, potentially including disabled tools info
+	systemMessage := state.system
+	if state.config.ToolsDisabled && len(state.config.DisabledToolsInfo) > 0 {
```
crspeller: This doesn't feel like a concern of each LLM provider. We should make what these have to handle as minimal as possible.

llm/tools.go
nickmisasi: Updated

anthropic/anthropic.go
nickmisasi: Fixed

anthropic/anthropic.go
```diff
@@ -208,11 +208,14 @@ func (a *Anthropic) streamChatWithTools(state messageState) {
 		params.Tools = convertTools(state.tools)
 	}
 
+	// Build system message, potentially including disabled tools info
```
crspeller: Remove leftover changes.

openai/openai.go
```diff
@@ -148,10 +148,12 @@ func NewCompatibleEmbeddings(config Config, httpClient *http.Client) *OpenAI {
 
 func modifyCompletionRequestWithRequest(params openai.ChatCompletionNewParams, internalRequest llm.CompletionRequest, cfg llm.LanguageModelConfig) openai.ChatCompletionNewParams {
 	params.Messages = postsToChatCompletionMessages(internalRequest.Posts)
+
```
crspeller: Remove leftover changes.

PR #443, opened 2025-11-17
Include channelIDs and rootIDs in the right places

BenCookie95:
#### Summary
We weren't including rootIDs in search results so the LLM couldn't always reply to threads. Also include ChannelIDs where appropriate.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/mcpserver/tools/posts.go
+++ b/mcpserver/tools/posts.go
@@ -170,8 +170,27 @@ func (p *MattermostToolProvider) toolReadPost(mcpContext *MCPToolContext, argsGe
 	// Format the response
 	var result strings.Builder
 	if channelName != "" && teamName != "" {
-		result.WriteString(fmt.Sprintf("Channel: %s (Team: %s)\n\n", channelName, teamName))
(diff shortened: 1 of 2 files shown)


mcpserver/tools/posts.go
```diff
@@ -170,9 +170,28 @@ func (p *MattermostToolProvider) toolReadPost(mcpContext *MCPToolContext, argsGe
...
+		result.WriteString(fmt.Sprintf("Channel ID: %s\n", posts[0].ChannelId))
+
+		// Find any post with a non-empty RootId - all replies share the same RootId
+		var rootID string
+		for _, post := range posts {
+			if post.RootId != "" {
+				rootID = post.RootId
+				break
+			}
+		}
```
nickmisasi, line 187: My understanding is that we will have either:
1. One post
2. A list of posts, sorted by creation date. 

So could we just:
1. Set the rootID when we loop through and sort earlier
2. Just grab the newest post's rootID

nickmisasi (approved):
Non-blocking suggetsion

Issue #393, opened 2025-10-06
🐛 bug: Reaction from "React for me" attributed to Copilot has led to abuse


jejbq:
## Description

<img width="312" height="144" alt="Image" src="https://github.com/user-attachments/assets/5e749aac-65a6-4c27-a2e7-fa342b690c41" />

The "React for me" feature in the Mattermost Agents Plugin currently adds reactions using the Copilot bot user. This is misleading and opens the door to potential abuse through impersonation, as AI-generated reactions appear to come from the bot rather than the actual user who initiated them. 

Additionally, the user who triggered the reaction cannot remove it, and since the reaction is owned by Copilot, no one else can either. It is also problematic that multiple reactions from Copilot can accumulate on the same post. This can frustrate users, especially when Copilot adds reactions to posts that are not intended to be humorous.

## Steps to reproduce

1. Trigger the AI "React for me" feature in the Mattermost Plugin.  
2. Observe that the reaction on the post is attributed to the Copilot bot user instead of the actual user.

## Expected Behavior

The AI-generated reaction should be attributed to the user who triggered it, rather than the bot, to prevent impersonation and confusion. This approach also allows the user to remove their own reaction, even if it was created by AI.

## Observed Behavior

The reaction is attributed to the Copilot bot user, allowing reaction impersonation.

## Suggested Patch

Changed the reaction UserId from the bot user to the actual user ID in the code to ensure reaction ownership reflects the user:

```diff
--- a/api/api_post.go
+++ b/api/api_post.go
@@ -88,7 +88,7 @@ func (a *API) handleReact(c *gin.Context) {
        // Add reaction to the post
        if err := a.pluginAPI.Post.AddReaction(&model.Reaction{
                EmojiName: emojiName,
-               UserId:    bot.GetMMBot().UserId,
+               UserId:    userID,
                PostId:    post.Id,
        }); err != nil {
                c.AbortWithError(http.StatusInternalServerError, fmt.Errorf("failed to add reaction: %w", err))
```

PR #449, opened 2025-11-24
MM-66111: Fix emoji react failing with OpenAI Responses API

nickmisasi:
Fixes emoji react returning empty responses when using OpenAI Responses API.

**Changes:**
- Add handler for `response.incomplete` event in Responses API streaming
- Increase max tokens from 25 to 500 for emoji selection

**Root cause:** The Responses API can return `response.incomplete` when tokens are exhausted before completion. Without a handler, the stream closed without events, resulting in empty responses. Certain models from Anthropic were also exhibiting this behaviour. I've chosen 500 as a safe limit that was consistently providing return results.


react/react.go
```diff
@@ -54,7 +54,9 @@ func (r *React) Resolve(message string, context *llm.Context) (string, error) {
 	}
 
 	// Get emoji from LLM
-	emojiName, err := r.llm.ChatCompletionNoStream(completionRequest, llm.WithMaxGeneratedTokens(25))
+	// Note: Using 1000 tokens to accommodate OpenAI Responses API overhead
```
davidkrauser, line 57: **question:** is 1000 the right value here? You have `llm.WithMaxGeneratedTokens(500)` below.

openai/openai.go
```diff
@@ -808,6 +808,53 @@ func (s *OpenAI) streamResponsesAPIToChannels(params openai.ChatCompletionNewPar
...
+				handleToolCalls()
+				return
+			}
+
+			// Send end event even for incomplete responses
+			output <- llm.TextStreamEvent{
+				Type:  llm.EventTypeEnd,
+				Value: nil,
+			}
+			return
```
davidkrauser, line 832: **question:** would it be better to emit some sort of error in this case instead? (and show an error message to the user?) Otherwise, the user may not understand something went "wrong" - they'll just see strangely truncated output (or maybe no output)?

That would make it obvious they need to try again (clicking regenerate), or they need to have an admin look at logs to fix a potential issue with token limits.

react/react.go
nickmisasi:
```suggestion
	// Note: Using 500 tokens to accommodate OpenAI Responses API overhead
```

openai/openai.go
nickmisasi: Yeah that's a good point. I've updated it so that the token logging is still there (because those tokens still count) but it will return an error

davidkrauser: > I've updated it so that the token logging is still there (because those tokens still count) but it will return an error

Awesome - thank you 🙂 

The token metric data (if I remember correctly) comes in the last packet. If we return here, will we miss that packet? Or does it come along with the `response.incomplete` message?

nickmisasi: The docs aren't exactly clear on this, but from testing: 

`response.incomplete` is the final event. No `response.completed` comes afterwards. 
There _is_ token data, but only input token, not output token:
```
{"timestamp":"2025-11-25 09:42:12.574 -05:00","level":"info","msg":"Token Usage","user_id":"n64tn7yzn7rp5dicw3jqm8g76c","team_id":"w3hbm6fzm7b1dbpa53buxmquxa","bot_username":"matty","input_tokens":829,"output_tokens":0,"total_tokens":829}
```
Worst case, this is done defensively, so it can either be there or not, and it'll be fine.

api/api_post.go
```diff
@@ -84,10 +84,10 @@ func (a *API) handleReact(c *gin.Context) {
 		return
 	}
 
-	// Add reaction to the post
+	// Add reaction to the post as the requesting user
 	if err := a.pluginAPI.Post.AddReaction(&model.Reaction{
 		EmojiName: emojiName,
-		UserId:    bot.GetMMBot().UserId,
+		UserId:    userID,
```
crspeller, line 90: Functional change, but I think we want this :P Fixes https://github.com/mattermost/mattermost-plugin-agents/issues/393

crspeller (approved):

api/api_post.go
nickmisasi: This was actually what MM-66111 ticket was originally for 🤣

PR #460, opened 2025-12-09
[MM-66591] Channel Summarization

nickmisasi:
#### Summary
Adds support for Channel Summarization/Analysis through the Channel Header "Ask Agents About This Channel" button. This is implemented by leveraging the embedded MCP Server, with the new addition of the concept of auto run tools (MCP tools that the LLM is allowed to run without requesting permission). 

A user provides either a timeline, or specific instructions, and the LLM is given the context of what channel/team/user requested the summary along with those specific instructions. The LLM is given auto run access to the read_channel and get_channel_info tools, and will leverage them to fetch the posts necessary to analyze a channel based on the provided criteria from the user. 

Citations are included

This branch also includes a significant refactor of both Anthropic and OpenAI implementations in order to support looped tool calling without needing to go back to the user for permission. 

**OpenAI**
- This primarily involved a massive single for loop function with hundreds of lines of code (that was even longer after adding auto run tool support). I've split up the code drastically, and tightened up repetitiveness between the Responses and Completion APIs. 
**Anthropic**
- Refactored to use more of a state-machine pattern. 

Bedrock also received support for auto run tools. It didn't have much code already, so it was mostly merging the functionality. 

As we move towards an Agents V2, I expect us to move the Agentic looping out of each of the providers, ideally into an existing framework. The scope of this refactor was pretty large, so I didn't attempt it for this PR.
The Anthropic and OpenAI diffs are a bit tough to read on Github. I recommend looking at the files holistically, rather than the diffs. They read much easier.

Related to: 
(Webapp) https://github.com/mattermost/mattermost/pull/34687
(Mobile) https://github.com/mattermost/mattermost/pull/34881

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66591

#### Release Note

```release-note
Support for Channel Analysis 
```

---

> [!NOTE]
> Enables channel-level summarization/analysis driven by MCP tools with auto-run capability and provider streaming refactors.
> 
> - **API/Backend**: New `POST /api/channel/:channelid/analyze` with `handleChannelAnalysis`; introduces `channels.AnalyzeChannel` that binds `read_channel`/`get_channel_info` MCP tools and streams results; minor title constant adjustments
> - **Anthropic**: Major refactor into smaller helpers (message building, stream event handling, tool call/result extraction), optional native `web_search`, improved image handling, annotations extraction, and reasoning budget calculation; supports auto-run tool loops
> - **Bedrock**: Implements auto-run tool flow and streaming helpers (tool call parsing, assistant/tool-result message builders) with comprehensive unit tests
> - **CI/E2E**: CI sets `MM_IMAGE`, excludes real API tests by default, and adds a real-API Channel Analysis job; upgrades Playwright to stable; adds helpers and multiple Channel Analysis test suites (quick actions, custom queries, integration, real API) with MCP enabled in test plugin config
> - **Misc**: Updates internal Claude agents configs (model/tool tweaks) and e2e ignores
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f8f1e41251bddff9917a396265afbb32df11938b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/webapp/src/components/assets/icon_cancel.tsx
+++ b/webapp/src/components/assets/icon_cancel.tsx
@@ -7,8 +7,8 @@ import Svg from '../svg';
 
 const IconCancel = () => (
     <Svg
-        width='10'
-        height='10'
+        width='20'
+        height='20'
         viewBox='0 0 10 10'
         fill='none'
         xmlns='http://www.w3.org/2000/svg'
@@ -16,7 +16,6 @@ const IconCancel = () => (
         <path
             d='M9.2048 1.6546L5.8496 5.0098L9.2048 8.365L8.3552 9.2146L5 5.8594L1.6448 9.2146L0.7952 8.365L4.1504 5.0098L0.7952 1.6546L1.6448 0.805005L5 4.1602L8.3552 0.805005L9.2048 1.6546Z'
             fill='currentColor'
-            fillOpacity='0.64'
         />
     </Svg>
 );
--- a/webapp/src/components/summarize_date_range_modal.tsx
+++ b/webapp/src/components/summarize_date_range_modal.tsx
@@ -3,9 +3,11 @@
 
 import React from 'react';
 import styled from 'styled-components';
-import {FormattedMessage} from 'react-intl';
+import {FormattedMessage, useIntl} from 'react-intl';
 
-import IconCancel from './assets/icon_cancel';
+import {CloseIcon} from '@mattermost/compass-icons/components';
+
+import {DatePicker} from '@/mm_webapp';
 
 const ModalOverlay = styled.div`
     position: fixed;
@@ -68,16 +70,16 @@ const CloseButton = styled.button`
     background: none;
     border: none;
     cursor: pointer;
-    padding: 4px;
+    padding: 0;
     border-radius: 4px;
-    color: rgba(var(--center-channel-color-rgb), 0.56);
+    color: var(--center-channel-color);
+    opacity: 0.64;
     display: flex;
     align-items: center;
     justify-content: center;
 
     &:hover {
         background: rgba(var(--center-channel-color-rgb), 0.08);
-        color: rgba(var(--center-channel-color-rgb), 0.72);
     }
 `;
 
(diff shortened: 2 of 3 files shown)


server/embedded_mcp_server.go
```diff
@@ -32,12 +32,41 @@ func NewEmbeddedMCPServer(pluginAPI *pluginapi.Client, logger pluginapi.LogServi
...
+		"internalServerURL", internalServerURL)
+
 	// Create configuration for in-memory transport
 	config := mcpserver.InMemoryConfig{
 		BaseConfig: mcpserver.BaseConfig{
 			MMServerURL: siteURL,
-			// Just use siteURL for now. Could make this configurable later
-			MMInternalServerURL: siteURL,
+			// Use the internal server URL for API communication within the container
+			MMInternalServerURL: internalServerURL,
```
nickmisasi, line 69: Prior to this we have been executing MCP API requests against the SiteURL, which meant that they would always go out to the public internet, through reverse proxy/load balancer/etc and then hit Mattermost. 

This caused issues for Playwright testing - siteURL is `http://localhost:11234`, but that's because 11234 is mapped to 8065 on the docker container. on the container, localhost 11234 has nothing listening on it.

I think this is overall actually an improvement, because it's just a loopback call that can happen purely within the container itself. No need to go out to the public net. Open to reverting and having a custom env setup for when playwright runs if there are any concerns. 

cc @BenCookie95

e2e/package.json
```diff
@@ -15,12 +15,11 @@
   "devDependencies": {
     "@mattermost/client": "10.0.0",
     "@mattermost/types": "10.0.0",
-    "@playwright/mcp": "^0.0.46",
```
nickmisasi, line 18: The actual mcp test server is part of core playwright. Different toolsets. This one is more about pure browser control. The other has proper test plan/debug/healer tooling.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,9 @@ jobs:
   e2e:
     timeout-minutes: 60
     runs-on: ubuntu-latest
+    env:
+      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:latest' }}
```
crspeller, line 18: Let's do a specific tag and not latest. As much as it is nice to see breaks when they happen, we would really need to run the e2e tests twice for this. We used to have this set for Playbooks and you would be surprised how many breaking changes we make to master on the server 😆

llm/tools.go
```diff
@@ -119,13 +120,152 @@ func (tc *ToolCall) SanitizeArguments() {
 
 type ToolArgumentGetter func(args any) error
 
+// MergeArguments creates a new ToolArgumentGetter that merges overrides into the original arguments.
+func MergeArguments(originalGetter ToolArgumentGetter, overrides map[string]interface{}) ToolArgumentGetter {
```
crspeller: This approach to overriding the I found to be confusing. Suggested improvements: https://github.com/mattermost/mattermost-plugin-agents/pull/479

prompts/citation_format.tmpl
```diff
@@ -0,0 +1,26 @@
...
+[permalink]({{.SiteURL}}/{{if .Team}}{{.Team.Name}}{{else}}_redirect{{end}}/pl/<post_id>?view=citation)
+
+To cite a channel:
+[permalink]({{.SiteURL}}/{{if .Team}}{{.Team.Name}}{{else}}_redirect{{end}}/channels/<channel_name>?view=citation)
+
+To cite a team:
+[permalink]({{.SiteURL}}/{{if .Team}}{{.Team.Name}}{{else}}_redirect{{end}}?view=citation)
+
+
+For example, if you return the text "John smith was sad no one attended his birthday party.", from post with ID abc123, on team 'missionops', where site URL is https://your-mattermost-url, your result should look like this:
```
crspeller, line 24: Poor guy 😆

.github/workflows/ci.yml
nickmisasi: `latest` in the context of mattermost/mattermost-enterprise-edition actually means "latest commit from the most recent release branch", not master. I think there's value in having this not be hard coded to say release-11.3 until we update the code. MM_IMAGE to override if necessary

llm/tools.go
nickmisasi: Thanks!

crspeller (approved):

nickmisasi:
image

api/api_channel.go
```diff
@@ -50,6 +49,130 @@ func (a *API) channelAuthorizationRequired(c *gin.Context) {
...
+			hasReadChannel = true
+		}
+	}
+
+	if !hasReadChannel {
+		a.pluginAPI.Log.Error("Channel analysis failed: read_channel tool not available",
+			"userID", userID,
+			"channelID", channel.Id,
+			"availableTools", toolNames)
+		c.AbortWithError(http.StatusInternalServerError, fmt.Errorf("channel analysis requires read_channel tool which is not available (found %d tools: %v) - ensure embedded MCP server is enabled and working", len(availableTools), toolNames))
```
BenCookie95, line 130: I find it a bit odd that we need the embedded MCP server to be enabled in order to use it under the hood? Should it always be available for features that are using pre-determined tools like this is doing?

ogi-m (approved):
LGTM!

PR #472, opened 2025-12-15
Add Docker image build for MCP server

nickmisasi:
## Summary
- Adds CI workflow to build and push MCP server Docker image to Docker Hub on release tags
- Adds multi-stage Dockerfile for standalone MCP server container (linux/amd64 and linux/arm64)
- Adds `make mcp-server-docker` target for local builds
- Adds documentation for Docker-based deployment

## Test plan
- [ ] Verify local build with `make mcp-server-docker`
- [ ] Test STDIO mode with Claude Code/Desktop
- [ ] Test HTTP mode connectivity

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,3 +292,36 @@ jobs:
                  -F "force=true" \
                  "$url/api/v4/plugins"
           done
+
+  docker-mcp-server-build-push:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: ci/setup-qemu
+        uses: docker/setup-qemu-action@v3
+
+      - name: ci/setup-buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: ci/docker-login
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: ci/extract-version
+        id: version
+        run: echo "version=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
+
+      - name: ci/build-push-mcp-server
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./mcpserver/Dockerfile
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -292,3 +292,36 @@ jobs:
                  -F "force=true" \
                  "$url/api/v4/plugins"
           done
+
+  docker-mcp-server-build-push:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/checkout@v4
```
NARSimoes, line 300: I'd consider using `commit SHA` pointing to a specific commit for better reproducibility. For example, instead of  `actions/checkout@v6.0.1` we can use something like `actions/checkout@ 8e8c483db84b4bee98b60c0593521ed34d9990e8  #v6.0.1`. The same applies for there steps.

.github/workflows/ci.yml
```diff
@@ -292,3 +292,36 @@ jobs:
...
+      - name: ci/docker-login
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: ci/extract-version
+        id: version
+        run: echo "version=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
+
```
NARSimoes, line 317: What do you think about adding a step to scan the Dockerfile & images on pull-request and before pushing to the repository. With that we can catch issues earlier in CI and before those images are published.

.github/workflows/ci.yml
```diff
@@ -292,3 +292,36 @@ jobs:
                  -F "force=true" \
                  "$url/api/v4/plugins"
           done
+
+  docker-mcp-server-build-push:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/v')
```
NARSimoes, line 298: What do you think about breaking this down to also validate the Dockerfile in pull-requests ? For example, something like:
- On `pull-requests`: checkout -> build -> scan / lint / test Dockerfile & images
- On `github.ref, 'refs/tags/v'`: push the images

NARSimoes (changes requested):
I'm just adding a few points for discussion - overall this looks good.

PR #479, opened 2026-01-16
Improve parameter overrides

crspeller:
PR against #460 for suggestions.

### DIFF
--- a/llm/tools.go
+++ b/llm/tools.go
@@ -311,19 +311,17 @@ func ExecuteAutoRunTools(
 	results := make([]AutoRunResult, 0, len(pendingToolCalls))
 
 	for _, tc := range pendingToolCalls {
-		// Capture tc in the closure to avoid issues with loop variable
-		toolCall := tc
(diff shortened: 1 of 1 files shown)


channels/channels.go
```diff
@@ -70,6 +70,24 @@ func (c *Channels) AnalyzeChannel(
...
+	boundReadChannel := readChannel.WithBoundParams(map[string]interface{}{"channel_id": channelID})
+
+	getChannelInfo := context.Tools.GetTool("get_channel_info")
+	if getChannelInfo == nil {
+		return nil, fmt.Errorf("get_channel_info tool not available")
+	}
+	boundGetChannelInfo := getChannelInfo.WithBoundParams(map[string]interface{}{"channel_id": channelID})
+
+	// Create scoped tool store with bound tools
+	scopedTools := llm.NewToolStore(nil, false)
```
crspeller, line 87: This is a behavioral change since this removes all the other tools. I think this might have been what you intended given the prompt so maybe it's better like this?

PR #487, opened 2026-01-28
Add is_default field to bridge agent info

nickmisasi:
#### Summary
- Add `is_default` field to `BridgeAgentInfo` to enable bridge clients to identify the default agent configured in the plugin.
- Set default profile picture for all bots/agents that don't have a custom image uploaded.

#### Details
**is_default field:**
- Bridge clients can now identify which agent is the default via the `/agents` endpoint response.

**Default profile picture:**
- Bots without a custom profile picture will automatically get the default agent profile picture.
- Uses `LastPictureUpdate` field to detect if a custom image has been uploaded - only sets the default if no custom image exists.
- This applies to both new bots and existing bots that never had a profile picture set.
- Custom images uploaded by admins are never overwritten.
- Updated `bot_icon.png` (plugin icon) to match the new default profile picture.

#### Ticket Link
MM-67343

#### Release Note
```release-note
- Added is_default field to bridge API /agents endpoint response.
- Bots now automatically receive a default profile picture if no custom image has been set.
```

## Summary by CodeRabbit

- New Features
  - Agent list now includes a new is_default flag to indicate the default agent.
  - Bots without a profile image automatically receive a default profile picture.
  - Added an embedded default agent icon to ensure consistent visuals out of the box.

- Tests
  - Expanded bot creation tests to validate automatic assignment of default profile images.


crspeller (approved):

nickmisasi:
@crspeller figured I'd tackle the icon thing in this PR as well. I'll update the PNG if it changes but can do that without re review

bots/bots.go
```diff
@@ -241,6 +246,22 @@ func (b *MMBots) EnsureBots() error {
...
 }
 
+func (b *MMBots) ensureDefaultProfileImage(bot *Bot) {
+	user, err := b.pluginAPI.User.Get(bot.mmBot.UserId)
+	if err != nil {
+		b.pluginAPI.Log.Error("Failed to get bot user for profile image check", "bot_name", bot.cfg.Name, "error", err.Error())
+		return
+	}
+
+	if user.LastPictureUpdate != 0 {
```
crspeller, line 256: Will this work if they set the profile picture immediately on creation?

bots/bots.go
```diff
@@ -224,6 +226,9 @@ func (b *MMBots) EnsureBots() error {
 				continue
 			}
 		}
+
+		b.ensureDefaultProfileImage(bot)
```
crspeller, line 230: Do we really need to set this for existing bots? I think maybe just changing the default is enough.

assets/bot_icon.png
crspeller, line 1: This looks smaller and differently shaped. I think you might need to ask Assad for one that goes here.

bots/bots.go
nickmisasi: Sure does. 

Image

The bot has already been created with the pp by the time this runs.

bots/bots.go
nickmisasi: I think we should. The default currently is the bot icon, not the agents icon, so it's even more confusing IMO

assets/bot_icon.png
nickmisasi: I exported it bigger, but still not quite as big. Big enough? Not sure. I think the square doesn't really matter since it's always shown in the bubble. But this may just be a placeholder anyways.

bots/bots.go
crspeller: Well ok then!

PR #506, opened 2026-02-16
MM-63635: Migrate Agents plugin to registerAppBarComponent API

crspeller:
#### Summary
Migrates the Agents plugin from the legacy `registerRightHandSidebarComponent` + `registerChannelHeaderButtonAction` pattern to the unified `registerAppBarComponent` API.

When the App Bar is enabled (the default), Mattermost already promotes channel-header plugin buttons into the App Bar and removes the duplicate channel-header icon, so the previous `registerChannelHeaderButtonAction` registration was effectively redundant. This change registers the App Bar component explicitly so the plugin gets consistent App Bar integration with the correct icon and tooltip, and removes the redundant/dead code.

**Changes:**
- Replaced `registerRightHandSidebarComponent` with `registerAppBarComponent`, passing the RHS component and title through the App Bar registration
- Removed the now-redundant `registerChannelHeaderButtonAction` call
- Removed the unused `ChannelHeaderIcon` component and `useBotlist` import
- Used the bundled `aiIcon` asset for the App Bar icon

#### Ticket Link
Jira https://mattermost.atlassian.net/browse/MM-63635

#### Release Note
```release-note
NONE
```


nickmisasi (approved):
LGTM - question - will this still respect the custom ordering we have to bubble the Agents icon to the top of the list?

PR #510, opened 2026-02-19
Harden embedding search: indexing, search, citations, and admin tooling

crspeller:
#### Summary

Refinement of the embedding search system focusing on reliability, operability, and feature completeness.

**Indexer hardening**
- Resumable indexing with cursor-based checkpointing (saves every 500 posts) and two-phase reindex (main pass + catch-up for posts created during reindex)
- Stale job detection with 10-minute heartbeat threshold to surface crashed jobs
- Node-aware orphaned job recovery: on plugin restart, marks jobs that were running on the current node as failed
- Model compatibility tracking: stores provider/model/dimensions at index time and detects mismatches on config change, disabling search until reindex
- Data retention support via `RunDataRetention()` hook that cleans up embeddings for permanently deleted posts
- Cluster-wide mutex to prevent concurrent reindex jobs in HA deployments
- Retry logic with exponential backoff for both single-post and batch indexing

**Search improvements**
- Post citation system: LLM embeds `!!CITE#!!` markers in responses, parsed into structured annotations with rune-aware byte offset tracking for Unicode support
- Results enriched with channel names, usernames, and chunk metadata
- Both async (streaming to posts) and sync (immediate response) search modes

**MCP semantic search tool**
- `search_posts` tool exposed via MCP with dual search strategy: runs semantic and keyword searches in parallel with deduplication
- Graceful degradation to keyword-only search when semantic search is unavailable
- HTTP callback endpoint (`/api/v1/mcp-server/semantic-search`) for external MCP servers to access semantic search

**pgvector storage layer**
- Chunking support with chunk ID generation, pre-insertion cleanup of stale chunks, and chunk metadata columns
- Orphaned embedding cleanup with batched deletion respecting soft-delete retention
- Search filtering: time-range, team/channel scoping, minimum similarity score, pagination
- Permission enforcement at the SQL layer via channel membership joins
- Input validation and search result limits

**Plugin lifecycle**
- Atomic pointer for thread-safe embedding search access without locks
- Config change listener that reinitializes embedding search, checks model compatibility, and atomically swaps the search instance
- Post lifecycle hooks: index on create, delete+reindex on update, delete on remove

**Admin API & frontend**
- Admin endpoints: reindex (with clear/resume), cancel, catch-up, health check, model compatibility, stale job status
- Reindex management UI with real-time progress bar, stale job warnings, model mismatch alerts, and health check display
- Post citation component with hover tooltip (author + channel) and click-to-navigate to source post
- Job status polling (2s interval while running) with automatic health check refresh on completion

**Testing**
- Unit tests across indexer, search, citations, pgvector, MCP tools, admin API, embeddings, prompts, and chunking
- Integration tests for the embedding pipeline
- Search eval tests
- Playwright e2e tests for search entry points, citations, sources, and bot selector

#### Ticket Link

Jira https://mattermost.atlassian.net/browse/MM-65668

#### Screenshots

N/A

#### Release Note
```release-note
Hardened embedding search with resumable indexing, post citations, data retention support, model compatibility detection, MCP semantic search tool, and admin management UI.
```

## Summary by CodeRabbit

* **New Features**
  * Post citations in search responses with clickable badges, tooltips and navigation
  * Raw semantic search endpoint and Agents/embedding search UI (bot selector)
  * Index health checks, catch‑up indexing, resume/clear reindex controls, and orphaned‑embedding cleanup

* **Improvements**
  * Stronger search validation, pagination/clamping, relevance scoring and combined semantic+keyword results
  * Safer rendering and sanitized LLM context fields; escaped search templates
  * Expanded system console reindex UI with health, staleness and model‑mismatch warnings

* **Tests**
  * Large expansion of unit and end‑to‑end test coverage for search, embeddings, indexing and tooling

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     env:
-      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:latest' }}
+      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:release-11' }}
       EXCLUDE_REAL_API_TESTS: 'true'
     defaults:
       run:
@@ -51,7 +51,7 @@ jobs:
     timeout-minutes: 180
     runs-on: ubuntu-latest
     env:
-      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:latest' }}
+      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:release-11' }}
     defaults:
       run:
         working-directory: ./e2e
@@ -89,7 +89,7 @@ jobs:
     timeout-minutes: 180
     runs-on: ubuntu-latest
     env:
-      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:latest' }}
+      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:release-11' }}
     defaults:
       run:
         working-directory: ./e2e
@@ -127,7 +127,7 @@ jobs:
     timeout-minutes: 180
     runs-on: ubuntu-latest
     env:
-      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:latest' }}
+      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:release-11' }}
     defaults:
       run:
         working-directory: ./e2e
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -481,7 +481,7 @@ func (a *Anthropic) extractAnnotations(message anthropicSDK.Message) []llm.Annot
 		}
 
 		startPos := textPosition
-		endPos := textPosition + len(textBlock.Text)
+		endPos := textPosition + llm.UTF16CodeUnitCount(textBlock.Text)
 		textPosition = endPos
 
 		for _, citation := range textBlock.Citations {
--- a/anthropic/anthropic_test.go
+++ b/anthropic/anthropic_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"testing"
+	"unicode/utf16"
 
 	anthropicSDK "github.com/anthropics/anthropic-sdk-go"
 	"github.com/google/jsonschema-go/jsonschema"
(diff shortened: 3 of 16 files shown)


indexer/indexer_job.go
```diff
@@ -181,17 +268,218 @@ func (s *Indexer) runReindexJob(jobStatus *JobStatus) {
 		}
 	}
 
+	// Run catch-up pass to index posts created during the main reindex
+	catchUpCount, catchUpErr := s.runCatchUpPass(ctx, jobStatus, search)
+	if catchUpErr != nil {
+		s.handleJobError(jobStatus, fmt.Sprintf("Catch-up pass failed: %s", catchUpErr), 0, "")
```
nickmisasi: The initial pass fail sets the lastCreateAt, but this one sets it to 0, so this means you only get one catch up pass and if it fails it'll start over. Could be intentional, but flagging

prompts/search_system.tmpl
```diff
@@ -4,7 +4,7 @@ Follow these guidelines:
 1. Answer questions directly and concisely based ONLY on the information in the provided context.
 2. If the context doesn't contain sufficient information to answer the question, clearly state this and don't make up information.
 3. Quote relevant parts of the context to support your answers when appropriate.
-4. Provide specific references to which messages contain the information, including the person's name and channel (e.g., "According to Jane Smith in Engineering Channel").
+4. IMPORTANT: When citing information from the search results, use the exact format !!CITE1!!, !!CITE2!!, etc. based on the message index. These markers will be automatically converted to clickable citation links. Do NOT write natural language references like "According to John in General channel".
```
nickmisasi: nit:

I'm ok with leaving the !!CITE1!! markers for initial pass. But, since this will only be citing internal Mattermost resources, we can use the https://github.com/mattermost/mattermost-plugin-agents/blob/master/prompts/citation_format.tmpl that's got built in support within core product. 

If it's a drop-in change we should go for it - if not, ok to defer.

indexer/indexer_job.go
```diff
@@ -181,17 +268,218 @@ func (s *Indexer) runReindexJob(jobStatus *JobStatus) {
 		}
 	}
 
+	// Run catch-up pass to index posts created during the main reindex
+	catchUpCount, catchUpErr := s.runCatchUpPass(ctx, jobStatus, search)
```
nickmisasi: If I'm reading this right, I think there's a potential problem where because this is unbounded, a heavy traffic server could have this job just running indefinitely - especially considering there's an inherent slowdown from the embeddings processing

server/main.go
```diff
@@ -159,19 +189,84 @@ func (p *Plugin) OnActivate() error {
...
+		}
+
+		// Check model compatibility
+		if newEmbeddingsSearch != nil {
+			embeddingsCfg := p.configuration.EmbeddingSearchConfig()
+			compatibility := indexerService.CheckModelCompatibility(embeddingsCfg.GetProviderType(), embeddingsCfg.Dimensions, embeddingsCfg.GetModelName())
+			if !compatibility.Compatible {
+				pluginAPI.Log.Warn("Embedding model configuration has changed, search disabled until re-index",
+					"reason", compatibility.Reason)
+				newEmbeddingsSearch = nil
```
nickmisasi, line 261: If this gets set to nil here, I think you end up with recovery deadlock - it will never be able to enter this branch again

server/main.go
```diff
@@ -159,19 +189,84 @@ func (p *Plugin) OnActivate() error {
...
+	}
+
+	// Check model compatibility and disable search if model has changed
+	if embeddingsSearch != nil {
+		embeddingsCfg := p.configuration.EmbeddingSearchConfig()
+		compatibility := indexerService.CheckModelCompatibility(embeddingsCfg.GetProviderType(), embeddingsCfg.Dimensions, embeddingsCfg.GetModelName())
+		if !compatibility.Compatible {
+			pluginAPI.Log.Warn("Embedding model configuration has changed, search disabled until re-index",
+				"reason", compatibility.Reason)
+			embeddingsSearch = nil // Disable search
```
nickmisasi, line 214: Same here

prompts/search_system.tmpl
crspeller: Done.

server/main.go
crspeller: Not a deadlock, added a comment.

PR #517, opened 2026-02-24
Add Scale AI support with OpenAI-compatible provider registry

nickmisasi:
#### Summary

Adds Scale AI as a new LLM provider and introduces a backend provider registry pattern for OpenAI-compatible services.

- **Provider registry** (`llm/providers.go`): OpenAI-compatible providers declare their config (known models, custom transport, API options) in a single map. Adding a future provider requires only a new map entry — no changes to `bots.go` or `api.go`.
- **Scale AI support**: Custom auth transport (`x-api-key` header instead of Bearer), static known models served from backend via `POST /admin/models/fetch`, configurable Account ID for ScaleGov.
- **Frontend simplification**: Removed hardcoded `scaleKnownModels` array from `service.tsx`. Scale now uses the same dynamic model dropdown as OpenAI/Anthropic, populated by the backend registry.

#### Ticket Link

N/A

#### Screenshots

N/A — UI behavior is unchanged; Scale model dropdown now loads from backend instead of hardcoded frontend list.

#### Release Note
```release-note
Added Scale AI as a new LLM provider with OpenAI-compatible API support.
```

## Summary by CodeRabbit

* **New Features**
  * Added Scale AI as a selectable LLM provider with localized UI labels, account-ID field, API URL help, and model selection (combobox when models available).
  * Provider-aware model discovery populates known models and sets a sensible default.

* **Bug Fixes**
  * Improved credential and transport handling to avoid hard-failing when provider-specific auth is used; adjusted API key requirements for supported providers.

* **Documentation**
  * Added Scale AI docs and configuration guidance.

* **Tests**
  * Added unit tests for provider registry, transport behavior, and Scale validation.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -214,12 +214,12 @@ func OpenAIConfigFromServiceConfig(serviceConfig llm.ServiceConfig, botConfig ll
 	}
 }
 
-// OpenAIConfigFromServiceConfigWithOptions creates an OpenAI config with additional options for OpenAI-compatible APIs
+// OpenAIConfigFromServiceConfigWithOptions creates an OpenAI config with additional options for OpenAI-compatible APIs.
+// SendUserID is preserved from the operator's service config rather than overridden here;
+// providers that don't support it should hide the toggle in the UI so it defaults to false.
 func OpenAIConfigFromServiceConfigWithOptions(serviceConfig llm.ServiceConfig, botConfig llm.BotConfig, disableStreamOptions bool, useMaxTokens bool) openai.Config {
 	cfg := OpenAIConfigFromServiceConfig(serviceConfig, botConfig)
 	cfg.DisableStreamOptions = disableStreamOptions
 	cfg.UseMaxTokens = useMaxTokens
-	// OpenAI-compatible APIs typically don't support the 'user' parameter
-	cfg.SendUserID = false
 	return cfg
 }


api/api.go
```diff
@@ -407,8 +409,12 @@ func (a *API) handleFetchModels(c *gin.Context) {
 	case "openai", "azure", "openaicompatible":
 		models, err = openai.FetchModels(req.APIKey, req.APIURL, req.OrgID, a.llmUpstreamHTTPClient)
 	default:
-		c.AbortWithError(http.StatusBadRequest, fmt.Errorf("model fetching not supported for service type: %s", req.ServiceType))
-		return
+		if providerCfg, ok := llm.GetOpenAICompatibleProvider(req.ServiceType); ok {
```
crspeller: Nit: bit awkward to call this twice. Just save the result above and use it here.

docs/providers.md
```diff
@@ -261,6 +262,41 @@ Obtain a [Mistral API key](https://console.mistral.ai/api-keys/), then select **
...
+### Configuration Options
+
+| Setting | Required | Description |
+|---------|----------|-------------|
+| **API Key** | Yes | Your Scale AI API key (sent as `x-api-key` header) |
+| **API URL** | Yes | Your Scale API endpoint (e.g., `https://sgp-api.scalegov.com/v5`) |
+| **Account ID** | No | Your Scale account ID (sent as `x-selected-account-id` header, required for ScaleGov) |
+| **Default Model** | Yes | The model to use by default in `vendor/model-name` format (e.g., `openai/gpt-4o`) |
+
+### Known Models
```
crspeller: Let's not list models, that will get out of date too fast. A link to their docs is enough.

llm/providers.go
```diff
@@ -0,0 +1,70 @@
...
+	// UseMaxTokens uses max_tokens instead of max_completion_tokens.
+	UseMaxTokens bool
+
+	// FixedAPIURL overrides the service config's APIURL
+	// (for providers like Cohere/Mistral with fixed endpoints).
+	FixedAPIURL string
+}
+
+// openAICompatibleProviders is the registry of known OpenAI-compatible providers.
+var openAICompatibleProviders = map[string]OpenAICompatibleProvider{
```
crspeller, line 32: We should migrate our existing OpenAI compatible models to this form.

llm/providers.go
```diff
@@ -0,0 +1,70 @@
...
+	ServiceTypeScale: {
+		KnownModels: []ModelInfo{
```
crspeller: I don't think this provides much value. It will go out of date too quickly. Let's remove it.

llm/providers.go
```diff
@@ -0,0 +1,70 @@
...
+			{ID: "bedrock/amazon.nova-lite-v1:0", DisplayName: "bedrock/amazon.nova-lite-v1:0"},
+		},
+		DefaultModel:         "openai/gpt-4o",
+		DisableStreamOptions: true,
+		CreateTransport: func(cfg ServiceConfig, base http.RoundTripper) http.RoundTripper {
+			headers := map[string]string{"x-api-key": cfg.APIKey}
+			if cfg.OrgID != "" {
+				headers["x-selected-account-id"] = cfg.OrgID
+			}
+			return &CustomAuthTransport{
```
crspeller, line 49: Maybe this shouldn't be generic in llm but a Scale specific thing. Not sure we will actually reuse this.

api/api.go
nickmisasi: Fixed — removed the double call entirely. Scale no longer participates in model fetching, so the `isRegistered` check and the `default` branch returning `KnownModels` are both gone.

docs/providers.md
nickmisasi: Removed the model list entirely. Replaced with a brief note about `vendor/model-name` format and a link to Scale AI docs.

llm/providers.go
nickmisasi: Done — migrated Cohere and Mistral into the provider registry. Their hardcoded `case` branches in `bots.go` are removed; they now go through the same `default` registry path as Scale.

docs/providers.md
nickmisasi: Removed the model list entirely. Replaced with a brief note about `vendor/model-name` format and a link to Scale AI docs.

llm/providers.go
nickmisasi: Done — migrated Cohere and Mistral into the provider registry. Their hardcoded `case` branches in `bots.go` are removed; they now go through the same `default` registry path as Scale.

llm/providers.go
nickmisasi: Removed `KnownModels` from the struct and all references. The system console now shows a plain text input with inline help text documenting the `vendor/model-name` format (e.g., `openai/gpt-4o`) instead.

nickmisasi: Removed `KnownModels` from the struct and all references. The system console now shows a plain text input with inline help text documenting the `vendor/model-name` format (e.g., `openai/gpt-4o`) instead.

llm/providers.go
nickmisasi: With Cohere and Mistral now migrated into the registry (3 providers total), the generic pattern has proven its reuse value. `CreateTransport` is only used by Scale but is a nil-optional field, so it doesn't add complexity for the others.

crspeller: I did not mean the whole registry thing I meant the CustomAuthTransport. It feels very specific to Scale. I don't think it makes sense as generic as it is. I think it makes more sense as just a Scale thing.

crspeller: Spoke offline, we will keep this.

PR #518, opened 2026-02-24
Fix MCP OAuth detection for bare 401 responses

nickmisasi:
#### Summary
Ensure MCP OAuth auth prompts still trigger when remote servers return HTTP 401 without a `WWW-Authenticate` header.

#### Ticket Link
NONE

#### Screenshots
NONE

#### Release Note
```release-note
Fixed MCP OAuth initiation for servers that return 401 responses without a WWW-Authenticate header.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced OAuth authentication error handling to ensure proper cleanup of HTTP resources across multiple error scenarios, improving reliability of authentication workflows.

### DIFF
--- a/mcp/client.go
+++ b/mcp/client.go
@@ -268,11 +268,9 @@ func (c *Client) createSession(ctx context.Context, serverConfig ServerConfig) (
 	}
 
 	// Check for OAuth error from Streamable HTTP attempt
-	var streamableAuthErr *mcpUnauthorized
-	streamableAsOAuth := errors.As(errStreamable, &streamableAuthErr)
-	streamableMetadataURL, streamableStringOAuth := extractOAuthMetadataURL(errStreamable)
-	if streamableAsOAuth {
-		authURL, oauthErr := c.oauthManager.InitiateOAuthFlow(ctx, c.userID, c.config.Name, serverConfig.BaseURL, streamableAuthErr.MetadataURL())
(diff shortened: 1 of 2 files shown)


mcp/client.go
crspeller, line 1: All the changes in this file seem pointless.

mcp/user_clients.go
```diff
@@ -66,7 +66,8 @@ func (c *UserClients) ConnectToRemoteServers(servers []ServerConfig) *Errors {
 
 			// Check if this is an OAuth authentication error
 			var oauthErr *OAuthNeededError
-			if errors.As(err, &oauthErr) {
```
crspeller, line 69: Pointless change.

mcp/oauth_transport.go
```diff
@@ -25,6 +26,15 @@ type mcpUnauthorized struct {
 	err         error
 }
 
+func drainAndCloseResponseBody(resp *http.Response) {
```
crspeller, line 29: Do we actually need this?

crspeller (changes requested):

mcp/oauth_transport.go
nickmisasi: It's a real finding. When a 401 happens we leak the connection and it can't be reused

PR #545, opened 2026-03-06
Make mcp sessions stateless and use internal URL for plugin mcp server

BenCookie95:
#### Summary
Without `Stateless: true`, the initialize call would hit node A, which stores the session in its in-memory sessions map. The next request hits node B, sends the same Mcp-Session-Id, the SDK looks it up in node B's empty map, finds nothing, and returns 404. With stateless mode, the lookup is skipped entirely.

Making sessions stateless means that we can't do server-initiated requests (i.e., the server asking the client a question mid-tool-call). That's a feature we don't use — our MCP server is a pure tool server where the client calls tools and gets results back.

See docs [here](https://github.com/modelcontextprotocol/go-sdk/blob/16d990b0416f63ca4948a5d8ba8f54ac6114b5a9/docs/protocol.md?plain=1#L204).

This also means we can't support SSE and can only support the streamable mode. We don't support SSE right now anyway.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67812

#### Release Note

```release-note
Fix intermittent tool failures when using Mattermost MCP on a HA instance
```

## Summary by CodeRabbit

* **New Features**
  * Added a configurable "stateless" mode for plugin-server communication (new flag/config option) to improve reliability in multi-node/HA deployments.

* **Refactor**
  * Centralized derivation of the internal server URL and now passes the resolved URL into plugin communication setup.
  * Added debug logging for the resolved internal URL to aid diagnostics during startup.

### DIFF
--- a/mcpserver/cmd/main.go
+++ b/mcpserver/cmd/main.go
@@ -53,6 +53,7 @@ Authentication is handled via Personal Access Tokens (PAT).`,
 	rootCmd.Flags().IntVar(&httpPort, "http-port", 8080, "Port for HTTP server (used when transport is http)")
 	rootCmd.Flags().StringVar(&httpBindAddr, "http-bind-addr", "127.0.0.1", "Bind address for HTTP server (defaults to localhost for security, use 0.0.0.0 for all interfaces)")
 	rootCmd.Flags().StringVar(&siteURL, "site-url", "", "External URL for OAuth and CORS (required when http-bind-addr is localhost or when using reverse proxy)")
+	rootCmd.Flags().Bool("stateless", false, "Enable stateless mode for multi-node/HA deployments (no server-side session tracking)")
 
 	// AI tracking flag
 	rootCmd.Flags().Bool("track-ai-generated", false, "Track AI-generated content in posts (default: false for stdio, true for http)")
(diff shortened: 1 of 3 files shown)


server/embedded_mcp_server.go
```diff
@@ -121,3 +100,27 @@ func (e *EmbeddedMCPServer) CreateClientTransport(userID, sessionID string, plug
...
+	internalServerURL := "http://localhost:8065"
+	if config := pluginAPI.Configuration.GetConfig(); config != nil {
+		if config.ServiceSettings.ListenAddress != nil && *config.ServiceSettings.ListenAddress != "" {
+			listenAddr := *config.ServiceSettings.ListenAddress
+			switch {
+			case len(listenAddr) > 0 && listenAddr[0] == ':':
+				internalServerURL = "http://localhost" + listenAddr
+			case len(listenAddr) > 7 && listenAddr[:7] == "0.0.0.0":
+				internalServerURL = "http://localhost" + listenAddr[7:]
+			case strings.HasPrefix(listenAddr, "[::]:"):
```
BenCookie95, line 118: We weren't handling ipv6 wildcard addresses before, suggested fix from coderabbitai

BenCookie95:
@coderabbitai full review

PR #552, opened 2026-03-11
MM-67891 Added support for MCP servers with no DCR

avasconcelos114:
#### Summary

Some MCP servers (e.g. GitHub) require a pre-registered OAuth application and don't support Dynamic Client Registration (RFC 7591). Previously, the plugin assumed all servers supported DCR, making it impossible to connect to these providers.

##### Backend:

- Add optional clientID / clientSecret fields to ServerConfig
- When static credentials are provided, skip DCR and use them directly in the OAuth flow
- Persist static credentials through the OAuth session so ProcessCallback can reconstruct the correct config without needing the original server config at callback time
- Fix extractOAuthMetadataURL bug where splitting on bare ":" would break on https:// URLs — now correctly splits on ": Got error:" delimiter
- Parse scopes_supported from protected resource metadata (RFC 9728)
- Add ServerConfig.StaticOAuthCreds() helper to consolidate repeated credential construction

##### Frontend (System Console):

- Add collapsible "OAuth Credentials (Optional)" section to each MCP server config with Client ID and Client Secret (masked) fields
- Section auto-expands when credentials are already configured, shows a "Configured" badge when collapsed

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-67891

#### Screenshots

|  before  |  after  |
|----|----|
| image | image |
| image | image |

#### Release Note

```release-note
Added support to MCP servers that don't support Dynamic Client Registration 
```

## Summary by CodeRabbit

* **New Features**
  * Support for static OAuth credentials (pre-provisioned client ID/secret) including UI to configure them in the MCP server editor.
  * Protected-resource discovery can surface supported scopes.
  * OAuth sessions now persist static credential info and use an expiry for session storage.
  * Key-value store gained expiry-capable set operation.

* **Bug Fixes**
  * Improved extraction of OAuth metadata URLs and clarified related error text.

* **Tests**
  * Added tests for metadata URL extraction, static credential flows, session/callback behavior, and KV expiry handling.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,41 @@ permissions:
   contents: write
 
 jobs:
+  e2e-build-artifact:
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
+      with:
+        node-version: 18
+    - name: Build plugin bundle for e2e jobs
+      run: make dist
+    - uses: actions/upload-artifact@v4
+      with:
+        name: e2e-plugin-dist
+        path: ./dist/*.tar.gz
+        retention-days: 1
+
   e2e:
+    needs: e2e-build-artifact
     timeout-minutes: 60
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        shard:
+          - name: shard-1
+            group: e2e-shard-1
+          - name: shard-2
+            group: e2e-shard-2
+          - name: shard-3
+            group: e2e-shard-3
+          - name: shard-4
+            group: e2e-shard-4
+    name: e2e-${{ matrix.shard.name }}
     env:
-      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:release-11' }}
+      MM_IMAGE: ${{ vars.MM_IMAGE || 'mattermost/mattermost-enterprise-edition:11.5.1' }}
       EXCLUDE_REAL_API_TESTS: 'true'
     defaults:
       run:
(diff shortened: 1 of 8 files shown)


mcp/oauth_manager.go
```diff
@@ -157,22 +179,24 @@ func (m *OAuthManager) InitiateOAuthFlow(ctx context.Context, userID, serverID,
...
 	if err := m.storeSession(&OAuthSession{
 		UserID:            userID,
 		ServerID:          serverID,
 		ServerURL:         serverURL,
 		ServerMetadataURL: metadataURL,
 		CodeVerifier:      codeVerifier,
 		State:             state,
+		StaticClientID:    staticCredsClientID(staticCreds),
+		StaticClientSecret: staticCredsClientSecret(staticCreds),
 		CreatedAt:         time.Now(),
```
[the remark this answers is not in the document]
avasconcelos114: Considering that:
1. The session is ephemeral (created, used once during callback, then deleted)
2. OAuthManager has no access to server configs, so a lookup would require a meaningful refactor
3. The secret is already gated behind having a non-empty ClientID
4. The CodeVerifier (PKCE secret) is already stored in the same session with the same lifetime

I'm not entirely certain this would be in-scope or worthwhile to have it fixed here

mcp/oauth_manager.go
```diff
@@ -157,36 +180,61 @@ func (m *OAuthManager) InitiateOAuthFlow(ctx context.Context, userID, serverID,
...
-		State:             state,
-		CreatedAt:         time.Now(),
+		UserID:             userID,
+		ServerID:           serverID,
+		ServerURL:          serverURL,
+		ServerMetadataURL:  metadataURL,
+		CodeVerifier:       codeVerifier,
+		State:              state,
+		StaticClientID:     staticCredsClientID(staticCreds),
+		StaticClientSecret: staticCredsClientSecret(staticCreds),
```
edgarbellot: Would it be possible to avoid storing the static client secret in the OAuth session? If the user abandons the flow (closes the browser, loses connectivity, etc.), the session persists in the KV store indefinitely - and with it, the secret.

Current flow:

`InitiateOAuthFlow → stores secret in session (KV store) → ... → ProcessCallback → reads secret from session`

Ideally, `ProcessCallback` would re-derive the static credentials from the plugin config using the ServerID already stored in the session, so the secret never hits the KV store:

`InitiateOAuthFlow → stores session without secret → ... → ProcessCallback → looks up secret from plugin config by ServerID`

I realize this isn't a one-liner since `OAuthManager` doesn't currently have access to the plugin config. It would need either a config lookup function injected into `OAuthManager`, or `ClientManager.ProcessOAuthCallback` could resolve the credentials and pass them into ProcessCallback as a parameter. 

If that refactor feels too big for this PR, a lighter mitigation would be switching `KVSet` to `KVSetWithExpiry` with ~10 minutes in storeSession (oauth_store.go:151) - that way even if the secret is stored temporarily, abandoned sessions get cleaned up automatically instead of lingering forever.

avasconcelos114: Ah great point! I reworked the oauth management so that static oauth credentials do not get stored in the KV store (and also added a 10 minute TTL to the oauth session for good measure so we don't have stale oauth connect attempts in the KV store)

edgarbellot: @avasconcelos114 that perfectly addresses it, thanks a mil Andre!

PR #558, opened 2026-03-13
Channel automation additions

BenCookie95:
#### Summary
- Added a new bridge function for other plugins to discover tools that are available to the requesting user. Useful for plugins if they are going to use the new bridge completion functionality. It won't return embedded mm mcp server tools unless a userID is passed.
- Added a new field to the completion bridge endpoints called `allowed_tools`. This allows completion calls to use from tools from a pre-defined list. This list is checked at runtime each time to ensure the requesting user actually permission to use those tools. You can service account tools if you don't pass a requesting userID
- Added MCP tools to interact with the new automation plugin. You can create, modify, delete and list automations. These tools are only added to context in channels where you are a channel admin or if you are in a DM with Matty. They are filtered out for other users in order to not pollute context.
- Added a new ListAgents tool so that Matty can see what agents are available on the server when we are creating automations. These Agent IDs are needed at automation creation time.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67712

#### Release Note

```release-note
Allow tool execution through bridge and add mcp tools for managing automations
```

## Release Notes

* **New Features**
  * Added tool allowlist support for agent completions—control which tools agents can execute
  * Added endpoint to discover eligible tools for a given agent
  * Added agent discovery and listing capabilities
  * Added channel automation management—create, list, update, and delete automations

* **Bug Fixes**
  * Improved input validation and error handling across bridge endpoints
  * Enhanced tool eligibility filtering based on user permissions

* **Documentation**
  * Updated bridge client documentation with tool allowlist usage and agent tools discovery


api/api_llm_bridge.go
```diff
@@ -131,6 +146,262 @@ func (a *API) buildLLMBridgeContext(bot *bots.Bot, req bridgeclient.CompletionRe
...
+	// Native web search is a provider-level feature (not an MCP tool),
+	// so it's not part of allowed_tools — it's always available when configured.
+	if bot.HasNativeWebSearchEnabled() {
+		opts = append(opts, llm.WithNativeWebSearchAllowed())
+	}
+
+	return bot, llmRequest, opts, 0, nil
+}
+
+func (a *API) discoverBridgeEligibleTools(ctx context.Context, userID string) ([]bridgeclient.BridgeToolInfo, error) {
```
crspeller: All of this tool restriction functionality I do not think belongs on the bridge here. This feels like it's something similar to what https://github.com/mattermost/mattermost-plugin-agents/pull/520 is accomplishing.

llmcontext/llm_context.go
```diff
@@ -176,6 +177,12 @@ func (b *Builder) getToolsStoreForUser(c *llm.Context, bot *bots.Bot, userID str
 		// Get tools from all connected servers
 		mcpTools, mcpErrors := b.mcpToolProvider.GetToolsForUser(userID)
 
+		// Filter out automation tools for users without channel admin (or sysadmin) permission
+		showAutomation := shouldShowAutomationTools(b.pluginAPI, userID, c.Channel)
```
crspeller: This doesn't feel like it belongs here. Shouldn't these be permission filtered out on an MCP server level?

mcpserver/tools/automations.go
```diff
@@ -0,0 +1,627 @@
...
+	"github.com/mattermost/mattermost-plugin-ai/llm"
+	"github.com/mattermost/mattermost/server/public/model"
+)
+
+const automationPluginAPIPath = "/plugins/com.mattermost.channel-automation/api/v1"
+
+// isAutomationPluginInstalled probes the channel automation plugin API to check if
+// the plugin is installed and reachable. Returns true if the plugin responds (even
+// with an auth error), false if it 404s or is unreachable.
+func (p *MattermostToolProvider) isAutomationPluginInstalled() bool {
```
crspeller, line 25: Ideally we implement a way for plugins to register their own MCP tools instead of having these here.

public/bridgeclient/client.go
```diff
@@ -41,6 +41,9 @@ type CompletionRequest struct {
 	Posts              []Post                 `json:"posts"`
 	MaxGeneratedTokens int                    `json:"max_generated_tokens,omitempty"`
 	JSONOutputFormat   map[string]interface{} `json:"json_output_format,omitempty"`
+	// AllowedTools is an optional allowlist of tool names for agent completions.
+	// When provided on agent endpoints, only these eligible tools may run without approval.
+	AllowedTools []string `json:"allowed_tools,omitempty"`
```
crspeller, line 54: This feels like the wrong way to go about this. The agent configuration should be enforcing this not the completion request.

api/api_llm_bridge.go
```diff
@@ -131,6 +146,262 @@ func (a *API) buildLLMBridgeContext(bot *bots.Bot, req bridgeclient.CompletionRe
...
+	if req.UserID != "" {
+		if err := bridgeclient.ValidateID(req.UserID); err != nil {
+			return nil, llm.CompletionRequest{}, nil, http.StatusBadRequest, fmt.Errorf("invalid user_id: %w", err)
+		}
+	}
+	if req.ChannelID != "" {
+		if err := bridgeclient.ValidateID(req.ChannelID); err != nil {
+			return nil, llm.CompletionRequest{}, nil, http.StatusBadRequest, fmt.Errorf("invalid channel_id: %w", err)
+		}
+	}
```
crspeller, line 251: This seems to happen everywhere. Some validation middleware?

api/api_llm_bridge.go
```diff
@@ -131,6 +146,262 @@ func (a *API) buildLLMBridgeContext(bot *bots.Bot, req bridgeclient.CompletionRe
...
+	oauthManager := a.mcpClientManager.GetOAuthManager()
+	httpClient := a.mcpClientManager.GetHTTPClient()
+	toolsCache := a.mcpClientManager.GetToolsCache()
+	if httpClient == nil {
+		return nil, errors.New("MCP HTTP client is unavailable")
+	}
+
+	toolMap := map[string]bridgeclient.BridgeToolInfo{}
+
+	// Discover tools from embedded server if enabled
```
crspeller: Tool discovery certainly shouldn't be a bridge concern.

api/api_llm_bridge.go
```diff
@@ -131,6 +146,262 @@ func (a *API) buildLLMBridgeContext(bot *bots.Bot, req bridgeclient.CompletionRe
...
+
+	eligibleDiscoveredTools, err := a.discoverBridgeEligibleTools(ctx, userID)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	eligibleTools := make([]bridgeclient.BridgeToolInfo, 0, len(eligibleDiscoveredTools))
+	eligibleToolMap := make(map[string]llm.Tool, len(eligibleDiscoveredTools))
+	for _, discovered := range eligibleDiscoveredTools {
+		tool := toolStore.GetTool(discovered.Name)
```
crspeller: We do a lot of work in `discoverBridgeEligibleTools` just to throw it all away and just use the name here.

api/api_llm_bridge.go
```diff
@@ -131,6 +146,262 @@ func (a *API) buildLLMBridgeContext(bot *bots.Bot, req bridgeclient.CompletionRe
...
+
+	return result, nil
+}
+
+func (a *API) filterEligibleToolsForContext(ctx context.Context, userID string, toolStore *llm.ToolStore) ([]bridgeclient.BridgeToolInfo, map[string]llm.Tool, error) {
+	if toolStore == nil {
+		return nil, map[string]llm.Tool{}, nil
+	}
+
+	eligibleDiscoveredTools, err := a.discoverBridgeEligibleTools(ctx, userID)
```
crspeller: What exactly is the purpose of this discovery? Why not just use the tool store? Doesn't it already do this? 

Actually maybe I am confused, but what is this function even doing?

crspeller (changes requested):

llmcontext/llm_context.go
BenCookie95: I could add a layer to the mcp server to filter stuff like this but it would mean passing down channel and team info through the embedded context. We already have this stuff available right here so I took the easier route. I think we can remove this filtering once we add lazy loading of tools? Which would make this addition temporary.

api/api_llm_bridge.go
BenCookie95: Moved some validation into middleware calls and request body ones into shared helpers

api/api_llm_bridge.go
BenCookie95: Removed, tools already filtered and discovered by the time it reaches this point.

mcpserver/tools/automations.go
BenCookie95: This will have to come later after the initial release. The goal is to remove this as soon as we have plugin registered tools

mcpserver/tools/automations.go
```diff
@@ -0,0 +1,656 @@
...
+	reqURL := p.automationAPIURL("/flows")
+	req, err := http.NewRequest(http.MethodGet, reqURL, nil)
```
BenCookie95, line 27: We were using some client4 doRequestWithHeaders function before but that changed in the latest release where it automatically prepends `<serverURL>/api/v4` so it no longer works for us with plugins

mcpserver/tools/provider.go
```diff
@@ -105,6 +119,69 @@ func (p *MattermostToolProvider) ProvideTools(mcpServer *mcp.Server) {
 	for _, mcpTool := range mcpTools {
 		p.registerDynamicTool(mcpServer, mcpTool)
 	}
+
+	// Add middleware to dynamically filter automation tools from tools/list
+	// when the channel automation plugin is not installed.
+	mcpServer.AddReceivingMiddleware(p.automationToolFilterMiddleware())
```
BenCookie95, line 128: Moved into a middleware to check for plugin existence when an agent requests a tool list. Previously if the automation plugin was uploaded after the agents plugin started up, we would never find the automation tools. We are caching the 'plugin installed` flow for 30 mins so we don't make redundant requests

llmcontext/llm_context.go
crspeller: I think we still want to filter out tools that you don't even have permission to use, even from search. Let's go ahead and pass whatever we need to the MCP server to make this determination. This also means that we can make this determination for external MCP users as well as this solution currently does not cover those.

mcpserver/tools/agents.go
```diff
@@ -0,0 +1,113 @@
...
+type ListAgentsArgs struct{}
+
+// getAgentTools returns agent discovery tools.
+func (p *MattermostToolProvider) getAgentTools() []MCPTool {
+	return []MCPTool{
+		{
+			Name: "list_agents",
+			Description: `List all available AI agents (bots). Returns each agent's ID, display name, and username.
+Use this tool to discover valid provider_id values for the ai_prompt action type when creating automations.
+The agent ID (26-character Mattermost user ID) is what you pass as config.provider_id with provider_type "agent".`,
```
crspeller: Not sure this last line makes sense. Is it specific to some automatons tool? If so it should be with that.

mcpserver/tools/agents.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	if len(bots) == 0 {
+		return "No agents are currently configured.", nil
+	}
+
+	var result strings.Builder
+	result.WriteString(fmt.Sprintf("Found %d agent(s):\n\n", len(bots)))
+
+	for i, a := range bots {
```
crspeller: All formatting should now be done in the format package. Can you move this there? 
The idea is to reference standardization so we always display entities in the same way.

mcpserver/tools/provider.go
```diff
@@ -105,6 +119,69 @@ func (p *MattermostToolProvider) ProvideTools(mcpServer *mcp.Server) {
...
+}
+
+// isAutomationPluginAvailable checks if the automation plugin is installed,
+// caching positive results for automationCacheTTL to avoid repeated probes.
+// Negative results are never cached so newly installed plugins are detected immediately.
+func (p *MattermostToolProvider) isAutomationPluginAvailable() bool {
```
crspeller: Seems over-complicated specifically since the common case of the plugin not being installed is not cached. In the ultimate design where we fetch the tools from the plugin we are going to be making a request anyway. Why the complexity here?

crspeller (changes requested):

llmcontext/llm_context.go
BenCookie95: @crspeller I've moved these checks into the mcpserver but it came with a bit of refactoring. You'll notice we are now passing the currentChannel around a bit more because the mcpserver now requires it when creating a connection so that we can determine if we need to show the automation tools or not. If there is no channel passed during connection creation, we auto include the automation tools (this is for external mcp clients). 

We now include a channelID in the mcp client memory for embedded server mcp sessions. If you change channel then we create a new embedded session so that we can re-evaluate channel permissions.

BenCookie95:
@crspeller I also had to fix an issue where we need to include a server name when choosing a tools to include in `allowed_tools` for the bridge requests.

crspeller (approved):

crspeller:
@BenCookie95 I think some merge conflicts are causing CI to fail. Need to fix those before merge.

BenCookie95:
@crspeller Just re-requesting review due to a few minor changes here: https://github.com/mattermost/mattermost-plugin-agents/pull/558/commits/880cf89e5477266b2744da7b53ba2bd84467dd4d.

Like we discussed on Tuesday, the bridge derives the server origin itself so the caller doesn't need to know it. Also, moved automation creation instructions over to the automation plugin.

This still needs to finish security review as well

api/api_llm_bridge.go
```diff
@@ -451,17 +487,58 @@ func (a *API) streamLLMResponse(c *gin.Context, bot *bots.Bot, llmRequest llm.Co
 }
 
 // handleNonStreamingLLMResponse handles non-streaming LLM responses.
-func (a *API) handleNonStreamingLLMResponse(c *gin.Context, bot *bots.Bot, llmRequest llm.CompletionRequest, opts ...llm.LanguageModelOption) {
-	response, err := bot.LLM().ChatCompletionNoStream(llmRequest, opts...)
+// When shouldExecute is non-nil, the call is routed through a toolrunner so
+// allowlisted tool calls are auto-executed; the runner's text stream is
+// drained into a single concatenated string before responding, mirroring
+// what ChatCompletionNoStream would have produced.
+func (a *API) handleNonStreamingLLMResponse(c *gin.Context, bot *bots.Bot, llmRequest llm.CompletionRequest, shouldExecute func(llm.ToolCall) bool, opts ...llm.LanguageModelOption) {
```
BenCookie95, line 547: REason for changes here: There is an API mismatch between toolrunner and the non-streaming HTTP endpoint:

The non-streaming endpoint must return one JSON body. It can't return until it has the final text.toolrunner.Run only exposes a streaming API. There is no RunNoStream. It returns a *ToolRunResult whose Stream field is the live event channel, with the final-round text events emitted through the same channel as everything else.
So to satisfy the non-streaming HTTP contract, the handler has to consume the whole runner stream synchronously and rebuild the string itself.

hmohammed-prodsec (approved):

crspeller (approved):

PR #589, opened 2026-04-02
MM-65671: Agents CRUD, legacy bot migration, and bot account sync

nickmisasi:
#### Summary

Implements user agent CRUD improvements, migration of legacy config-defined bots into user agents, and synchronization of Mattermost bot accounts when agents are updated or deleted (display name patch, bot deactivation). Updates the agents UI, system console wiring, i18n strings, and adds e2e coverage plus a small permissions helper.

image
image
image
image
image
image
image
image
image

Tools list on RHS is based on allowed tools for selected agent
image
Vs an agent with no tools enabled
image

#### Ticket Link

Jira https://mattermost.atlassian.net/browse/MM-65671

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Full Agents product: REST APIs and UI to create/edit/delete agents (avatar upload), access controls, service & model selection, per-agent MCP tool controls, delegated admins, license gating, Agents product page and nav integration, and RHS/tool-picker respect per-agent allowlists.

* **Migrations**
  * One-time migration of legacy bots into Agents and new Agents DB table + persistence APIs.

* **Tests**
  * Expanded unit, integration, and Playwright E2E coverage for agents, access control, MCP tools, provider/config flows, and migrations.


api/api_agents.go
```diff
@@ -0,0 +1,624 @@
...
+		displayNameChanged = agent.DisplayName != *req.DisplayName
+		agent.DisplayName = *req.DisplayName
+	}
+	if req.Username != nil {
+		if !validUsernameRe.MatchString(*req.Username) {
+			c.AbortWithError(http.StatusBadRequest, errors.New("invalid username: must start with a lowercase letter and contain only lowercase letters, numbers, dots, hyphens, or underscores"))
+			return
+		}
+		agent.Username = *req.Username
+	}
```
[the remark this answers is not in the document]
nickmisasi: I've removed username changes for now. I think this will be a bit of a nightmare to manage.

api/api_agents.go
```diff
@@ -0,0 +1,624 @@
...
+	if req.ServiceID != nil {
+		agent.ServiceID = *req.ServiceID
```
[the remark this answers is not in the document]
nickmisasi: Updates now run the same service lookup as create against configured services. If service_id isn’t found, we return 4xx and do not persist it, so invalid IDs can’t be saved with 200.

api/api_agents.go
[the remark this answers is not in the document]
nickmisasi: **Resolved in code:** Username changes after creation are rejected in `handleUpdateAgent` (see `UpdateAgentRequest` / username handling). This matches the direction you noted—no separate bot rotation path needed while updates are disallowed.

api/api_agents.go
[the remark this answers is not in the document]
nickmisasi: **Resolved in code:** `handleUpdateAgent` validates `service_id` against configured services before persisting (same lookup pattern as create). Unknown services return 4xx and are not saved.

api/api_agents.go
```diff
@@ -0,0 +1,624 @@
...
+	c.JSON(http.StatusOK, services)
+}
+
+// FetchModelsForServiceRequest is the JSON body for POST /agents/models/fetch.
+type FetchModelsForServiceRequest struct {
+	ServiceID string `json:"service_id" binding:"required"`
+}
+
+// handleFetchModelsForService lists models for a configured service using stored credentials (non-admin).
+func (a *API) handleFetchModelsForService(c *gin.Context) {
```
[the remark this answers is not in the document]
nickmisasi: **Resolved in code:** Both `GET /services` and `POST /agents/models/fetch` are gated with `canConfigureAgentServices` (requires `PermissionManageOwnAgent` or `PermissionManageOthersAgent`), not only auth + license. See `handleListServices` and `handleFetchModelsForService` in `api/api_agents.go`.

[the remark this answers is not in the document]
nickmisasi: **Resolved in code:** Same as the sibling thread on this file — `canConfigureAgentServices` guards both `GET /services` and `POST /agents/models/fetch`.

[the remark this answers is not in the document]
nickmisasi: **Resolved in code:** Same gating — `canConfigureAgentServices` on both listing services and fetching models; outbound model-list calls are not available to arbitrary licensed users without agent-management permissions.

useragents/model.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// EnabledToolsJSON returns the JSON-encoded string for the EnabledTools slice.
+func (u *UserAgent) EnabledToolsJSON() string {
+	b, err := json.Marshal(u.EnabledTools)
+	if err != nil {
+		return "[]"
+	}
+	return string(b)
+}
```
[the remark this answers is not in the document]
nickmisasi: **Closing as by design:** `EnabledToolsJSON` intentionally preserves nil vs empty semantics (documented on the method): `nil` → `"null"` (no filter / all tools) and empty slice → `"[]"` (explicitly no tools). We are not routing this through `mustMarshalSlice`, which would collapse nil to `"[]"` and change behavior.

e2e/helpers/agent-page.ts
```diff
@@ -0,0 +1,216 @@
...
+    async clickEditAction(displayName: string): Promise<void> {
+        const rowScope = this.page.getByText(displayName, { exact: true }).locator(
+            'xpath=ancestor::div[.//button[@aria-label="Agent actions"]][1]',
+        );
+        await rowScope.getByRole('button', { name: 'Edit' }).click();
+    }
+
+    async clickDeleteAction(): Promise<void> {
+        await this.page.getByText('Delete', { exact: true }).click();
+    }
```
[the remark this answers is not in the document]
nickmisasi: **Resolved in code:** `clickEditAction` already mirrors `openAgentActions`: same row scope from `displayName`, then `getByRole('button', { name: 'Edit' })` within that row (`e2e/helpers/agent-page.ts`). No additional XPath divergence from the actions menu pattern.

server/main.go
```diff
@@ -476,13 +496,14 @@ func (p *Plugin) OnActivate() error {
 		mcpHandlers,
 		llmUpstreamHTTPClient,
 		p.store,
+		p.store,
 		&p.configuration,
 		p,
+		p,
```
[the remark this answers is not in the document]
nickmisasi: **Not a bug:** `api.New` takes distinct interface-typed parameters. `*store.Store` implements both `ConfigStore` and `AgentStore`, so `p.store` is passed twice for two different parameter slots. Likewise `&p.configuration` is the `Config` value vs `ConfigUpdater`, and `p` implements both `ClusterNotifier` and `ClusterAgentNotifier`. The call matches `api.New` in `api/api.go`.

bots/agents_test.go
```diff
@@ -0,0 +1,187 @@
...
+			{ServerOrigin: "https://other.com", ToolName: "tool_b"},
+		}
+		result := convertEnabledTools(input)
+		require.Len(t, result, 2)
+		assert.Equal(t, "https://server.com", result[0].ServerOrigin)
+		assert.Equal(t, "tool_a", result[0].ToolName)
+		assert.Equal(t, "https://other.com", result[1].ServerOrigin)
+		assert.Equal(t, "tool_b", result[1].ToolName)
+	})
+}
```
crspeller: **[Test Quality] Tests Go's for-range loop and struct assignment, not our code.**

`convertEnabledTools` copies `ServerOrigin` and `ToolName` from one struct to a structurally identical struct in a loop. The "populated input converts correctly" subtest just verifies that Go's loop copied the fields — this can only fail if the compiler is broken.

The nil-vs-empty subtest has slight documentary value for the nil/empty contract, but the "populated input" subtest should be removed. Consider validating nil-semantics in the store round-trip test where it actually matters.

Also note: `TestUserAgentToBotConfigEnabledTools` and `TestUserAgentToBotConfigNilEnabledTools` (lines 140-170) re-test this same struct-copy through the parent function, making three tests for the same line of code.

webapp/src/components/agents/tabs/config_tab.tsx
```diff
@@ -0,0 +1,405 @@
...
+            tokenLimit: 0,
+            streamingTimeoutSeconds: 0,
+            sendUserId: false,
+            outputTokenLimit: selectedService.output_token_limit || 4096,
+            useResponsesAPI: selectedService.use_responses_api,
+            region: '',
+            awsAccessKeyID: '',
+            awsSecretAccessKey: '',
+        };
+    }, [selectedService]);
```
crspeller: **[Complexity] Fabricates a full `LLMService` with 12 dummy fields just to satisfy `ReasoningConfigItem`'s prop type.**

Fields like `apiKey: ''`, `awsAccessKeyID: ''`, `awsSecretAccessKey: ''` are populated here solely to match the System Console service type. `ReasoningConfigItem` presumably only needs `type`, `outputTokenLimit`, and `useResponsesAPI`.

Same applies to `reasoningBot` below (lines 114-133) which constructs another shim object. Consider refactoring `ReasoningConfigItem` to accept a slimmer interface with just the fields it actually uses.

bots/agents_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// Bot-level fields map from UserAgent when set
+	assert.Equal(t, "", cfg.Model)
+	assert.False(t, cfg.DisableTools)
+	assert.False(t, cfg.EnableVision)
+	assert.Nil(t, cfg.EnabledNativeTools)
+	assert.False(t, cfg.ReasoningEnabled)
+	assert.Equal(t, "", cfg.ReasoningEffort)
+	assert.Equal(t, 0, cfg.ThinkingBudget)
+	assert.False(t, cfg.StructuredOutputEnabled)
+}
```
crspeller: This doesn't test anything other than go works. 
Some of these other tests are the same way. We can probably eliminate most of these when we do the struct simplification.

useragents/model.go
```diff
@@ -0,0 +1,160 @@
...
+	DisableTools            bool          `json:"disable_tools"`
+	EnabledNativeTools      []string      `json:"enabled_native_tools"`
+	ReasoningEnabled        bool          `json:"reasoning_enabled"`
+	ReasoningEffort         string        `json:"reasoning_effort"`
+	ThinkingBudget          int           `json:"thinking_budget"`
+	StructuredOutputEnabled bool          `json:"structured_output_enabled"`
+	CreateAt                int64         `json:"create_at" db:"CreateAt"`
+	UpdateAt                int64         `json:"update_at" db:"UpdateAt"`
+	DeleteAt                int64         `json:"delete_at" db:"DeleteAt"`
+}
```
crspeller: Maybe we can eliminate `UserAgent` and use the existing structs instead? We can do some renaming to make things more clear. This package is causing a lot of duplication I think. 

Some help from Claude:

`UserAgent` shares ~18 fields with `llm.BotConfig`, and at runtime it's immediately converted to `BotConfig` via `userAgentToBotConfig()` before any consumer sees it. The only fields `UserAgent` adds are ownership/lifecycle metadata: `BotUserID`, `CreatorID`, `AdminUserIDs`, `CreateAt`, `UpdateAt`, `DeleteAt`.

These 6 fields could be added directly to `BotConfig` (with `omitempty` — they'd be zero-valued for config-defined bots). This would eliminate:

- This entire `useragents` package
- The duplicate `EnabledTool` / `EnabledMCPTool` types (the "import cycle" comment on `EnabledMCPTool` is incorrect — both packages are leaves with no internal imports, verified)
- `userAgentToBotConfig()` and `convertEnabledTools()` in `bots/bots.go`
- `mcpToolsToEnabled()` in `legacy_bot_migration.go`
- The duplicated `canUserAccessAgent` logic in `api_agents.go` (could reuse `CheckUsageRestrictionsForUser` from `bots/permissions.go` since the types would match)
- The risk of field drift between the two types when new config fields are added

The store layer already uses its own `agentRow` struct for DB scanning, so the DB schema doesn't need to change. The API request/response types (`CreateAgentRequest`, `UpdateAgentRequest`) are separate DTOs and stay as-is.

api/api_agents.go
```diff
@@ -0,0 +1,677 @@
...
+	EnabledTools            *[]useragents.EnabledTool `json:"enabled_tools"`
+	Model                   *string                   `json:"model"`
+	EnableVision            *bool                     `json:"enable_vision"`
+	DisableTools            *bool                     `json:"disable_tools"`
+	EnabledNativeTools      *[]string                 `json:"enabled_native_tools"`
+	ReasoningEnabled        *bool                     `json:"reasoning_enabled"`
+	ReasoningEffort         *string                   `json:"reasoning_effort"`
+	ThinkingBudget          *int                      `json:"thinking_budget"`
+	StructuredOutputEnabled *bool                     `json:"structured_output_enabled"`
+}
```
crspeller, line 135: We never use the functionality for partial updates we have here. Do we need the extra complexity? Maybe just use CreateAgentRequest? (with a new name)

api/api_agents.go
```diff
@@ -0,0 +1,677 @@
...
+			if svc.ID == *req.ServiceID {
+				found = true
+				break
+			}
+		}
+		if !found {
+			c.AbortWithError(http.StatusBadRequest, fmt.Errorf("service %q not found in configuration", *req.ServiceID))
+			return
+		}
+		agent.ServiceID = *req.ServiceID
```
crspeller: This is duplicated in create agent.

api/api_agents.go
```diff
@@ -0,0 +1,677 @@
...
+	}
+	if req.StructuredOutputEnabled != nil {
+		agent.StructuredOutputEnabled = *req.StructuredOutputEnabled
+	}
+	if req.ReasoningEffort != "" {
+		agent.ReasoningEffort = req.ReasoningEffort
+	}
+	if len(req.EnabledNativeTools) > 0 {
+		agent.EnabledNativeTools = req.EnabledNativeTools
+	}
```
crspeller: A lot of logic here to set defaults and stuff. Adds extra complexity for not a lot of gain. Maybe we can just handle it like we did before where the system console deals with setting defaults. Changed defaults here: #617

server/cluster_events.go
```diff
@@ -9,38 +9,48 @@ import (
...
-	cfg, err := p.store.GetConfig()
-	if err != nil {
-		p.pluginAPI.Log.Error("Failed to reload config from database on cluster event", "error", err.Error())
-		return
-	}
-	if cfg != nil {
-		p.configuration.Update(cfg)
+	case clusterEventAgentUpdate:
+		// Force the next EnsureBots to re-read DB agents, then trigger it.
+		p.bots.ForceRefreshOnNextEnsure()
```
crspeller, line 53: Seems like an odd way to do this. Why not just delete the cache and let it sort itself out naturally?

llm/configuration.go
```diff
@@ -48,6 +48,13 @@ const (
 	UserAccessLevelNone
 )
 
+// EnabledMCPTool identifies a single MCP tool on a specific server.
+// Mirrors useragents.EnabledTool but lives in the llm package to avoid import cycles.
```
crspeller: Not a problem when that package no longer exists.

server/legacy_bot_migration.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	mmByUsername := make(map[string]string)
+	for _, b := range previousMMBots {
+		if b.DeleteAt == 0 {
+			mmByUsername[b.Username] = b.UserId
+		}
+	}
+
+	// If any legacy config bot still needs migration but has no Mattermost bot row,
+	// defer the entire migration: do not create partial agents, wipe config, or set the flag.
```
crspeller, line 69: Not sure we should fail hard here. What are we going to do about this? When would this actually happen?

webapp/src/components/agents/agent_config_modal.tsx
```diff
@@ -0,0 +1,472 @@
...
+            setActiveTab('config');
+            return;
+        }
+        setErrors({});
+        setSaving(true);
+
+        try {
+            let savedAgent: UserAgent;
+
+            if (mode === 'create') {
```
crspeller, line 242: Yikes. I think we can get rid of this.

webapp/src/components/agents/delete_agent_dialog.tsx
```diff
@@ -0,0 +1,213 @@
...
+import {PrimaryButton} from '@/components/assets/buttons';
+
+type Props = {
+    agentName: string;
+    confirmPending?: boolean;
+    onConfirm: () => void;
+    onCancel: () => void;
+}
+
+const DeleteAgentDialog = (props: Props) => {
```
crspeller, line 16: We should really have some standard components we can actually use. Lots of duplication from my modal for the custom prompts.

crspeller (changes requested):
Looks awesome!

The biggest issue is the whole `useragents` package which seems entirely unnecessary duplication. 

One other issue I seen when testing is that newly created agents do not appear for the user without a refresh. We should make sure that newly created agents are available immediately. 

One suggestion is that maybe we should have a manage agents option in the agents drop-down like the one for custom prompts.

useragents/model.go
nickmisasi: I've refactored this to use BotConfig

llm/configuration.go
nickmisasi: Addressed as part of the same refactor: `EnabledMCPTool` now lives only in `llm`, so the duplicate type/import-cycle workaround is gone on the current branch.

server/legacy_bot_migration.go
nickmisasi: Updated this path earlier today so a missing Mattermost bot row no longer fails the migration. We now log and soft-defer the migration so activation can retry later instead of partially migrating or erroring.

useragents/model.go
nickmisasi: Addressed earlier today: the separate `useragents` package is gone in the current branch state, and persisted agents now use `llm.BotConfig` directly end-to-end. That removed the conversion helpers and field-drift risk from having two parallel models.

webapp/src/components/agents/delete_agent_dialog.tsx
nickmisasi: Would be even better if they came from core product :P

api/api_agents.go
nickmisasi: I've refactored this to leverage the front-end as the defaults for now. My intention behind this originally was to allow admins in the future to configure what they want the defaults to be, but for now we can just go with simplification.

api/api_agents.go
nickmisasi: Updated to be a PUT instead of PATCH

server/cluster_events.go
nickmisasi: ForceRefreshOnNextEnsure() doesn't actually clear the bot registry. It only nils the fast-path snapshot
  (lastEnsuredBotCfgs/lastEnsuredServiceCfgs) and flips forceRefresh = true. The in-memory b.bots slice stays untouched until something calls EnsureBots(). EnsureBots() is only called from three places: plugin activation, the config-update listener, and refreshBotsAndNotify, and here. 

If we went lazy with it, it would cause weirdness with an agent created on node A but talked to on node B

webapp/src/components/agents/agent_config_modal.tsx
nickmisasi: Cleaned up because of the simplification around PUT vs PATCH and front-end source of truth.

api/api_agents.go
```diff
@@ -0,0 +1,685 @@
...
+// substitutes hidden defaults for omitted fields. All fields below are taken verbatim
+// from the request payload.
+//
+// `enabledMCPTools` is a required tri-state field:
+//   - `null` means "allow all MCP tools"
+//   - `[]`   means "allow no MCP tools"
+//   - `[..]` means "allow only the listed tools"
+//
+// Omitting `enabledMCPTools` from the JSON is rejected as an invalid request.
+type CreateAgentRequest struct {
```
crspeller, line 32: The `enabledMCPToolsProvided` is a fun trick, but this seems over-complicated for a simple CRUD endpoint. How about we just have a separate bool instead of relying on nil vs empty state? Really this is handling the state where the user has every MCP server on in the UI and just wants every new server added.

api/api_agents.go
```diff
@@ -0,0 +1,685 @@
...
+// to keep must be sent on every save. Field names mirror CreateAgentRequest so clients
+// can send the same document shape for both flows.
+//
+// `enabledMCPTools` is a required tri-state field with the same semantics as
+// CreateAgentRequest: `null` = all, `[]` = none, `[..]` = allowlist. Omitting it is
+// rejected as an invalid request.
+//
+// `username` is still validated against the stored agent's name and cannot be changed
+// after creation; see handleUpdateAgent.
+type UpdateAgentRequest struct {
```
crspeller, line 57: Not sure we really need a whole other struct here. Maybe simplify?

api/api_agents.go
```diff
@@ -0,0 +1,685 @@
...
+	}
+
+	_, enabledMCPToolsProvided := raw["enabledMCPTools"]
+	*r = CreateAgentRequest(decoded)
+	r.enabledMCPToolsProvided = enabledMCPToolsProvided
+	return nil
+}
+
+// UpdateAgentRequest is the JSON body for PUT /agents/:agentid.
+// This endpoint is a full-object replacement, not a patch: every field the caller wants
```
crspeller: Lots of LLM excessive comments in here or just talking about changes made. Comments should be addressed at someone in the future who doesn't care about our iterations here. (I think a harness change recently has caused this to be a bigger problem)

llm/configuration.go
```diff
@@ -97,6 +111,15 @@ type BotConfig struct {
 	// to constrain the model's output to valid JSON matching the schema.
 	// Only applicable to Anthropic (Claude 4.5/4.6+ models)
 	StructuredOutputEnabled bool `json:"structuredOutputEnabled"`
+
+	// Admin / lifecycle metadata. These are populated only for DB-backed user agents;
+	// they stay zero for config-defined bots so JSON omits them (omitempty).
```
crspeller: The only ones that exist now?

llm/configuration.go
```diff
@@ -48,6 +50,12 @@ const (
 	UserAccessLevelNone
 )
 
+// EnabledMCPTool identifies a single MCP tool on a specific server (config bots and persisted agents).
+type EnabledMCPTool struct {
+	ServerOrigin string `json:"server_origin"`
+	ToolName     string `json:"tool_name"`
+}
+
 type BotConfig struct {
```
crspeller, line 59: Seperate PR, but we should rename this now.

crspeller (changes requested):

api/api_agents.go
nickmisasi: This was a backwards compatibility thing - there were 3 states - some tools, no tools, null tools, the third state was what I used to represent the existing config-migrated bots since they pre date the concept of agents being able to only use certain tools. 

I've refactored to a boolean 
image
that will apply to them automatically

api/api_agents.go
nickmisasi: Cleaned up

llm/configuration.go
nickmisasi: This was in reference to https://github.com/mattermost/mattermost-plugin-agents/pull/589#discussion_r3111850804

But cleaned up now.

api/api_agents.go
nickmisasi: I think a distinct `UpdateAgentRequest` while it does have overlap is more declarative than a single request for Create/Update that has specific behaviour depending on the action type

api/api_agents.go
```diff
@@ -0,0 +1,610 @@
...
+		if errors.As(err, &appErr) && appErr.Id == "app.user.save.username_exists.app_error" {
+			c.AbortWithError(http.StatusConflict, fmt.Errorf("username %q is already taken", req.Username))
+			return
+		}
+		c.AbortWithError(http.StatusInternalServerError, fmt.Errorf("failed to create bot account: %w", err))
+		return
+	}
+
+	agent := buildAgentConfigForCreate(req, userID, mmBot.UserId)
+
```
edgarbellot, line 345: Nit: would it make sense to call `IsValid()` on the built config before persisting?

Something like this (and the equivalent spot for update):

```
if !agent.IsValid() {
    c.AbortWithError(http.StatusBadRequest, errors.New("invalid agent configuration"))
    return
}
```

api/api.go
```diff
@@ -212,6 +233,20 @@ func (a *API) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Reques
 	router.PUT("/mcp/user-preferences", a.handlePutUserPreferences)
 	router.DELETE("/mcp/oauth/:serverName", a.handleDeleteUserMCPOAuth)
 
+	// Agent CRUD routes — authenticated, license-gated
+	agentRouter := router.Group("/agents")
```
edgarbellot, line 241: Not a security concern, just an operational thought: [agentLicenseRequired](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api.go#L238) (238) gates the entire agent router, including [GET](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api.go#L240) (240) and [DELETE](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api.go#L244) (244). Since DB agents [intentionally keep running after a license downgrade](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/bots/bots.go#L261-L263) (261-263), would it make sense to exempt GET and DELETE from the license gate so admins can still list and clean up existing agents if the license expires?

api/api_agents.go
```diff
@@ -0,0 +1,610 @@
...
+// buildAgentConfigForCreate builds a new llm.BotConfig from req and the new bot/user IDs.
+func buildAgentConfigForCreate(req CreateAgentRequest, userID, botUserID string) *llm.BotConfig {
+	return &llm.BotConfig{
+		BotUserID:               botUserID,
+		CreatorID:               userID,
+		DisplayName:             req.DisplayName,
+		Name:                    req.Username,
+		ServiceID:               req.ServiceID,
+		CustomInstructions:      req.CustomInstructions,
```
edgarbellot, line 218: Can we add a max length check for `customInstructions` before persisting?

Right now the field goes from the JSON body into the DB ([TEXT column, no limit](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/store/migrations/000005_create_user_agents_table.up.sql#L8)) and then into the LLM system prompt on [every conversation](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/prompts/standard_personality_without_locale.tmpl#L43-L45) (43-45), with no length validation in [handleCreateAgent](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api_agents.go#L267) (267) or [handleUpdateAgent](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api_agents.go#L363) (363). Other endpoints in the plugin already guard against this - e.g. [handlePutUserPreferences](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api_mcp.go#L201) (201) uses `MaxBytesReader`.

api/api_agents.go
```diff
@@ -0,0 +1,610 @@
...
+		}
+	}
+	if a.mmClient != nil {
+		// PublishWebSocketEvent requires a non-nil broadcast (server dereferences it).
+		a.mmClient.PublishWebSocketEvent(WebsocketEventBotsInvalidate, map[string]interface{}{}, &model.WebsocketBroadcast{})
+	}
+}
+
+// handleCreateAgent handles POST /agents: creates the bot user and persisted agent config.
+func (a *API) handleCreateAgent(c *gin.Context) {
```
edgarbellot, line 288: Is there an intentional reason for not having a per-user agent limit here?

Each creation triggers [refreshBotsAndNotify](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api_agents.go#L315) (315), which reloads all agents into memory via `EnsureBots`.

api/api_agents.go
```diff
@@ -0,0 +1,610 @@
...
+	}
+
+	accessible := make([]*llm.BotConfig, 0, len(agents))
+	for _, cfg := range agents {
+		if a.canUserAccessAgent(cfg, userID) {
+			accessible = append(accessible, cfg)
+		}
+	}
+
+	c.JSON(http.StatusOK, accessible)
```
edgarbellot, line 375: Can we strip `customInstructions` (and possibly `adminUserIDs`) from the response for non-admin/non-creator users?

Right now [handleListAgents](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api_agents.go#L336) (336) and [handleGetAgent](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api_agents.go#L359) (359) return the full `BotConfig` to any user that passes the [canUserAccessAgent](https://github.com/mattermost/mattermost-plugin-agents/blob/a0f586a84eb806108b7407ce5b5ca9a9b923c002/api/api_agents.go#L602) (602) check. `customInstructions` could contain sensitive organizational procedures or internal guidelines that shouldn't be visible to every user who can interact with the agent.

An agent that is created by an admin (where access to users --> all and channel --> a private channel where the user doesn't have access) exposes all the configuration:

Image

webapp/src/components/agents/tabs/access_tab.tsx
```diff
@@ -0,0 +1,111 @@
...
+                    <FormattedMessage defaultMessage='Agent admins'/>
+                </SectionTitle>
+                <SectionDescription>
+                    <FormattedMessage defaultMessage='These users can edit and delete this agent. The agent creator is always an admin.'/>
+                </SectionDescription>
+                <SelectUser
+                    userIDs={draft.adminUserIds}
+                    teamIDs={[]}
+                    onChangeIDs={(userIds: string[]) => onChange({adminUserIds: userIds})}
+                />
```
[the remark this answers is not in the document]
edgarbellot: Unresolving this since `teamIds` are still discarded. This is a small UX bug not a security concern.

edgarbellot (changes requested):
@nickmisasi great work! Left a few comments with some suggestions

api/api_agents.go
nickmisasi: Added

api/api.go
nickmisasi: Made some tweaks for this, good catch

api/api_agents.go
nickmisasi: Added, wrapped into IsValid

api/api_agents.go
nickmisasi: My thinking is that given this ability is gated behind manage_own permission (which can be granted or removed in scenarios where there may be untrusted users) we can trust that the people with the role are allowed to create them. The lookup is quick per node and then cached in memory so long lists shouldn't be very taxing.

Can consider limits in the future, when we look at more in depth permissions around tools that people can apply to agents and such.

api/api_agents.go
nickmisasi: I've added a sanitize, but the instructions shouldn't be sensitive because they're in the context and usually the user can just ask what they are :P

api/api_agents.go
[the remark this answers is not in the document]
nickmisasi: Same as the sibling thread: **`handleListServices`** and **`handleFetchModelsForService`** both enforce **`canConfigureAgentServices`** up front (ManageOwn / ManageOthers / ManageSystem), so this is not open to arbitrary authenticated users. The distinction vs **`canCreateAgent`** is intentional and covered by tests in `api/api_agents_test.go`.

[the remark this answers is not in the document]
nickmisasi: Thanks for the review — this doesn’t match the current code on this branch.

Both **`GET /services`** (`handleListServices`) and **`POST /agents/models/fetch`** (`handleFetchModelsForService`) call **`canConfigureAgentServices`** at the start of the handler, before any outbound work. That allows **`PermissionManageOwnAgent`**, **`PermissionManageOthersAgent`**, or **`PermissionManageSystem`** — not merely auth + license on the router.

Using **`canCreateAgent`** here would be a *different* policy: create does **not** grant **`ManageOthersAgent`**, while listing services / fetching models intentionally does (see `TestListServicesForbiddenWithoutManageOwnPermission`, `TestFetchModelsForbiddenWithoutManageOwnPermission`, and the ManageOthers variants in `api/api_agents_test.go`).

So the reported “any authenticated user” impact is **not valid** for the current implementation.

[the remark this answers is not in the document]
nickmisasi: Agree with the other replies on this line: service listing and model fetch are gated by **`canConfigureAgentServices`**, not only session auth + license. **`canCreateAgent`** is a narrower check by design. No change needed for the security claim as stated.

crspeller (approved):

api/api_agents.go
```diff
@@ -0,0 +1,680 @@
...
+// canUserAccessAgent reports whether userID may view or use the agent (admin, then usage restrictions).
+func (a *API) canUserAccessAgent(cfg *llm.BotConfig, userID string) bool {
+	if cfg == nil {
+		return false
+	}
+	if cfg.IsAdmin(userID) {
+		return true
+	}
+	return a.bots.CheckUsageRestrictionsForUserConfig(*cfg, userID) == nil
+}
```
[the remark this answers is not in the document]
nickmisasi: Fixed in the latest push: user-scope checks no longer go through `a.bots`. `canUserAccessAgent` now calls `bots.UsageRestrictionsForUserConfig(a.pluginAPI, …)`, with a nil `pluginAPI` guard. The shared logic lives in `bots/UsageRestrictionsForUserConfig` and `MMBots.CheckUsageRestrictionsForUserConfig` delegates to it so behavior stays identical.

edgarbellot (approved):
LGTM, thank you Nick!

api/api_config.go
```diff
@@ -23,20 +36,24 @@ func (a *API) handleGetConfig(c *gin.Context) {
...
 			WebSearch: config.WebSearchConfig{
 				DomainDenylist: []string{},
 			},
-		})
+		}))
 		return
 	}
 
-	c.JSON(http.StatusOK, cfg)
+	c.JSON(http.StatusOK, normalizeAdminConfig(*cfg))
```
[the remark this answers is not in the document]
nickmisasi: Addressed: `handleGetConfig` now runs `normalizeAdminConfig` on `cfg.Clone()` so normalization never writes through the cached store slice. Added `TestHandleGetConfigDoesNotMutateStoredServices` to lock this in.

api/api_agents.go
```diff
@@ -0,0 +1,682 @@
...
+	if !canConfigureAgentServices(a.pluginAPI, userID) {
+		c.AbortWithError(http.StatusForbidden, errors.New("user does not have permission to fetch models"))
+		return
+	}
+
+	var req FetchModelsForServiceRequest
+	if err := c.BindJSON(&req); err != nil {
+		c.AbortWithError(http.StatusBadRequest, fmt.Errorf("invalid request body: %w", err))
+		return
+	}
```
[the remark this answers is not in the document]
nickmisasi: Switched `handleFetchModelsForService` to `ShouldBindJSON` so parse failures match the other agent handlers and avoid double `AbortWithError` from Gin’s `BindJSON` path.

nickmisasi:
/test-analysis-override chimed in at the 11th hour after passing all day. Need to merge this for RC, so will address gaps in follow up PR.

PR #602, opened 2026-04-08
Conversation entities: first-class LLM conversation storage

crspeller:
#### Summary

Conversations become first-class entities stored in plugin tables (`LLM_Conversations`, `LLM_Turns`). Posts remain display artifacts, but the conversation model is the source of truth. A unified `ToolRunner` replaces the divergent auto-run code paths, and thread analysis, channel analysis, and search are all modeled as conversations.

**Key changes:**

- **Schema** — New `LLM_Conversations` and `LLM_Turns` tables with JSONB content blocks. `LLM_PostMeta` dropped.
- **Content blocks** — Typed JSONB blocks (text, thinking, tool_use, tool_result, file, image, annotations) with block-level `shared` flag for channel privacy.
- **Unified ToolRunner** — Single call-execute-recall loop replacing `WithAutoRunTools()`, `wrapStreamWithMCPAutoApproval()`, and all provider-level auto-run handling.
- **Conversation service** — Creates/continues conversations, builds `CompletionRequest` from turns (not posts), writes tool turns, generates titles, merges channel mention context.
- **API privacy filtering** — `GET /conversations/{id}` returns full content to owner, redacts unshared tool data for other channel members.
- **Streaming turn persistence** — Placeholder turn at stream start, finalized with content blocks + token counts at stream end.
- **DM conversations** — DM path uses conversation service + ToolRunner instead of `PostToAIPost`/`ThreadToLLMPosts`.
- **Channel mentions** — Multi-bot thread isolation via `(RootPostID, BotID)` lookup. Tool blocks default to `shared=false` in channels.
- **Thread analysis, channel analysis, search** — All create conversation entities with appropriate `Operation` types.
- **Webapp** — `LLMBotPost` rewritten to fetch from conversation API via `useConversation` hook. RHS threads backed by `LLM_Conversations`.
- **Cleanup** — Removed ~5,700 lines of old code: auto-run wrappers, KV store tool data, post-based conversation reconstruction, legacy post props.

**What stays independent (not converted):** Meeting summarization, emoji reactions, title generation, bridge API.

#### Ticket Link

N/A

#### Screenshots

N/A — no visual changes to the UI. Data source changed from post props to conversation entity API.

#### Release Note
```release-note
Conversations are now stored as first-class entities in new database tables (`LLM_Conversations`, `LLM_Turns`). Added new `GET /conversations/{id}` API endpoint with privacy filtering. Added new database migrations (000004-000006). Added `conversation_id` post prop. Removed `LLM_PostMeta` table. Pre-v2 bot posts render message text only (no tool calls, reasoning, or annotations).
```

## Release Notes

* **New Features**
  * Added persistent conversation storage with turn history tracking for direct messages and channel mentions.
  * Implemented conversation-based tool approval/rejection workflow allowing users to accept or decline tool calls.
  * Integrated search results directly into conversation history with persistent context.

* **Enhancements**
  * Improved tool call tracking with clearer separation of execution results and conversation flow.
  * Enhanced channel analysis by creating conversation records for auditability.
  * Better support for continuing conversations while maintaining context across interactions.


channels/channels.go
```diff
@@ -165,28 +219,55 @@ func (c *Channels) Interval(
...
+	botID string,
+	systemPrompt string,
+	userPrompt string,
+	promptName string,
+) (*AnalysisResult, error) {
+	// Create conversation entity
+	convResult, err := c.convSvc.CreateConversation(conversation.CreateConversationParams{
+		UserID:       userID,
+		BotID:        botID,
+		ChannelID:    &channelID,
```
nickmisasi: If I'm reading this right, we're keying off the channel_id then later in handleGetConversation it checks that the requester has access to the channel - for this scenario the result is delivered via DM, so we should probably enforce user == user

conversation/convert.go
```diff
@@ -0,0 +1,196 @@
...
+		switch block.Type {
+		case BlockTypeText:
+			textParts = append(textParts, block.Text)
+
+		case BlockTypeThinking:
+			// Last thinking block wins
+			post.Reasoning = block.Text
+			post.ReasoningSignature = block.Signature
+
+		case BlockTypeToolUse:
```
nickmisasi, line 45: We should check `redactUnshared` here too - tool parameters could be sensitive - in an Auto Approve round where it's not set to Everywhere, we want "Keep private" to also keep the tool call itself private

nickmisasi (changes requested):
Found a couple things on my first pass. 

Another thing that is sort of out of diff and more of a question - Data Retention - when these were in post props that was free, but I think as this is now we're soft deleting

This was probably already a problem with the KVStore approach to multiplayer tool calling

channels/channels.go
crspeller: Removed the channelID so this one works just like the DM ones.

conversation/convert.go
crspeller: Agreed.

store/migrations/000007_create_conversations_table.up.sql
```diff
@@ -0,0 +1,49 @@
+CREATE TABLE IF NOT EXISTS LLM_Conversations (
```
nickmisasi, line 1: 000005 already exists on master so this file needs to be renamed

store/migrations/000006_conversations_unique_index_per_user.up.sql
```diff
@@ -0,0 +1,11 @@
+-- Repair the unique index on LLM_Conversations for servers that applied an
```
nickmisasi: 000006 also already exists

nickmisasi: This also looks like it may be a subsequent migration to account for a refactor - can probably just consolidate to the one?

nickmisasi (approved):
Approving so as not to block on the small changes. 

Lot's here - focused my review especially around the multiplayer tool calling flow to ensure safety there. Things look in a good state, but we're going to need a thorough test and should aim to get this onto Hub ASAP for soak time.

conversations/conversations.go
```diff
@@ -100,446 +86,180 @@ func (c *Conversations) SetMeetingsService(meetingsService MeetingsService) {
...
-	result := make([]llm.Post, 0, len(threadData.Posts))
-
-	for _, post := range threadData.Posts {
-		aiPost := c.PostToAIPost(bot, post)
-
-		// Add username prefix for user messages in multi-user threads
-		if aiPost.Role == llm.PostRoleUser {
-			if user, exists := threadData.UsersByID[post.UserId]; exists {
-				aiPost.Message = "@" + user.Username + ": " + aiPost.Message
+			policy, _ := c.toolPolicyChecker.GetToolPolicy(origin, tc.Name)
```
edgarbellot: Nit: [allToolsAutoRunEverywhere](https://github.com/mattermost/mattermost-plugin-agents/blob/54f26b53badf3f583d18e01b86cd3f07a5252802/conversations/conversations.go#L248-L265) (248-265) discards the `enabled` return value from [GetToolPolicy](https://github.com/mattermost/mattermost-plugin-agents/blob/54f26b53badf3f583d18e01b86cd3f07a5252802/conversations/conversations.go#L258) (258) with `_`, while [shouldAutoExecuteTool](https://github.com/mattermost/mattermost-plugin-agents/blob/54f26b53badf3f583d18e01b86cd3f07a5252802/conversations/conversations.go#L234-L236) (234-236) correctly checks it.

The practical impact is narrow - it would only matter when an individual tool is disabled at the tool config level while its parent server is still enabled, AND the user manually approves execution from the UI. In that case the tool results would be written with `shared=true`, making them visible to other channel members without the requester explicitly choosing to share them.

Should we add the same `enabled` check here for consistency with `shouldAutoExecuteTool`? Something like:

```go
policy, enabled := c.toolPolicyChecker.GetToolPolicy(origin, tc.Name)
if !enabled || !mcp.IsToolPolicyAutoRunEverywhere(policy) {
    return false
}

File: streaming/streaming.go [hunk 6]
```diff
@@ -363,99 +317,152 @@ func (p *MMPostStreamService) FinishStreaming(postID string) {
...
 	broadcast := &model.WebsocketBroadcast{ChannelId: post.ChannelId}
 	p.sendPostStreamingControlEventWithBroadcast(post, PostStreamingControlStart, broadcast)
+
+	// Create turn accumulator if turn persistence is enabled and a conversation_id is set
+	var acc *turnAccumulator
+	if p.turnStore != nil {
+		if convID, ok := post.GetProp(ConversationIDProp).(string); ok && convID != "" {
+			isDM := false
+			if ch, chErr := p.mmClient.GetChannel(post.ChannelId); chErr == nil {
+				isDM = ch.Type == model.ChannelTypeDirect || ch.Type == model.ChannelTypeGroup
```
edgarbellot: The `isDM` definition in the [turnAccumulator](https://github.com/mattermost/mattermost-plugin-agents/blob/54f26b53badf3f583d18e01b86cd3f07a5252802/streaming/streaming.go#L448-L451) (448-451) includes `ChannelTypeGroup`

But everywhere else in the codebase (regeneration, tool_approval, handle_messages) isDM is resolved via [IsDMWith](https://github.com/mattermost/mattermost-plugin-agents/blob/54f26b53badf3f583d18e01b86cd3f07a5252802/mmapi/channels.go#L12-L17) (12-17), which only matches ChannelTypeDirect.

This means in group DMs the turnAccumulator writes tool_use blocks with shared=true ([streaming.go:136](https://github.com/mattermost/mattermost-plugin-agents/blob/54f26b53badf3f583d18e01b86cd3f07a5252802/streaming/streaming.go#L136)) on the final assistant turn, while the business logic treats group DMs as channels and writes intermediate tool turns with shared=false. The practical impact is low - tool_use inputs on the final response turn would be visible to other group DM members via the conversation API without the requester explicitly sharing them.

Should this use IsDMWith for consistency with the rest of the codebase, or is the broader definition intentional here?

edgarbellot (changes requested):
@crspeller great work! Just two nits.

conversations/conversations.go
crspeller: Fixed.

streaming/streaming.go
crspeller: No I think the narrow definition makes more sense. Fixed.

edgarbellot (approved):
LGTM! 🚀

PR #609, opened 2026-04-09
MCP: prefer static Authorization header over stored OAuth tokens

romzes5000:
## Summary

When an MCP server is configured with a **custom `Authorization` header** (e.g. Outline API key `ol_api_*`), the HTTP client must not wrap requests with `oauth2.Transport` using a **previously stored** per-user OAuth token from the plugin KV store.

Without this, switching from **OAuth** to **token-only** auth leaves stale `mcp_oauth_token_v1_*` rows; the plugin still calls `createOAuthConfig` / token refresh on every request, which can fail (e.g. DCR rate limits) even though the admin intends to use only the static header.

## Changes

- If custom headers include `Authorization` (case-insensitive), skip loading the OAuth token and do not wrap with `oauth2.Transport`.
- Add `TestAuthorizationHeaderInCustomHeaders`.

## Test steps

1. Configure MCP with OAuth, complete flow (token in KV).
2. Add `Authorization: Bearer <api key>` in custom headers and remove OAuth client fields if any.
3. Requests should succeed using the header without requiring manual KV cleanup.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced authorization header handling to detect and respect custom Authorization headers (case-insensitive) in HTTP requests.
  * Updated OAuth token behavior to conditionally apply based on whether a static Authorization header is provided, preventing conflicts and improving request handling flexibility.
  * Added comprehensive test coverage for authorization header detection across various scenarios.

### DIFF
--- a/mcp/http_client.go
+++ b/mcp/http_client.go
@@ -3,7 +3,10 @@
 
 package mcp
 
-import "net/http"
+import (
+	"net/http"
+	"strings"
+)
 
 // headerTransport is a custom RoundTripper that adds headers to requests
 type headerTransport struct {
@@ -23,15 +26,28 @@ func (t *headerTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	return t.base.RoundTrip(req)
 }
 
+// authorizationHeaderInCustomHeaders is true when MCP server custom headers
+// include Authorization. Then oauth2.Transport must not use KV OAuth tokens
+// (e.g. after switching from OAuth to API-key-only auth).
+func authorizationHeaderInCustomHeaders(headers map[string]string) bool {
+	for k := range headers {
+		if strings.EqualFold(k, "Authorization") {
+			return true
+		}
+	}
+	return false
+}
+
 func (c *Client) httpClientForMCP(headers map[string]string) *http.Client {
 	// Wrap with discovery-aware transport for 401 handling
 	authenticationTransport := &authenticationTransport{
-		userID:      c.userID,
-		serverName:  c.config.Name,
-		manager:     c.oauthManager,
-		serverURL:   c.config.BaseURL,
-		staticCreds: staticOAuthCreds(c.config),
-		base:        c.httpClient.Transport,
+		userID:                          c.userID,
+		serverName:                      c.config.Name,
+		manager:                         c.oauthManager,
+		serverURL:                       c.config.BaseURL,
+		staticCreds:                     staticOAuthCreds(c.config),
+		base:                            c.httpClient.Transport,
+		preferStaticAuthorizationHeader: authorizationHeaderInCustomHeaders(headers),
 	}
 
 	// Create HTTP client with discovery-aware transport
--- a/mcp/oauth_transport.go
+++ b/mcp/oauth_transport.go
@@ -20,6 +20,8 @@ type authenticationTransport struct {
 	manager     *OAuthManager
 	staticCreds *StaticOAuthCredentials
 	base        http.RoundTripper
+	// If true, skip oauth2.Transport; see authorizationHeaderInCustomHeaders.
+	preferStaticAuthorizationHeader bool
 }
 
 type mcpUnauthorized struct {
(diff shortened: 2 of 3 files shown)


mcp/oauth_transport.go
edgarbellot, line 109: Nice work on skipping the OAuth token load and `oauth2.Transport` when a static `Authorization` header is present - the [outgoing request path](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/oauth_transport.go#L67-L86) (67-86) looks correct.

However, should we also guard the [401 response handling block](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/oauth_transport.go#L109-L131) (109-131) with the same `preferStaticAuthorizationHeader` check? Right now, if a static-auth MCP server responds with a 401 (e.g. expired API key), the transport still returns an `mcpUnauthorized` error. The caller in [createSession](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/client.go#L293-L302) (293-302) then catches it via `errors.As` and calls [InitiateOAuthFlow](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/oauth_manager.go#L178-L213), which triggers outbound HTTP requests for metadata discovery ([discoverProtectedResourceMetadata](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/oauth_metadata.go#L34-L76), [discoverAuthorizationServerMetadata](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/oauth_metadata.go#L79-L131)) and potentially DCR ([loadOrCreateClientCredentials](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/oauth_manager.go#L76-L116)).

This means a compromised MCP server could craft its 401 `WWW-Authenticate` header to point the plugin at an attacker-controlled URL, causing the Mattermost server to make outbound requests to that URL during discovery. The `MattermostTransport` does block reserved IPs, which limits internal SSRF, but external endpoints are still reachable.

Would it make sense to short-circuit the 401 path when `preferStaticAuthorizationHeader` is true - something like returning a plain error instead of `mcpUnauthorized`? For example:

```go
if resp.StatusCode == http.StatusUnauthorized {
    if t.preferStaticAuthorizationHeader {
        drainAndCloseResponseBody(resp)
        return nil, fmt.Errorf("static authorization rejected by server (HTTP 401)")
    }
    // ... existing OAuth 401 handling
}

mcp/http_client.go
edgarbellot, line 17: This is a pre-existing issue in [headerTransport](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/http_client.go#L12-L27), but I think this PR makes it worth flagging because the credential at stake changes significantly.

Go's `http.Client` strips the `Authorization` header on cross-domain redirects (since Go 1.12) before calling `Transport.RoundTrip()`. However, [headerTransport.RoundTrip](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/http_client.go#L17-L27) unconditionally re-adds all configured headers - including `Authorization` - on every call. Since `http.Client.Do()` calls `RoundTrip` again for each redirect hop, this effectively bypasses Go's built-in protection.

Before this PR, the `Authorization` header would typically be overwritten by `oauth2.Transport` with a short-lived OAuth token, so a leaked credential would expire quickly. With [preferStaticAuthorizationHeader](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/http_client.go#L50) set to `true`, `oauth2.Transport` is skipped and the static API key flows through unchanged. That means a compromised MCP server responding with a `302 Location: https://attacker.com/steal` would receive the long-lived API key in the redirected request - silently, with no error or log.

As mentioned in the other comment, the [MattermostTransport](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/http_client.go#L50) blocks reserved/internal IPs, which mitigates internal SSRF, but external hosts are fully reachable. There's also no `CheckRedirect` policy on the [http.Client built by httpClientForMCP](https://github.com/mattermost/mattermost-plugin-agents/blob/2e8dbbdc/mcp/http_client.go#L54), so Go's default 10-redirect limit applies with no domain restrictions.

Would it make sense to add a `CheckRedirect` policy that strips credentials on cross-domain redirects? Something like:

```go
httpClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
    if len(via) >= 10 {
        return errors.New("stopped after 10 redirects")
    }
    if len(via) > 0 && req.URL.Host != via[0].URL.Host {
        req.Header.Del("Authorization")
    }
    return nil
}

edgarbellot (changes requested):
@romzes5000 Thanks for the contribution! The approach of skipping `oauth2.Transport` when a static `Authorization` header is present makes sense and the tests look clean.

I left a couple of comments on areas where the new flag could be checked more broadly to keep the static-auth path fully isolated from the OAuth machinery. We might need to update the test to cover the edge cases mentioned in the comments. Happy to provide more context if needed :)

nickmisasi (commented):
Thanks for the contribution! Just leaving a comment so you can re request review once you've addressed Edgar's requested changes.

nickmisasi:
Looks like this has gone stale. I'm going to close it for now - if you pick it back up feel free to re open.

PR #611, opened 2026-04-10
Restrict bot-triggered channel tools to auto-run everywhere

nickmisasi:
#### Summary
When a bot opts into the agent via `activate_ai` on a channel mention, channel tool calling is now allowed, but only MCP tools configured as **Auto Run (Everywhere)** are exposed and auto-approved. Built-in Mattermost tools and are omitted in this path so automation does not stall on interactive approval. The restriction is recorded on AI posts (`channel_tools_auto_run_everywhere_only`) so follow-up tool handling and regeneration stay consistent.

**QA / test steps**
- Enable channel mention tool calling; configure at least one MCP tool as `auto_run_everywhere` and optionally others as `ask` or `auto_run`.
- Post as a bot user with `@agent` and `activate_ai`; confirm only everywhere tools run without pending approval.
- Confirm OAuth-backed MCP servers still behave as before for bot users (no token).

#### Ticket Link
N/A

#### Screenshots
N/A (server-side behavior)

#### Release Note
```release-note
Channel mentions from bots with `activate_ai` can use MCP tools only when each tool is enabled and set to Auto Run (Everywhere); built-in and native provider tools are not offered in that flow.
```

## Summary by CodeRabbit

* **New Features**
  * Channel-level tool filtering: restricts available tools in non-DM channel flows to those marked for auto-run-everywhere and records disabled tool info; new post flag persists this strict mode and is threaded through request/regeneration and tool-handling flows.
  * MCP auto-approval can operate in a stricter mode requiring auto-run-everywhere policy.

* **Tests**
  * Added unit tests covering filter behavior, strict auto-approval, and interactions with existing tool-availability logic.

### DIFF
--- a/conversations/bot_channel_tool_filter_test.go
+++ b/conversations/bot_channel_tool_filter_test.go
@@ -93,3 +93,21 @@ func TestApplyToolAvailabilityBeforeBotChannelFilterPreservesDisabledToolsInfo(t
 	}
 	require.ElementsMatch(t, []string{"builtin", "ask_tool"}, disabledNames)
 }
+
+func TestApplyBotChannelAutoEverywhereToolFilter_nilCheckerFailClosed(t *testing.T) {
+	origin := "https://mcp.example.com/mcp"
+	c := &Conversations{toolPolicyChecker: nil}
+
+	llmContext := &llm.Context{
+		Tools: llm.NewToolStore(nil, false),
+	}
+	llmContext.Tools.AddTools([]llm.Tool{
+		{Name: "builtin", ServerOrigin: "", Resolver: func(*llm.Context, llm.ToolArgumentGetter) (string, error) { return "", nil }},
+		{Name: "mcp_tool", ServerOrigin: origin, Resolver: func(*llm.Context, llm.ToolArgumentGetter) (string, error) { return "", nil }},
+	})
+
+	c.applyBotChannelAutoEverywhereToolFilter(llmContext)
+
+	require.Empty(t, llmContext.Tools.GetTools())
+	require.Len(t, llmContext.DisabledToolsInfo, 2)
+}
--- a/conversations/bot_channel_tool_filter_test.go
+++ b/conversations/bot_channel_tool_filter_test.go
@@ -1,0 +1,95 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package conversations
+
+import (
+	"testing"
+
+	"github.com/mattermost/mattermost-plugin-agents/llm"
+	"github.com/mattermost/mattermost-plugin-agents/mcp"
+	"github.com/stretchr/testify/require"
+)
+
+type mapPolicyChecker map[string]map[string]struct {
+	policy  string
+	enabled bool
+}
+
+func (m mapPolicyChecker) GetToolPolicy(serverOrigin, toolName string) (string, bool) {
+	byServer, ok := m[serverOrigin]
+	if !ok {
+		return mcp.ToolPolicyAsk, false
+	}
+	cfg, ok := byServer[toolName]
+	if !ok {
+		return mcp.ToolPolicyAsk, true
+	}
+	return cfg.policy, cfg.enabled
+}
+
+func TestApplyBotChannelAutoEverywhereToolFilter(t *testing.T) {
+	origin := "https://mcp.example.com/mcp"
+	c := &Conversations{
+		toolPolicyChecker: mapPolicyChecker{
+			origin: {
+				"everywhere_tool": {policy: mcp.ToolPolicyAutoRunEverywhere, enabled: true},
+				"auto_run_tool":   {policy: mcp.ToolPolicyAutoRun, enabled: true},
+				"ask_tool":        {policy: mcp.ToolPolicyAsk, enabled: true},
+			},
+		},
+	}
+
+	llmContext := &llm.Context{
+		Tools: llm.NewToolStore(nil, false),
+	}
+	llmContext.Tools.AddTools([]llm.Tool{
+		{Name: "builtin", ServerOrigin: "", Resolver: func(*llm.Context, llm.ToolArgumentGetter) (string, error) { return "", nil }},
+		{Name: "everywhere_tool", ServerOrigin: origin, Resolver: func(*llm.Context, llm.ToolArgumentGetter) (string, error) { return "", nil }},
+		{Name: "auto_run_tool", ServerOrigin: origin, Resolver: func(*llm.Context, llm.ToolArgumentGetter) (string, error) { return "", nil }},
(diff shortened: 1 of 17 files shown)


nickmisasi:
re: dryrun

I think this is a false positive given the nuance here. 

1. The post that would be edited is a bot post, not a user post, you can't edit it
2. We don't even expose regeneration for in-channel agent responses

conversations/tool_handling.go
```diff
@@ -191,6 +191,9 @@ func (c *Conversations) HandleToolCall(userID string, post *model.Post, channel
 		contextOpts...,
 	)
 	toolsDisabled := applyToolAvailability(llmContext, isDM, allowToolsInChannel)
+	if !isDM && channelToolsAutoRunEverywhereOnlyFromPost(post) {
+		c.applyBotChannelAutoEverywhereToolFilter(llmContext)
+	}
```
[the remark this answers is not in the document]
nickmisasi: Addressed in `applyBotChannelAutoEverywhereToolFilter` (same helper used from both call sites here and in `ProcessUserRequestWithContext`). When `c.toolPolicyChecker` is nil we no longer no-op: we remove all tools from the store, set `DisabledToolsInfo` to the removed tool list (replacing any prior channel-disabled list so we don’t duplicate), and return—so replayed posts with `channel_tools_auto_run_everywhere_only` cannot fall back to the full tool set without policy validation. See `TestApplyBotChannelAutoEverywhereToolFilter_nilCheckerFailClosed` in `bot_channel_tool_filter_test.go`.

conversations/handle_messages.go
```diff
@@ -49,10 +49,29 @@ func isAutomatedInvoker(post *model.Post, postingUser *model.User) bool {
...
+}
+
 // computeAllowToolsInChannel returns whether tools should be allowed for a channel mention,
-// given the config flag and whether the invoker is automated.
-func computeAllowToolsInChannel(configEnabled bool, post *model.Post, postingUser *model.User) bool {
-	return configEnabled && !isAutomatedInvoker(post, postingUser)
+// given the config flag and whether the invoker is automated. Bot activate_ai requires a
+// tool policy checker: without it, strict filtering and MCP auto-approval are no-ops and tools
+// must stay disabled so automated invokers cannot strand pending approvals.
+func computeAllowToolsInChannel(configEnabled bool, post *model.Post, postingUser *model.User, hasToolPolicyChecker bool) bool {
```
edgarbellot, line 67: Can we add a warning in the system console UI near the policy picker making it explicit that `auto_run_everywhere` means "this tool will execute automatically without user confirmation when triggered by bots in channels"? Something that helps admins understand the trust boundary they're accepting when they assign that policy to a tool.

conversations/mcp_auto_approval.go
edgarbellot, line 84: This isn't introduced by this PR, but the new bot `activate_ai` flow makes it more relevant since errors now reach the channel automatically without human review.

When a tool call fails here, `err.Error()` is stored as the tool result and eventually passed to the LLM as conversation context. The LLM then generates a response that gets streamed to a public channel post. The error content comes from two sources:

1. Our own `fmt.Errorf` wrappers in [client.go](https://github.com/mattermost/mattermost-plugin-agents/blob/287547228e23c7d5a094ab2d41d8946a338cfe74/mcp/client.go#L410-L420) (lines 410-420) - these include `c.config.Name` and wrap transport errors from the MCP SDK, which can contain connection details (hostnames, ports, TLS errors).
2. Remote MCP server error responses in [client.go](https://github.com/mattermost/mattermost-plugin-agents/blob/287547228e23c7d5a094ab2d41d8946a338cfe74/mcp/client.go#L436-L441) (lines 436-441) - when `IsError=true`, the server's text content is passed through as-is via `errors.New(trimmed)`. We don't control what remote servers put in their error messages.

Pre-PR, a human would see the error in the approval UI and could choose not to share it. Post-PR, in the bot `activate_ai` + `auto_run_everywhere` flow, the full chain is automatic: error → LLM context → LLM response → public channel post.

The same pattern exists in [HandleToolCall](https://github.com/mattermost/mattermost-plugin-agents/blob/287547228e23c7d5a094ab2d41d8946a338cfe74/conversations/tool_handling.go#L209-L212) (lines 209-212).

Would it make sense to sanitize these errors before storing them as tool results - something like replacing the raw error with a generic message (e.g., "Tool execution failed") and logging the full error server-side? That way the LLM gets enough context to respond gracefully without risking leaking internal details from remote MCP servers into a public channel.

conversations/bot_channel_tool_filter_test.go
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
```
edgarbellot, line 3: Do you think it is worth adding integration-level tests for the flows that read `channelToolsAutoRunEverywhereOnly` from a post prop and act on it - specifically [HandleToolCall](https://github.com/mattermost/mattermost-plugin-agents/blob/287547228e23c7d5a094ab2d41d8946a338cfe74/conversations/tool_handling.go#L194-L196) (lines 194-196), [HandleToolResult](https://github.com/mattermost/mattermost-plugin-agents/blob/287547228e23c7d5a094ab2d41d8946a338cfe74/conversations/tool_handling.go#L409-L411) (lines 409-411), and [completeAndStreamToolResponse](https://github.com/mattermost/mattermost-plugin-agents/blob/287547228e23c7d5a094ab2d41d8946a338cfe74/conversations/tool_handling.go#L508-L525) (lines 508-525) propagating the prop to the response post? These are the paths where the security decision is consumed, so having test coverage there would help catch regressions if the flow changes later.

edgarbellot (approved):
@nickmisasi great rework! No major findings, just a few suggestions - making them non-blocking. Thank you!

conversations/handle_messages.go
nickmisasi: I think I'm going to defer this to a future ticket because I'm not sure where on the screen to put it. 
image


Note for the documentation agent: this is an important thing to clarify about the tool calling policies

conversations/mcp_auto_approval.go
nickmisasi: We actually recently made changes to enhance the error that's returned to the LLM so that it can intelligently either retry with different parameters or ask the user for clarification (think "404 team not found" -> "I couldn't find a team with that ID, can you double check the name?"). Going generic would gimp this ability pretty badly. 

I think we have enough guards here:
1. The admin has opted in to this functionality (the default isn't to allow everywhere, its to allow in DM)
2. The bot account is only exercising tools that everyone else in the server has access too with their own accounts
3. I feel like us being too defensive for external services' bad http errors is a bit of a slipper slope :P 

I am thinking we leave this as is for now.

BenCookie95 (approved):

conversations/bot_channel_tool_filter_test.go
nickmisasi: Working on a followup to this right now, will merge in a separate PR

PR #619, opened 2026-04-14
Fix OAuth discovery to strip path from MCP server URL

crspeller:
## Summary

- Fix OAuth metadata discovery to strip the path component from MCP server URLs before constructing `.well-known` URLs, per the MCP spec's authorization base URL requirement
- Add regression tests for both `GetRegistrationEndpoint` and `createOAuthConfig` with path-bearing server URLs

## Details

When connecting to MCP servers whose URL contains a path (e.g., `https://mcp.atlassian.com/v1/mcp`), OAuth discovery failed with:

```
server metadata request to https://mcp.atlassian.com/.well-known/oauth-authorization-server/v1/mcp failed with HTTP 404
```

**What went wrong:** The code passed the full MCP server URL (including path `/v1/mcp`) to `constructWellKnownURL`, which per RFC 8414 Section 3.1 inserts `/.well-known/oauth-authorization-server` between the host and path, producing `/.well-known/oauth-authorization-server/v1/mcp`. This path-aware construction is correct when the input is an OAuth **issuer identifier** with a path (for multi-tenant hosting), but the MCP server URL is a **resource URL**, not an issuer.

**What the MCP spec says:** The authorization base URL MUST be derived by discarding the path component from the MCP server URL. For `https://mcp.atlassian.com/v1/mcp`, the base URL is `https://mcp.atlassian.com`, and metadata should be fetched from `https://mcp.atlassian.com/.well-known/oauth-authorization-server`.

**What changed:**
- `oauth_manager.go`: `createOAuthConfig` now uses `baseURL` (path stripped) instead of `serverURL` for the fallback authorization server metadata discovery and the default `authServerURL`
- `dcrp.go`: `GetRegistrationEndpoint` now strips the path from the server URL before constructing the well-known URL

The protected resource metadata discovery (RFC 9728) path is unchanged — it correctly uses the full server URL with path, since RFC 9728 operates on the resource identifier.

## Test plan
- [x] `TestGetRegistrationEndpoint_ServerURLWithPath` — verifies `GetRegistrationEndpoint` strips the path and discovers metadata at the base URL
- [x] `TestCreateOAuthConfig_FallbackStripsPathFromServerURL` — verifies the full discovery fallback chain: protected resource metadata 404 → auth server metadata discovered at base URL
- [x] All existing `mcp` package tests continue to pass
- [x] Manually verified against `https://mcp.atlassian.com/v1/mcp`

## Bug Fixes
* Fixed OAuth server discovery to properly handle server URLs containing path components. The system now correctly strips the path and uses only the scheme and host for OAuth configuration discovery, ensuring reliable metadata retrieval when server URLs include additional path segments.

### DIFF
--- a/mcp/dcrp.go
+++ b/mcp/dcrp.go
@@ -198,9 +198,16 @@ func DiscoverAndRegisterClient(ctx context.Context, httpClient *http.Client, ser
 
 // GetRegistrationEndpoint discovers the registration endpoint from server metadata
 func GetRegistrationEndpoint(ctx context.Context, httpClient *http.Client, serverURL string) (string, error) {
+	// Strip path from server URL per MCP spec: the authorization base URL is derived
+	// by discarding the path component from the server URL.
+	parsedURL, err := url.Parse(serverURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse server URL %s: %w", serverURL, err)
+	}
+	baseURL := fmt.Sprintf("%s://%s", parsedURL.Scheme, parsedURL.Host)
+
 	// Construct the metadata URL according to RFC 8414 Section 3.1
-	// The well-known URI must be inserted between the host and path components
-	metadataURL, err := constructWellKnownURL(serverURL, "oauth-authorization-server")
+	metadataURL, err := constructWellKnownURL(baseURL, "oauth-authorization-server")
 	if err != nil {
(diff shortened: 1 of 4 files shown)


nickmisasi (approved):

edgarbellot (approved):
LGTM!

PR #621, opened 2026-04-15
MCP OAuth and tool provider auth adjustments

nickmisasi:
#### Summary

This change tightens MCP OAuth and tool-auth behavior on the server, trims obsolete conversation helpers, and updates the webapp tool-provider UX (RHS popover, toggle layout, system console row). It adds MCP OAuth e2e coverage with a small mock helper and registers the spec in the e2e CI group file.

**Server / API**
- Refine OAuth and MCP API handling (`api_oauth`, `api_mcp`, tests).
- Adjust MCP client session wiring, OAuth manager/store, and user MCP clients.
- Remove unused code from `conversations`.

**Webapp**
- Tool provider popover and toggle switch layout for consistent alignment.
- Client updates for disconnect / tool flows; minor system console row tweak.

**E2E**
- New `mcp-oauth-auth` spec and `mcp-oauth-mock` helper; shard registration in `ci-test-groups.mjs`.

#### Ticket Link

N/A

#### Screenshots

image
image
image

#### Release Note

```release-note
Improved MCP OAuth handling and tool provider controls in the Agents plugin; added e2e tests for MCP OAuth authorization flows.
```

## Summary by CodeRabbit

* **New Features**
  * Per-server OAuth endpoints to start flows and disconnect per-user OAuth sessions; server lists now include needsOAuth and authURL fields.

* **UI**
  * Provider rows show Connect/Disconnect actions, open provider auth, and refresh state after connect/disconnect; updated OAuth description text.

* **Tests**
  * Added unit and end-to-end tests and e2e helpers covering OAuth start/redirect/callback, disconnect endpoint, token handling, and related UI flows.


api/api_oauth.go
```diff
@@ -4,95 +4,122 @@
...
+	if !ok {
+		a.renderOAuthErrorPage(c, http.StatusNotFound, "Authorization Failed", "The selected MCP server was not found.")
+		return
+	}
+	if !serverConfig.Enabled || serverConfig.BaseURL == "" {
+		a.renderOAuthErrorPage(c, http.StatusBadRequest, "Authorization Failed", "The selected MCP server is not available for OAuth.")
+		return
+	}
+
+	metadataURL := c.Query("resource_metadata")
```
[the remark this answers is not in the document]
edgarbellot: **On the SSRF itself:** the HTTP client used for these requests is Mattermost's untrusted client, created via `MakeClient(false)` in [main.go](https://github.com/mattermost/mattermost-plugin-agents/blob/879ba3b638766d08005c4e0935aeac389fd954ab/server/main.go#L127). This client already blocks connections to private/loopback/link-local IP ranges by default, so the `169.254.169.254` cloud metadata scenario is mitigated unless the admin has explicitly relaxed `AllowedUntrustedInternalConnections`. The SSRF to external URLs is still valid, but the internal network scanning risk is lower than described.

**The bigger concern is credential exfiltration.** The attacker-controlled `resource_metadata` value is [persisted in the OAuth session](https://github.com/mattermost/mattermost-plugin-agents/blob/879ba3b638766d08005c4e0935aeac389fd954ab/mcp/oauth_manager.go#L237) as `ServerMetadataURL`. When the OAuth callback fires, [ProcessCallback](https://github.com/mattermost/mattermost-plugin-agents/blob/879ba3b638766d08005c4e0935aeac389fd954ab/mcp/oauth_manager.go#L294) re-runs discovery using that stored value, which means the attacker still controls the `token_endpoint`. At the same time, the admin-configured `client_secret` is [re-derived from the live plugin config](https://github.com/mattermost/mattermost-plugin-agents/blob/879ba3b638766d08005c4e0935aeac389fd954ab/mcp/oauth_manager.go#L284-L286) and placed in the [oauth2.Config](https://github.com/mattermost/mattermost-plugin-agents/blob/879ba3b638766d08005c4e0935aeac389fd954ab/mcp/oauth_manager.go#L189-L191). The [Exchange](https://github.com/mattermost/mattermost-plugin-agents/blob/879ba3b638766d08005c4e0935aeac389fd954ab/mcp/oauth_manager.go#L301-L302) call then sends `client_id` and `client_secret` to the attacker's endpoint.

This means a regular user can extract the shared OAuth client credentials configured by the admin for any MCP server. The victim only needs to click a legitimate-looking Mattermost link; the redirects can complete automatically without any interaction on the attacker's page.

I think the suggested fix of matching the metadata URL's host against the configured `BaseURL` would address both the SSRF and this credential leak. Alternatively, dropping `resource_metadata` from user input entirely and always deriving it from the admin-configured server would close the whole attack surface.

conversations/conversations.go
```diff
@@ -265,54 +265,22 @@ func (c *Conversations) ProcessUserRequest(bot *bots.Bot, postingUser *model.Use
...
-	var disabledOrigins map[string]bool
 	if channel.Type == model.ChannelTypeDirect || channel.Type == model.ChannelTypeGroup {
 		prefs, err := mcp.LoadUserPreferences(c.mmClient, postingUser.Id)
 		if err != nil {
 			c.mmClient.LogWarn("Failed to load user tool preferences, proceeding without filtering", "error", err.Error(), "userID", postingUser.Id)
 		} else if len(prefs.DisabledServers) > 0 {
-			disabledOrigins = make(map[string]bool, len(prefs.DisabledServers))
-			for _, origin := range prefs.DisabledServers {
-				disabledOrigins[origin] = true
-			}
```
edgarbellot, line 279: I see that `sendOAuthNotifications` was removed here. The previous behavior sent an ephemeral post directly in the conversation thread when an MCP server needed OAuth, which worked across all Mattermost clients.

The new approach relies on the `needsOAuth`/`authURL` fields in the `GET /mcp/tools` response and the Connect/Disconnect buttons in the RHS popover - which is a webapp-only component.

Could this create a gap for users on mobile clients or API integrations? If someone messages the bot from the Mattermost mobile app and an MCP server requires OAuth, the tools would silently fail with no indication that authentication is needed. The previous ephemeral post would have surfaced that information regardless of the client.

Not necessarily suggesting keeping the old approach as-is - just wanted to flag this in case it hasn't been considered. Would it make sense to keep a lightweight server-side fallback (e.g. a simpler ephemeral message) for non-webapp clients, or is there a plan to handle this at a different layer?

mcp/client_manager.go
edgarbellot: Nit: [DisconnectUserOAuth](https://github.com/mattermost/mattermost-plugin-agents/blob/879ba3b638766d08005c4e0935aeac389fd954ab/mcp/client_manager.go#L210-L211) correctly calls `uc.Close()` before deleting from the map, but this older method doesn't. Should this also close the client before deleting?

edgarbellot (changes requested):
@nickmisasi great work! Left some comments and suggestions. Happy to clarify anything if needed

mcp/user_clients.go
```diff
@@ -81,9 +82,18 @@ func (c *UserClients) ConnectToRemoteServers(servers []ServerConfig) *Errors {
 		}
 	}
 
+	c.connectionErrors = mcpErrors
 	return mcpErrors
 }
 
+// popConnectionErrors returns and clears the last connect-time errors so
+// cached UserClients lookups do not keep surfacing stale OAuth state.
+func (c *UserClients) popConnectionErrors() *Errors {
```
crspeller: Do we really need this?

mcp/client.go
```diff
@@ -172,9 +191,23 @@ func NewClient(ctx context.Context, userID string, serverConfig ServerConfig, lo
 		return nil, fmt.Errorf("failed to create MCP session for server %s: %w", serverConfig.Name, err)
 	}
 
-	// Try to get tools from global cache first
 	serverID := serverConfig.Name
-	if toolsCache != nil {
+	useSharedToolsCache := shouldUseSharedToolsCache(serverConfig)
+	if !useSharedToolsCache && toolsCache != nil && oauthManager != nil {
```
crspeller: This whole section is a bit hard to reason though. Wonder if we can simplify by splitting up some of the logic or making it more obvious.

crspeller (approved):
Some optional nits.

conversations/conversations.go
nickmisasi: Thanks for pointing this out - definitely a gap, but something that we'll need to fix with a proper porting of this functionality to mobile. The ephemeral post is spammy, and one of the biggest complaints we've had so it's removal is intentional. I have https://mattermost.atlassian.net/browse/MM-68354 to track this effort. 

For now, connections can't be initiated from mobile. I think there already would have been some issues with this, as the redirect URL would have not been the proper deeplink

api/api_oauth.go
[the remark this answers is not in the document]
nickmisasi: I've added the suggested mitigation - should be good now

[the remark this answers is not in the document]
edgarbellot: It is! that was quick, thank you!

edgarbellot (approved):

PR #634, opened 2026-04-20
Add before and after hooks for MCP tool calls through the bridge

BenCookie95:
#### Summary
https://mattermost.atlassian.net/wiki/spaces/DE/pages/4473028618/MCP+tool+hooks has most of the information

Some extra changes I had to make:
- You don't have to manually add a before and after hook to each tool, their execution is handled inside registerTool. In order to accommodate that I had to change the return type of every tool to a proper struct and then call the formatter in `registerTool`.
- I removed the filtering for hiding the automation tools if you weren't a channel admin/sysadmin. This was causing issues because I had to kill any existing MCP connection if you changed channel. This meant that you can't make any bridge calls while an MCP tool is executing because the bridge call would kill the connection you needed for the earlier MCP tool. I'm less concerned about context being filled up now though because we have the automation instructions hidden behind a tool call. 
- We run after hooks for both success and errors. Sometimes the error could leak something that a plugin might not want surfaced. Right now we are just giving a string to the afterHook but in future we can maybe give more detailed information to redact certain things.   

Screenshot 2026-04-20 at 1 38 25 PM 

#### Release Note

```release-note
Add before and after hooks to MCP tool calls
```

## Release Notes

* **New Features**
  * Added tool execution hooks allowing plugins to intercept and transform AI tool calls before and after execution
  * System administrator role is now included in user context information

* **Improvements**
  * Tool outputs are now more consistently structured and include plugin annotations
  * Tool accessibility is no longer restricted by channel membership context


prompts/standard_personality_without_locale.tmpl
```diff
@@ -37,6 +37,7 @@ CRITICAL: Do NOT attempt to invoke, simulate, or produce any tool call output fo
 Username: '{{.RequestingUser.Username}}'
 {{if .RequestingUser.FirstName}}Full name: {{.RequestingUser.FirstName}} {{.RequestingUser.LastName}}
 {{end}}{{if .RequestingUser.Position}}Position: '{{.RequestingUser.Position}}'
+{{end}}{{if .RequestingUser.IsSystemAdmin}}System role: System Administrator
```
BenCookie95: Since I removed the automation visibility tool step I added to the automation instructions who can build automations. The llm needs to know if they're dealing with a sysadmin.

BenCookie95:
@coderabbitai full review

mcpserver/tools/hooks.go
```diff
@@ -0,0 +1,211 @@
...
+		return "", fmt.Errorf("missing hook plugin id")
+	}
+	if callbackPath == "" {
+		return "", fmt.Errorf("empty callback path")
+	}
+	if !strings.HasPrefix(callbackPath, "/") {
+		return "", fmt.Errorf("callback path must start with /")
+	}
+	base := strings.TrimRight(baseURL, "/")
+	return fmt.Sprintf("%s/plugins/%s%s", base, pluginID, callbackPath), nil
```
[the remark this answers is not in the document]
BenCookie95: PluginID and hooks come from trusted plugins

mcp/client.go
```diff
@@ -133,10 +131,13 @@ func (c *EmbeddedServerClient) CreateClient(ctx context.Context, userID, session
 		if mmSession == nil {
 			return nil, fmt.Errorf("session not found")
 		}
+		if mmSession.UserId != userID {
```
BenCookie95, line 134: Suggested by dry run security bot, makes sense to check

hmohammed-prodsec (approved):

hmohammed-prodsec (approved):

hmohammed-prodsec (approved):

prompts/standard_personality_without_locale.tmpl
```diff
@@ -53,6 +53,9 @@ Their name is{{if .RequestingUser.FirstName}} {{.RequestingUser.FirstName}}{{end
 {{- if .RequestingUser.Position}}
 Their position is '{{.RequestingUser.Position}}'.
 {{- end}}
+{{- if .RequestingUser.IsSystemAdmin}}
+Their system role is System Administrator.
```
crspeller, line 64: No sure if this is helpful without further context?

crspeller (changes requested):
Is it possible to reduce this PR to only the before hooks? They have less complexity cost and I think everything that you want to do for your plugin can be done with before hooks. 

All of the DTOs add a lot of complexity that we have to manage here and I am not sure that we will get our complexity's worth out of them. 

Consumers can still reject and override parameters, they just can't rewrite output.

prompts/standard_personality_without_locale.tmpl
BenCookie95: I added this because in the create automation instructions it says you need to be a sysadmin or a channel admin to run an automation. We can return the user's role in the channel in the channel info but we also need to surface if the user is a sysadmin somehow. This felt like the place to put it

BenCookie95:
@crspeller @hmohammed-prodsec After hooks removed, let me know if you have anymore questions

hmohammed-prodsec (approved):

BenCookie95:
@dryrunsecurity fp drs_c8349710 Mattermost-Plugin-ID is a server-trusted header. The Mattermost server strips it from all external client requests in servePluginRequest

wiggin77:
@crspeller @BenCookie95 what is the ETA to get this reviewed and merged?

llm/context.go
```diff
@@ -47,6 +54,8 @@ type Context struct {
 	Tools             *ToolStore
 	DisabledToolsInfo []ToolInfo // Info about tools that are unavailable in the current context (e.g., DM-only tools in a channel)
 	Parameters        map[string]interface{}
+
+	MCPInvocation *MCPInvocationContext
```
crspeller: This doesn't feel like it belongs here. This is the collection of stuff that we build the prompt from. 
This feels more like additional parameters per tool?

crspeller (changes requested):
A lot less complexity. Just one comment.

llm/context.go
BenCookie95: The reason I avoided putting this into tool parameters is that the before-hook key is runtime/protocol metadata, not model-chosen tool input. If we make it a parameter, we either have to add it to every tool schema/args shape or wrap every tool with hidden bound params, and then make sure it never leaks into the actual tool semantics. 

There is some precedent for `llm.Context` carrying non-template execution data already: `Tools` carries resolver state/tool schemas, `BotUserID` is used for MCP call metadata and token tracking, and `Parameters` is also used by web search as per-run state, not just static prompt inputs.

What I’m trying to represent here is: “for this completion run, when resolving tools from this MCP server, attach this MCP call metadata.” It is forwarded through MCP `CallToolParams.Meta`, not shown to the LLM and not part of the tool input schema. Right now we don't have a good way to inject runtime things to a tool call from agents to the embedded server, kind of stuck with the metadata field.

I'm not sure of a good place for this, maybe the `Parameters` field above my new `MCPInvocation` field?

BenCookie95:
@crspeller @hmohammed-prodsec I had to change how we pass beforeHook callback endpoints around. Previously we just passed them down through the `_meta` field which meant that even people using the external MCP server could set their own callback endpoints which was not the intention. We only want beforeHook callbacks coming from trusted sources (the bridge).

The bridge now stores each trusted before-hook callback URL in KV with a short-lived key, and passes only that key in embedded MCP metadata.

During embedded MCP tool execution, the MCP server resolves the key through an in-process resolver callback, validates the tool/user binding, then calls the stored callback URL. Keys are reused for the bridge request if the LLM makes multiple calls to the same tool in 1 request, it's deleted when it completes, and TTL is a fallback cleanup.

https://github.com/mattermost/mattermost-plugin-agents/pull/634/commits/85ea47105210465bab1039a8789a3d939dc29c90

hmohammed-prodsec (approved):

crspeller (approved):

PR #689, opened 2026-04-28
Fix reindex Cancel/Resume race in HA + idempotent migration indexes

crspeller:
#### Summary

Two related fixes to make the embedding-search reindex feature reliable in HA Mattermost deployments.

**1. Cancel→Resume race in `indexer/`.** In an HA deployment with a read replica, pressing Cancel and then Resume on the reindex job left the system wedged for 10 minutes. The plugin KV API splits reads (replica) and writes (master), so the resumed worker's first ``KVGet`` could observe the *previous* run's ``Status: canceled`` while replication was still in flight, log ``"Reindex job was canceled"``, and exit — leaving a ``Status: running`` row on master with no live worker. ``StaleJobThreshold`` (10 min) was the only path that self-healed.

The fix mirrors how Mattermost's ``JobServer`` avoids the same problem:
- Every ``StartReindexJob`` / ``StartCatchUpJob`` assigns a fresh ``JobID`` (``model.NewId()``).
- ``CancelJob`` CAS-transitions ``running → cancel_requested`` (a new intermediate state) instead of writing ``canceled`` directly.
- The worker's cancel check is ``JobID``-scoped: it only exits when ``current.JobID == myJobID && current.Status == cancel_requested``, then CASes to ``canceled``.
- A stale replica read for *any other* ``JobID`` is silently ignored — the race is dissolved at the source.

A new ``KVCompareAndSet`` primitive on ``mmapi.Client`` (backed by ``pluginapi.SetAtomic``, which routes through master) gates the state transitions atomically. The cluster mutex is kept for cross-node Start/Cancel serialization.

**2. Migration ``000005`` idempotency.** ``CREATE INDEX`` (×3) in ``000005_create_user_agents_table.up.sql`` lacked ``IF NOT EXISTS``. If the migration partially completed and was retried, ``pq: relation "idx_useragents_creator" already exists (42P07)`` blocked plugin restart. Added ``IF NOT EXISTS`` to all three ``CREATE INDEX`` statements. Audited the rest of ``store/migrations/`` — every other migration is already guarded.

#### Ticket Link

N/A — internal bug report.

#### Release Note

```release-note
Fixed a race in the embedding-search reindex job that left the job wedged for up to 10 minutes after Cancel-then-Resume in HA deployments.
Made the user_agents migration index creation idempotent so a retried plugin upgrade does not fail with "relation idx_useragents_creator already exists".
```

## Summary by CodeRabbit

* **New Features**
  * UI now shows “Canceling…” with the cancel button disabled while cancellation is processed; polling continues to reflect final outcome and progress.

* **Bug Fixes**
  * More robust, atomic job lifecycle transitions to prevent race conditions, stale cancellations and cross-run interference.
  * Jobs now use per-run identifiers to avoid stale/duplicate operations and improve reclaiming of stale or orphaned runs.

* **Tests**
  * Expanded coverage for cancellation/resume races, CAS semantics and recoverability scenarios.

* **Chores**
  * Made database index creation idempotent.


nickmisasi (approved):
LGTM

PR #716, opened 2026-05-06
Add OpenAPI spec for Agents plugin APIs

nickmisasi:
#### Summary
Adds a canonical OpenAPI 3 spec for the Agents plugin HTTP API under `api/openapi.yaml`, covering the existing plugin routes served under `/plugins/mattermost-ai`.

Also adds `make openapi-validate` with route coverage checking so the spec can be validated independently of the plugin build setup.

Validation:
- `make openapi-validate`
- Core API merge validated in the companion Mattermost PR with `AGENTS_OPENAPI_SPEC` pointed at this local spec.

#### Ticket Link

N/A

#### Screenshots

N/A

#### Release Note
```release-note
Added OpenAPI documentation for the Agents plugin HTTP API.
```

## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive OpenAPI 3.0 specification documenting the plugin HTTP API surface, security schemes, endpoints, request/response schemas and data models.

* **Chores**
  * Added an OpenAPI validation target that validates the specification and enforces route coverage checks.
  * Validation now skips unrelated build steps when invoked.


crspeller (changes requested):
Can we automatically generate this? I don't mean via LLM but having one write the automation to automatically generate this. 
If we can we don't need to commit it to the repo verbatim like this.

nickmisasi:
I'm not sure if there's some javadoc-type way of documenting inline with comments and then having a script generate the file later. 

This is how [playbooks does it](https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/api/api.yaml) so I conformed to that path. Core Mattermost works the same way, it just splits the yaml up into different files rather than one large one. 

I would imagine this to only ever be touched by agents anyways, so doesn't really matter if the files are split up.

PR #731, opened 2026-05-08
Combine each LLM conversation turn into a single post

crspeller:
#### Summary

Previously, when an agent responded across tool-approval gates the plugin created a fresh post for each follow-up stream — an artifact of the old conversation-tracking system. This PR combines the entire LLM conversation turn (auto-run rounds + every user-approval continuation) into a single post and renders the back-and-forth as a sequence of rounds (Reasoning → Text → Tools per round).

**Backend**
- New `Store.UpdateTurnPostID` (with real-DB integration tests) so the streaming layer can demote a prior anchor turn before creating the new one — preserves the "exactly one assistant turn anchored per post" invariant the webapp's anchor lookup depends on.
- New `Store.DeleteResponseTurns` (with real-DB integration tests) — used by the regen flow to scrub the prior response (anchor + auto-run intermediates + tool_results) so the stream that follows is byte-for-byte equivalent to a first stream. No special update-in-place path.
- Split `streaming.StreamToPost` (first stream + regen) from `streaming.StreamContinuationToPost` (tool-approval resume). The two paths used to be auto-detected from "is there a prior anchor?", which mis-classified a regen of a continuation post as another continuation. The split makes the caller's intent explicit.
- `finalizeTurn` is now two cases:
  - fresh stream / regen → create anchor (the regen caller scrubbed prior turns first, so there's nothing to update)
  - continuation → demote prior anchor (`PostID = nil`), then create new anchor at the back of the conversation
- New `PostStreamingControlContinue = "continue"` control event so the webapp keeps resolved tool cards visible across the user-approval gap.
- `conversation.BuildCompletionRequest` with `ExcludeAfterPostID` now walks back from the anchor to the originating user turn and truncates after it. This stops bifrost models that disallow assistant prefill from rejecting the request when regenerating a tool-approval continuation post.
- `conversations.regenerateViaConversation` builds the completion request **before** scrubbing — `ExcludeAfterPostID` walks back from the anchor that `DeleteResponseTurns` is about to remove.
- `conversations.streamToolFollowUp` calls `streamContinuationToExistingPost` (the gated continuation path), no longer creates a new post.
- Removed the `next: ""` broadcast on resolved tool calls — the empty event used to clear the round's preamble before the webapp could attribute it to its round.

**Webapp**
- New `Round` type and `buildRoundsFromTurns(conversation, postId)` in `turn_content_utils.ts`, sharing extraction helpers (`buildToolResultMap`, `toolUseBlockToToolCall`) with `extractToolCallsForPost`.
- `llmbot_post.tsx` rebuilt around `persistedRounds + liveRounds + currentRound`. Resolved tool_call events snapshot the in-progress round into `liveRounds` (toolrunner persists tool turns mid-stream but the anchor isn't written until end-of-stream, so local state bridges the gap until the post-end refetch lands). A `pendingRefetch` flag with `useLayoutEffect` clears local state synchronously after the refetch so persisted and live copies don't double-render.
- `regenerating` flag with a dedicated render branch: while a regen is in flight, suppress the cached `persistedRounds` (still holds the pre-regen turn until refetch) and render `liveRounds + currentRound`, so the post doesn't visually empty out between rounds.
- `currentRound` is rendered whenever it has content. The earlier `generating && currentRound` guard hid the thinking block on first generate, because `reasoning_summary` events flip `generating` to false while reasoning chunks stream.
- Approval stage applies only when the last rendered round is also the last persisted round — the live round never inherits a cached `'call'` stage, so auto-approved tools no longer flash open with approve buttons.
- Accepts empty-string `next` updates so server-side message clears propagate properly.

**Tests**
- New `streaming/turn_persistence_test.go` cases: continuation demote-then-create ordering; multi-post cross-post isolation; table-driven continuation-detection guards; lookup-error fall-through; regen via `StreamToPost` creates a fresh anchor and never demotes; event-ordering invariant for the round preamble.
- New `store/turns_test.go: TestUpdateTurnPostID` (real-DB) covering clear / reassign / others-untouched / non-existent-id.
- New `store/turns_test.go: TestDeleteResponseTurns` (real-DB) covering anchor + intermediates removal, leaves earlier turns untouched, no-op when the post has no anchor.
- New `conversation/service_test.go: TestBuildCompletionRequest_ExcludeAfterPostID_ToolApprovalContinuationLeavesUserTail` regression for the bifrost prefill bug.
- New `turn_content_utils.test.ts: buildRoundsFromTurns` cases including the user-approval shape (tool_result at sequence > anchor), missing tool_result, demoted prior anchor, sibling-post isolation.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-66990

#### Screenshots

image

#### Release Note
```release-note
Multi-round agent responses (auto-run tool rounds and post-approval continuations) now land in a single post that renders the LLM's back-and-forth round by round, instead of creating a separate post per continuation.
```

## Release Notes

* **New Features**
  * Tool approval continuations now stream responses into the existing post instead of creating new posts, providing a more streamlined approval workflow.
  * Multi-round responses are now displayed with improved round-by-round organization within a single post.

* **Bug Fixes**
  * Fixed cache synchronization race condition when multiple conversation refreshes occur in quick succession, ensuring the latest data is always displayed.

* **Improvements**
  * Enhanced conversation state handling for tool approval workflows and improved error recovery during continuation operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mattermost/mattermost-plugin-agents/pull/731?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/api/api_no_tools_test.go
+++ b/api/api_no_tools_test.go
@@ -153,6 +153,10 @@ func (m *mockConvServiceStore) UpdateTurnPostID(_ string, _ *string) error {
 	return nil
 }
 
+func (m *mockConvServiceStore) DeleteResponseTurns(_ string, _ string) error {
+	return nil
+}
+
 func (m *mockConvServiceStore) GetMaxSequenceForConversation(conversationID string) (int, error) {
 	return len(m.turns[conversationID]), nil
 }
--- a/api/api_no_tools_test.go
+++ b/api/api_no_tools_test.go
@@ -142,6 +142,14 @@ func (m *mockConvServiceStore) UpdateTurnTokens(_ string, _, _ int64) error {
 	return nil
 }
 
+func (m *mockConvServiceStore) GetTurnByPostID(_ string) (*store.Turn, error) {
+	return nil, nil
+}
+
+func (m *mockConvServiceStore) UpdateTurnPostID(_ string, _ *string) error {
+	return nil
+}
+
 func (m *mockConvServiceStore) GetMaxSequenceForConversation(conversationID string) (int, error) {
 	return len(m.turns[conversationID]), nil
 }
--- a/api/api_no_tools_test.go
+++ b/api/api_no_tools_test.go
@@ -67,6 +67,9 @@ func (s *noToolsStreamingService) StreamToNewDM(_ context.Context, botID string,
 func (s *noToolsStreamingService) StreamToPost(context.Context, *llm.TextStreamResult, *model.Post, string, string) {
 }
 
+func (s *noToolsStreamingService) StreamContinuationToPost(context.Context, *llm.TextStreamResult, *model.Post, string, string) {
+}
+
 func (s *noToolsStreamingService) StopStreaming(string) {}
 
 func (s *noToolsStreamingService) GetStreamingContext(ctx context.Context, _ string) (context.Context, error) {
--- a/channels/analysis_conversation_test.go
+++ b/channels/analysis_conversation_test.go
@@ -132,6 +132,47 @@ func (s *inMemoryStore) UpdateTurnPostID(id string, postID *string) error {
 	return nil
 }
 
+func (s *inMemoryStore) DeleteResponseTurns(conversationID, postID string) error {
+	anchorSeq := -1
+	for _, id := range s.turnsByConv[conversationID] {
+		t, ok := s.turns[id]
+		if !ok {
+			continue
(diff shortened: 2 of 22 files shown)


nickmisasi (approved):
LGTM - but let's run a /deslop on the comments

PR #764, opened 2026-05-20
fix: 409 on first Full Reindex click for fresh installs

crspeller:
#### Summary

On a fresh install, the first click of "Full Reindex" in the system console returned 409 and could not recover. Regression from #689 (commit fb01e43f), present in v2.0.0–v2.0.3.

Root cause: `mmapi.KVGet` passed upstream `pluginapi.KV.Get`'s `(nil err, empty bytes)` reply through unchanged, so callers could not distinguish "missing" from "present-but-zero." `mmapi.IsKVNotFound` then matched the literal string `"not found"`, which the real pluginapi never returns for missing keys. The indexer's CAS ran with a marshaled zero `JobStatus` as `oldValue` against an empty KV store and lost — surfacing as "job already running" → 409.

Fix:
- `mmapi.KVGet` reads raw bytes and translates the empty-bytes reply into a new `ErrKVNotFound` sentinel. `IsKVNotFound` uses `errors.Is`.
- Updated 4 sites in `mcp/oauth_store.go` and 1 in `mcp/user_preferences.go` that silently depended on the old "(nil, empty struct)" behavior to preserve their documented "default on missing" semantics. `loadSession` deliberately propagates the error — its caller (`ProcessCallback`) dereferences the returned session unguarded, so a missing session must surface as `"invalid or expired session"`, not nil-panic.
- Swept indexer/api tests that mocked the wrapper with `errors.New("not found")` to use the sentinel directly.

Tests:
- New `mmapi/client_test.go` pins the wrapper contract (missing key, present-key unmarshal, bytes passthrough, unmarshal error preserves `%w` chain, underlying `*model.AppError` survives the wrapper).
- New `TestStartReindexJob_FreshInstall_AuthenticUpstreamContract` exercises the real `mmapi.Client` wrapping `plugintest.API` returning `(nil, nil)` (the authentic upstream signal previous test sites mocked over). Captures the CAS call and asserts the production invariant: fresh-install CAS must use an empty `OldValue`.
- `TestHandleGetJobStatusIncludesStale` extended to cover `404 {"status":"no_job"}` for fresh installs.
- `TestProcessCallback_InvalidSession` updated to use `mmapi.ErrKVNotFound` and assert wrap-chain preservation.

#### Ticket Link

Regression from #689 (commit fb01e43f). Affects v2.0.0–v2.0.3.

#### Test plan

- [x] `go test ./mmapi/ ./api/ ./indexer/ ./mcp/` — all green
- [ ] Deploy to a fresh local Mattermost (no prior reindex), click "Full Reindex" in the system console, confirm 200 + progress UI

#### Release Note
```release-note
Fixed a regression where the first click of "Full Reindex" in the system console returned an error on a fresh install and could not recover. Affects v2.0.0–v2.0.3.
```

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for key-value storage and fresh-install semantics, including new checks for missing-job and missing-key behaviours.

* **Bug Fixes**
  * Standardised handling of missing key data: returns sensible defaults (e.g. user preferences, token state) or a clear 404 {"status":"no_job"} for cancel/status endpoints on fresh installs, reducing spurious errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mattermost/mattermost-plugin-agents/pull/764?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/mcp/oauth_store.go
+++ b/mcp/oauth_store.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/mattermost/mattermost-plugin-agents/mmapi"
 	"golang.org/x/oauth2"
 )
 
@@ -50,6 +51,9 @@ func (m *OAuthManager) loadToken(userID, serverID string) (*oauth2.Token, error)
 	var oauth2Token oauth2.Token
 	err := m.pluginAPI.KVGet(tokenKey, &oauth2Token)
 	if err != nil {
+		if mmapi.IsKVNotFound(err) {
+			return nil, nil
+		}
 		return nil, fmt.Errorf("failed to retrieve token from KV store: %w", err)
 	}
 
@@ -84,6 +88,9 @@ func (m *OAuthManager) LoadAuthNeededState(userID, serverID string) (*OAuthNeede
 	var state OAuthNeededState
 	err := m.pluginAPI.KVGet(authNeededKey, &state)
 	if err != nil {
+		if mmapi.IsKVNotFound(err) {
+			return nil, nil
+		}
 		return nil, fmt.Errorf("failed to retrieve OAuth-needed state from KV store: %w", err)
 	}
 
@@ -155,6 +162,9 @@ func (m *OAuthManager) loadClientCredentials(serverURL string) (*ClientCredentia
 	var creds ClientCredentials
 	err := m.pluginAPI.KVGet(credKey, &creds)
 	if err != nil {
+		if mmapi.IsKVNotFound(err) {
+			return nil, nil
+		}
 		return nil, fmt.Errorf("failed to retrieve client credentials from KV store: %w", err)
 	}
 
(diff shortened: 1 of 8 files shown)


mcp/oauth_store.go
```diff
@@ -84,6 +88,9 @@ func (m *OAuthManager) LoadAuthNeededState(userID, serverID string) (*OAuthNeede
 	var state OAuthNeededState
 	err := m.pluginAPI.KVGet(authNeededKey, &state)
 	if err != nil {
+		if mmapi.IsKVNotFound(err) {
+			return nil, nil
+		}
```
nickmisasi, line 93: I think this + the rest of them will panic. For ex in api_mcp.go if err is nil, it will immediately try to check OAuthNeededState.AuthURL on line 156

nickmisasi (changes requested):

mcp/oauth_store.go
crspeller: This actually matches the behavior from before the change. I an not sure what nil you are referring too but I think all paths are safe.

nickmisasi: I missed that there was already a `nil` check here: https://github.com/mattermost/mattermost-plugin-agents/blob/e6c8334ee4173ffab6399e4beac4c8d34baf8670/api/api_mcp.go#L180 NVM!

nickmisasi (approved):

PR #768, opened 2026-05-22
fix: synthesize a final answer when toolrunner hits MaxToolRounds

BenCookie95:
#### Summary

The toolrunner caps tool-call rounds at `MaxToolRounds = 10` to keep runaway models bounded. When every round still requested tools, the runner executed the last round's tools and then **exited silently** without ever asking the model to synthesize. Callers received only the intermediate \"let me search…\" preambles streamed from each round, with no closing answer — this is the symptom reported by bridge users invoking the agents bridge from another plugin.

This PR fixes the bug by forcing a synthesis on the final round. At the end of round `MaxToolRounds-2`, the runner appends a new `ToolIterationLimitUserMessage` user post and adds `WithToolsDisabled()` to the option set, so the loop's normal final iteration becomes a tools-disabled synthesis. The \"no tool calls → final response\" early-return already in the loop then streams the synthesis text and emits `End`.

The fix lives in `toolrunner/`, so every caller benefits — bridge, conversations, channels, search, react, threads, etc.

#### Behavior change

- Before: 10 LLM calls max; runner emits `End` with no synthesis if the 10th round still had tools.
- After: still 10 LLM calls max. The 10th call is forced to be tools-disabled and seeded with a user message instructing the model to summarize from prior results, so the final response is always a synthesized answer rather than an abrupt cutoff.

#### Changes

**`toolrunner/`**
- At round `MaxToolRounds-2`, appends `ToolIterationLimitUserMessage` (a user-role post) and adds `WithToolsDisabled()` for the next iteration — same pattern as the `MaxConsecutiveToolCallFailures` handler a few lines above.
- Adds `FinalText string` to `ToolRunResult` — the assistant text from the round where the loop exited with no tool calls. Empty when the run ended with unresolved tool calls, a stream error, or a failed forced synthesis (preamble-only, tool calls dropped). Safe to read after the stream is consumed.
- `finalAssistantText` helper discards preamble text from a synthesis round where the provider misbehaved and still emitted tool calls (defense-in-depth, since `tool_choice="none"` may not be honored by all providers).

**`llm/`**
- `ToolIterationLimitUserMessage` constant and `EnsureToolIterationLimitUserMessage` (appends a user post, idempotent via `strings.Contains`).
- `EnsureToolRetryLimitSystemMessage` now delegates to an unexported `ensureSystemMessage` helper, which also uses `strings.Contains` for idempotency so embedding the message in a larger system prompt is detected correctly.

**`bifrost/`**
- When `ToolsDisabled` is set but the conversation history contains `tool_use` blocks, providers like Anthropic require the tools array to remain present in the request to validate the schema. `convertTools` / `convertToResponsesTools` now keep the tools array in that case and rely on `tool_choice="none"` (set by `convertToBifrostRequest` / `convertToBifrostResponsesRequest`) to forbid further calls.
- `hasToolUseHistory` helper scans posts for non-empty `ToolUse` slices.

**`api/api_llm_bridge.go`**
- Non-streaming bridge response now reads `runResult.FinalText` directly instead of accumulating text from the stream.
- `drainToolRunnerStream` helper drains the stream for error detection only (text events are ignored).

#### Tests

- Updated `TestToolRunner_MaxRoundsExhausted`: asserts synthesized text is forwarded, `inner.callCount == MaxToolRounds`, and `MaxToolRounds-1` tool turns are recorded.
- New `TestToolRunner_MaxRoundsExhausted_SynthesisCallHasToolsDisabled`: verifies the final call has `WithToolsDisabled` and the request posts contain the iteration-limit user message; earlier calls do not have tools disabled.
- New `TestToolRunner_MaxRoundsExhausted_ProviderEmitsToolCallDuringSynthesis`: provider returns tool calls on the forced synthesis round — runner drops them, emits `End`, and `FinalText` is empty.
- New `TestToolRunner_FinalText_OmitsToolRoundPreamble` and `TestToolRunner_FinalText_DropsFailedSynthesisPreamble`: cover correct population of `FinalText`.
- New `TestFinalAssistantText`: unit tests for the `finalAssistantText` helper.
- New `TestEnsureToolIterationLimitUserMessage`: table-driven tests covering append / idempotent / embedded cases.
- Updated `TestEnsureToolRetryLimitSystemMessage`: adds case for message embedded in a larger system prompt.
- New `TestConvertToBifrostRequestToolChoiceNone` and `TestConvertToBifrostResponsesRequestToolChoiceNone`: verify `tool_choice="none"` is set when tools-disabled + history, and tools are omitted (no override) when tools-disabled without history.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68901

#### Release Note
```release-note
Fixed an issue where the agents bridge (and other tool-calling flows) would silently truncate the assistant response when the maximum number of tool-use iterations was reached. The runner now produces a final synthesized answer instead.
```

### DIFF
--- a/api/api_llm_bridge.go
+++ b/api/api_llm_bridge.go
@@ -519,7 +519,7 @@ func (a *API) streamLLMResponse(c *gin.Context, bot *bots.Bot, llmRequest llm.Co
 	var err error
 	if shouldExecute != nil {
 		var runResult *toolrunner.ToolRunResult
-		runResult, err = toolrunner.New(bot.LLM()).WithLogger(&a.pluginAPI.Log).Run(c.Request.Context(), llmRequest, shouldExecute, nil, opts...)
+		runResult, err = toolrunner.New(bot.LLM()).Run(c.Request.Context(), llmRequest, shouldExecute, nil, opts...)
 		if runResult != nil {
 			streamResult = runResult.Stream
 		}
@@ -578,7 +578,7 @@ func (a *API) handleNonStreamingLLMResponse(c *gin.Context, bot *bots.Bot, llmRe
 		return
 	}
 
-	runResult, err := toolrunner.New(bot.LLM()).WithLogger(&a.pluginAPI.Log).Run(c.Request.Context(), llmRequest, shouldExecute, nil, opts...)
+	runResult, err := toolrunner.New(bot.LLM()).Run(c.Request.Context(), llmRequest, shouldExecute, nil, opts...)
 	if err != nil {
 		c.JSON(http.StatusInternalServerError, bridgeclient.ErrorResponse{
 			Error: fmt.Sprintf("failed to complete LLM request: %v", err),
--- a/api/api_llm_bridge.go
+++ b/api/api_llm_bridge.go
@@ -506,19 +506,25 @@ func (a *API) checkBridgePermissions(userID, channelID string, bot *bots.Bot) er
 }
 
 func drainToolRunnerStream(stream *llm.TextStreamResult) error {
+	var firstErr error
 	for event := range stream.Stream {
 		if event.Type != llm.EventTypeError {
 			continue
 		}
+		if firstErr != nil {
+			continue
+		}
 		if e, ok := event.Value.(error); ok {
-			return e
+			firstErr = e
+			continue
 		}
 		if s, ok := event.Value.(string); ok {
-			return errors.New(s)
+			firstErr = errors.New(s)
+			continue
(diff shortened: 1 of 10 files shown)


BenCookie95:
@coderabbitai full review

BenCookie95:
@coderabbitai review

nickmisasi (changes requested):
Can you clean up the overly verbose LLM comments?

nickmisasi (approved):

PR #771, opened 2026-05-23
dynamic mcp: add persisted loading state

nickmisasi:
#### Summary
Adds the config and persistence foundation for dynamic MCP tool loading: the `MCPDynamicToolLoading` bot/agent setting (default-on for new, migrated, and legacy agents), a database migration, metrics hooks for dynamic tool events, and per-tool `RetrievalDescriptionOverride` in MCP config. Removes unused EnableLLMTrace plumbing and an unshipped loaded-tool store that was superseded by turn-based derivation on later floors.

**Stack:** 1/6
- ↑ [772 — dynamic-mcp-02-client-catalog](https://github.com/mattermost/mattermost-plugin-agents/pull/772)

#### Ticket Link

N/A

#### Screenshots

N/A

#### Release Note

```release-note
Added MCP dynamic tool loading agent setting (defaults to on), database migration, metrics, and retrieval description override configuration.
```

## Release Notes

* **New Features**
  * MCP dynamic tool loading is now enabled by default for agents, improving tool availability and performance.
  * Added support for retrieval description overrides on MCP tools for customized tool presentation.
  * New metrics tracking for MCP dynamic tool events, including bot name, event type, and result status.

* **Tests**
  * Enhanced test coverage for MCP configuration, tool loading defaults, and agent persistence.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mattermost/mattermost-plugin-agents/pull/771?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


store/migrations/000009_create_loaded_mcp_tools.up.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS LLM_LoadedMCPTools (
```
crspeller: This seems to be maybe over-complicated? Is there some reason we would need to query this independently or would a JSON field on the conversations table be enough? 

Or maybe even simpler and DRY would be to look though the conversation history and derive this from successful calls of the load tool? That way we don't store anything more.

crspeller (changes requested):

store/migrations/000009_create_loaded_mcp_tools.up.sql
nickmisasi: >This seems to be maybe over-complicated? Is there some reason we would need to query this independently or would a JSON field on the conversations table be enough?

I agree this is slightly more complicated than JSON, but I think from a performance perspective the db route makes the most sense. We're opening the door for a lot more tool calling to happen with this feature, and if we go the JSON route every tool call (ie, every tool loaded) will require a complete rewrite of the JSON to the db. JSON write size grows linearly based on how many tools are _already_ loaded - if you continue to load more tools, you're continuing to rewrite the JSON payload, so the amount of data written grows quadratically

Extrapolated for 500 tools (which is a lot now, but not unrealistic some day), say at around 200B per tool - with db it's ~100KB of writes. JSON would be closer to 31MB 

Another thing - while right now our tool calls are sequential, I don't think it's unrealistic to want to be able to concurrently execute tools in the future, and the db approach is immediately safe for that too.

>Or maybe even simpler and DRY would be to look though the conversation history and derive this from successful calls of the load tool? That way we don't store anything more.

There's some similar perf issues with this - every rebuild needs to scan ALL previous turns' JSONB, and would be problematic if we ever needed to make changes because we're coupling the runtime behaviour to the existing JSON-shape

crspeller: > concurrently execute tools in the future

I think if instead of having the tool execution itself write the DB entry and have it batched after a tool call round this isn't relevant.

Also I think doing that would make the performance issue less as well because its one write per LLM turn which I doubt is particularly important in an LLM turn. 

> There's some similar perf issues with this - every rebuild needs to scan ALL previous turns' JSONB

Given the issues you pointed out with the JSON one I am leaning towards this approach. I don't think the cost of the scan is particularly relevant since we are reading all of this anyway to build the prompt on each LLM call. It's DRY which is relevant for potential regeneration or possible branching / truncation / compaction. Not sure if your future PRs handle that well or not. Its probably even more performant because we skip an extra DB call.

nickmisasi: > I don't think the cost of the scan is particularly relevant since we are reading all of this anyway to build the prompt on each LLM call.

The data is already fetched, but the _parse-and-walk_ cost is new, running on every message. The problem is we can't just take the fact that load_tool was called. Let's take the 500 tool calls scenario. You'd need to:
1. Walk all 500+ turns and parse `[]ContentBlock` (it would really be more like 1000+ turns because of search -> load) - we're already doing this so no big deal
2. For each `tool_use` block where `Name == "load_tool"` you then have to look up the matching `tool_result` block by its `ToolCallID` - a `tool_use` block for a particular tool name only tells you that the LLM _asked to run a tool_, but doesn't tell you that that tool was actually loaded (for ex - it hallucinated a tool name)
3. Then you have to parse the tool_result.Result and check for Loaded: true - can't rely on it returning information as it could have just returned an error
4. Deduplication has to happen every time here, because you don't have an upsert
5. Remove stale tools - if jira was turned off, or a singular tool is no longer available, the db path removes it as soon as its found to be stale. Derivation means it needs to be filtered out on every turn. 

Steps 1-4 are work that the db persistence approach does once, at the time the tool is loaded. Derivation would do them every turn, forever, growing with the conversation's history.

I think long term, especially now that we have access to the database, we should be leaning more on proper schema's and less on JSON parsing. I don't see JSON derivation as a reliable way to scale to the numbers of users we hope to support

crspeller: 1. Like you said we already do this. 
2. This is already done for us in BlocksToPost and the check is handled by step 3
3. This is the actual new work. It's this vs another DB call. I suspect at any reasonable scale this lookup is going to win. 
4. Correct, but we are probably shoving this into a map anyway so a map insert here likely won't matter. 
5. Isn't the DB behavior a bug in itself? What if I remove JIRA then re-add it? If that happens the model will think, given the conversation, that it has loaded the tool but it won't actually be loaded. Potentially confusing? 

There is also the correctness issue keeping two sets of truth here. Having a look at the regeneration flow in future PRs it looks like that is not currently handled correctly. We would need a solution for that.

nickmisasi: I have adopted your suggestion for derivation throughout the stack. There are some downstream flaky pipelines but overall should be ready for another review.

config/config.go
```diff
@@ -37,6 +37,7 @@ type Config struct {
 	WebSearch                       WebSearchConfig                  `json:"webSearch"`
 	TelemetryOutput                 string                           `json:"telemetryOutput"`
 	OpenTelemetryEndpoint           string                           `json:"openTelemetryEndpoint"`
+	EnableLLMTrace                  bool                             `json:"enableLLMTrace"`
```
crspeller: Why are we resurrecting this setting?

crspeller (changes requested):

config/config.go
nickmisasi: It got brought back from merge conflict resolution. At first glance I thought it was unintentionally removed from the db config migration so I left it. Forgot you had removed it in the otel pr

crspeller (approved):

PR #772, opened 2026-05-23
dynamic mcp: harden client catalog discovery

nickmisasi:
#### Summary
Hardens MCP client catalog discovery with safer tool identity handling (duplicate registrations are no longer silently dropped), plugin/embedded discovery refresh behavior, user-client synchronization, and retrieval override plumbing, with backwards-compatible tool policy lookup for configs without namespace plumbing. Also adds bare MCP tool name resolution, LLM context scaffolding for dynamic loading (request context, registry stash, telemetry hooks, preloaded tools), threads request context through conversation and API entry points, fixes channel analysis tool discoverability under narrow allowlists, and removes legacy log-based tool-trace plumbing superseded by OpenTelemetry.

**Stack:** 2/6
- ↓ [771 — dynamic-mcp-01-config-store](https://github.com/mattermost/mattermost-plugin-agents/pull/771)
- ↑ [773 — dynamic-mcp-03-registry-meta-tools](https://github.com/mattermost/mattermost-plugin-agents/pull/773)

#### Ticket Link

N/A

#### Screenshots

N/A

#### Release Note

```release-note
Improved MCP tool catalog discovery, bare tool name resolution, and channel analysis tool availability for dynamic loading.
```

## Summary by CodeRabbit

* **New Features**
  * Runtime tracking and telemetry for dynamic MCP tool search/load and per-conversation IDs.
  * Namespacing and bare-name helpers plus retrieval-override keys to disambiguate tools across servers.
  * Per-user plugin/embedded MCP clients with optional shared tools cache.

* **Improvements**
  * Request-context propagation across tool resolution and MCP flows for cancellation/timeouts.
  * Thread-safe client/tool caching, normalized server-origin handling and safer reconnection/refresh behavior.

* **Tests**
  * Expanded unit, integration and e2e coverage for namespaced tools, caching, discovery and telemetry.

### DIFF
--- a/llm/context.go
+++ b/llm/context.go
@@ -20,7 +20,7 @@ type ToolInfo struct {
 	ServerOrigin string
 }
 
-// Context represents the data necessary to build the context of the LLM.
+// Context represents the per-turn data necessary to build the context of the LLM.
 // For consumers none of the fields can be assumed to be present.
 type Context struct {
 	// Server
@@ -56,6 +56,13 @@ type Context struct {
 	DisabledToolsInfo []ToolInfo // Info about tools that are unavailable in the current context (e.g., DM-only tools in a channel)
 	Parameters        map[string]interface{}
 
+	// ToolRuntime holds non-prompt tool execution state for this turn.
+	ToolRuntime ToolRuntimeContext
+}
+
+// ToolRuntimeContext holds request-scoped tool runtime state that should not be
+// rendered into the prompt.
+type ToolRuntimeContext struct {
 	// MCPDynamicToolLoading indicates this context uses strict MCP dynamic loading.
 	MCPDynamicToolLoading bool
 	// MCPDynamicToolTelemetry receives low-cardinality dynamic MCP tool events.
@@ -77,16 +84,7 @@ type Context struct {
 	// catalog and are request scoped.
 	PreloadedMCPTools []EnabledMCPTool
 
-	// MCPToolRegistry holds the strict MCP tool registry that was built
-	// alongside Tools, when MCP dynamic tool loading is enabled. It is stashed
-	// here so callers can replay loaded-tool restoration after the conversation
-	// row exists without rebuilding the entire tool store.
-	//
-	// Stored as `any` to avoid an llm -> mcp import cycle: the mcp package
-	// already imports llm, and the only consumer that needs the concrete type
-	// is the llmcontext package, which can import both. Type-assert to
-	// *mcp.ToolRegistry there.
-	MCPToolRegistry any
+	restoreMCPDynamicTools func(names []string)
 }
 
 type MCPDynamicToolTelemetry interface {
@@ -144,7 +142,7 @@ func (c *Context) CustomPromptVars() map[string]string {
 }
 
 func (c *Context) ObserveMCPDynamicToolEvent(event, result string) {
-	if c == nil || c.MCPDynamicToolTelemetry == nil {
+	if c == nil {
 		return
 	}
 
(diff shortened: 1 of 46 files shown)


nickmisasi:
@coderabbit review

llm/context.go
```diff
@@ -47,6 +55,42 @@ type Context struct {
...
+	// MCPToolRegistry holds the strict MCP tool registry that was built
+	// alongside Tools, when MCP dynamic tool loading is enabled. It is stashed
+	// here so callers can replay loaded-tool restoration after the conversation
+	// row exists without rebuilding the entire tool store.
+	//
+	// Stored as `any` to avoid an llm -> mcp import cycle: the mcp package
+	// already imports llm, and the only consumer that needs the concrete type
+	// is the llmcontext package, which can import both. Type-assert to
+	// *mcp.ToolRegistry there.
+	MCPToolRegistry any
```
crspeller: These don't belong here. This struct is for prompt building not passing stuff like this.

llm/context.go
```diff
@@ -36,6 +37,13 @@ type Context struct {
 	// User that is making the request
 	RequestingUser *model.User
 
+	// RequestContext carries the caller's request-scoped context for downstream
+	// work such as MCP tool discovery. May be nil in tests.
+	RequestContext stdcontext.Context
```
crspeller: Doesn't belong here, we should update to pass context as first parameter if needed.

llm/context.go
crspeller: Specifically for this one I don't think we want `any`

mcp/user_clients.go
```diff
@@ -287,6 +372,77 @@ func (c *UserClients) createToolResolver(client *Client, toolName string) func(l
...
+func mcpServerSlug(serverID string, client *Client) string {
+	if client != nil && (client.config.BaseURL == EmbeddedClientKey || client.config.Name == EmbeddedClientKey || serverID == EmbeddedClientKey) {
+		return "mattermost"
+	}
+
+	candidates := []string{}
+	if client != nil {
+		candidates = append(candidates, client.config.Name)
+	}
+	candidates = append(candidates, serverID)
```
crspeller, line 370: Not sure the opeque serverID makes sense over the base URL. These are LLM visible so ideally they are not something strange. 

There is also the reported name from the MCP server we could use.

llm/tools.go
```diff
@@ -350,16 +353,45 @@ func NewToolStore() *ToolStore {
 func (s *ToolStore) AddTools(tools []Tool) {
 	for _, tool := range tools {
 		s.tools[tool.Name] = tool
+		if s.unloadedMCPTools != nil {
```
crspeller: Unnecessary check.

llm/tools.go
```diff
@@ -318,8 +320,9 @@ type ToolAuthError struct {
 }
 
 type ToolStore struct {
-	tools      map[string]Tool
-	authErrors []ToolAuthError
+	tools            map[string]Tool
+	unloadedMCPTools map[string]ToolInfo
```
crspeller, line 324: Seems more complicated than nessisary? Why have toolinfo? Also why is it unloaded MCP tools and not what is loaded?

mcp/client.go
```diff
@@ -46,22 +49,30 @@ type EmbeddedMCPServer interface {
...
-	sessionID      string                // session ID for embedded server reconnection
+	session         *mcp.ClientSession
+	config          ServerConfig
+	toolsMu         sync.RWMutex
+	discoveryMu     sync.Mutex
+	tools           map[string]*mcp.Tool
+	toolsDirty      bool
+	toolsGeneration uint64
+	notifyOwnerMu   sync.RWMutex
+	notifyOwner     *Client
```
crspeller: Feels like we should find a better solution for this.

mcp/client.go
```diff
@@ -349,7 +485,11 @@ func (c *Client) createSession(ctx context.Context, serverConfig ServerConfig) (
 			Name:    "mattermost-agents",
 			Version: "1.0",
 		},
-		&mcp.ClientOptions{},
+		&mcp.ClientOptions{
+			ToolListChangedHandler: func(ctx context.Context, _ *mcp.ToolListChangedRequest) {
```
crspeller: Is this for supporting the MCP feature where an MCP server can add tools while holding the connection open? This feels orthogonal to what this PR is trying to accomplish. Maybe a separate PR? Not sure it's even worth supporting.

crspeller (changes requested):

llm/tools.go
nickmisasi: unloadedMCPTools are tools we know to exist in the catalog (ie, from a search) but haven't been loaded yet. ToolInfo is a lightweight struct to reflect what we actually know about it at this point. I've added a comment to clarify

mcp/user_clients.go
nickmisasi: Remote servers: require/configure a name through system console; serverID is that name anyway.
Plugin MCP with a configured name: client.config.Name is the plugin-provided name, so that wins.
Plugin MCP without a configured name: client.config.Name is empty, then fallback uses serverID, which is plugin://<pluginID>.
For that fallback case, serverID and BaseURL are basically the same synthetic origin, so there isn't much of a difference in ordering. I preserved both fallbacks more for safety than anything.

mcp/client.go
nickmisasi: This was [in the original spec](https://mattermost.atlassian.net/wiki/spaces/agents/pages/4526866481/Dynamic+MCP+Tools#Tier-1-%E2%80%94-Foundational-correctness) 

But I think we can kick support for this down the road now that we have https://github.com/mattermost/mattermost-plugin-agents/pull/732 opened, at least end users will be able to refresh. 

Removed in next commit

llm/context.go
nickmisasi: I cleaned this up by moving the dynamic MCP/tool execution fields into a nested ToolRuntimeContext on llm.Context, so they’re no longer loose prompt-building fields.

I think this should stay reachable from llm.Context because Context is already the per-turn carrier used by tool construction and execution (Tools already lives there), not just rendered prompt text. The dynamic MCP state needs to be available across the same flow: building the tool store, resolving meta-tools, recording per-turn load/search telemetry, and restoring loaded tools. Splitting it into a separate top-level parameter would force the same runtime object through all resolver/toolrunner/context-builder paths without changing the ownership model. If you're passing it all around to the same places alongside the Context it's looking like a shared tool context anyways.

mcp/client.go
nickmisasi: Addressed in https://github.com/mattermost/mattermost-plugin-agents/pull/772#discussion_r3312927764

llm/context.go
crspeller: Ok makes sense. I think this could use a refactor to make that clear, but out of scope here.

llm/context.go
crspeller: You thumbs upped but it is still there.

crspeller (changes requested):
Just need to remove that context from the struct and I think this is good.

llm/context.go
nickmisasi: Should be good now. Did refactor the first time but missed a few spots

crspeller (approved):

PR #773, opened 2026-05-23
dynamic mcp: add registry and loading tools

nickmisasi:
#### Summary
Adds the dynamic MCP tool registry, BM25 search, and the `search_tools` / `load_tool` meta-tools used to discover and materialize MCP tools on demand. Centralizes tool loading on a single `ToolStore.LoadMCPTools` path (used by `load_tool` and, in later floors, cross-turn restoration), refactors `load_tool` to emit a canonical turn-derivable JSON envelope (replacing in-process loaded-tool recording), and guards zero-value registry searches.

**Stack:** 3/6
- ↓ [772 — dynamic-mcp-02-client-catalog](https://github.com/mattermost/mattermost-plugin-agents/pull/772)
- ↑ [774 — dynamic-mcp-04-context-toolrunner](https://github.com/mattermost/mattermost-plugin-agents/pull/774)

#### Ticket Link

N/A

#### Screenshots

N/A

#### Release Note

```release-note
Added dynamic MCP search and load tools for on-demand MCP tool discovery.
```

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Relevance-ranked search across registered tools (BM25) with deterministic tie-breaking and result limits.
  * In-memory tool registry with name/summary search and a "closest matches" fallback for partial hits.
  * Two meta-operations exposed: search available tools by keyword and load a specific tool at runtime.

* **Chores**
  * Removed legacy per-request MCP tool restorer API; dynamic load tracking simplified.

PR #774, opened 2026-05-23
dynamic mcp: enforce loaded tool execution

nickmisasi:
#### Summary
Builds strict MCP tool registries in `llmcontext`, enforces loaded-vs-unloaded tool visibility in toolrunner (rejecting calls to unloaded MCP tools), adds retry guidance for unloaded tools, and serializes loaded-tool state in conversation content blocks. Keeps meta-tool loading internal to the visible catalog, preloads channel-analysis tools before context build, materializes unloaded tools through the shared `ToolStore.LoadMCPTools` path (no context-level restorer hook), and includes e2e adjustments for bare tool names plus multiplayer opt-out from dynamic loading.

Wires dynamic MCP loading through conversation, approval, regeneration, and channel-analysis flows. Derives loaded MCP tools from retained turn history, binds channel-analysis tools as bare aliases, updates bot channel tool filtering and the agent API config surface, and adds prompt guidance for the dynamic search/load workflow when enabled.

**Stack:** 4/6
- ↓ [773 — dynamic-mcp-03-registry-meta-tools](https://github.com/mattermost/mattermost-plugin-agents/pull/773)
- ↑ [776 — UI and E2E wiring](https://github.com/mattermost/mattermost-plugin-agents/pull/776)

#### Ticket Link

N/A

#### Screenshots

N/A

#### Release Note

```release-note
Improved dynamic MCP tool execution safety by requiring tools to be loaded before use.
```

## Summary by CodeRabbit

* **New Features**
  * Agents can opt into configurable dynamic MCP tool loading; meta-tools (search/load) auto-run without manual approval.
  * Conversation restoration of previously loaded tools so approved tools can run using prior load state.

* **Improvements**
  * Better tool identity/display using bare names and server-origin normalization.
  * More detailed errors for unavailable/unloaded/unknown tools; channel analysis preloads required tools.
  * Approval metadata is redacted for non-requesters; prompts conditionally show dynamic-tool workflow.


nickmisasi:
@coderabbitai review

conversation/service.go
```diff
@@ -478,7 +478,17 @@ func turnsToLLMPosts(
...
+		post := BlocksToPost(blocks, turn.Role, redactUnshared, mmClient, enableVision, maxFileSize)
+		if turn.Role == "assistant" {
+			// Anthropic signed thinking must be replayed byte-for-byte. Our stored
+			// content blocks intentionally normalize assistant output (merge text
+			// blocks, pair tool_use/result turns, redact private content), so the
+			// persisted thinking block is not safe to send back as provider history.
+			// Keep reasoning for UI/persistence, but omit it from rebuilt requests.
+			post.Reasoning = ""
+			post.ReasoningSignature = ""
+		}
```
nickmisasi, line 515: This is needed until:

https://mattermost.atlassian.net/browse/MM-68977

api/api_llm_bridge.go
```diff
@@ -164,6 +164,15 @@ func (a *API) buildLLMBridgeContext(bot *bots.Bot, req bridgeclient.CompletionRe
 	return context, nil
 }
 
+func bridgeToolCatalogBot(bot *bots.Bot) *bots.Bot {
```
crspeller: Maybe not for this PR, but this needs to be cleaned up.

conversation/content_block.go
```diff
@@ -42,12 +42,15 @@ type ContentBlock struct {
...
-	Input        json.RawMessage `json:"input,omitempty"`
-	Status       string          `json:"status,omitempty"`
-	Shared       *bool           `json:"shared,omitempty"` // pointer to distinguish unset from false
+	ID              string          `json:"id,omitempty"`
+	Name            string          `json:"name,omitempty"`
+	ServerOrigin    string          `json:"server_origin,omitempty"`
+	Input           json.RawMessage `json:"input,omitempty"`
+	InputSchema     json.RawMessage `json:"input_schema,omitempty"`
+	MCPBareName     string          `json:"mcp_bare_name,omitempty"`
+	ToolDescription string          `json:"tool_description,omitempty"`
```
crspeller: Do we need these fields? I don't see them read anywhere except MCPBareName in 775 which appears to be a test that can never fail.

llmcontext/llm_context.go
```diff
@@ -231,9 +297,242 @@ func (b *Builder) getToolsStoreForUser(ctx stdcontext.Context, bot *bots.Bot, us
...
+		}
+		markUnloadedMCPTools(store, recomputed)
+	})
+}
+
+// AttachConversationID late-binds a ConversationID onto an already-built
+// llm.Context. Tool restoration is invoked separately via
+// (*llm.Context).RestoreMCPDynamicTools by callers that derive loaded
+// tool names from retained turns.
+func (b *Builder) AttachConversationID(c *llm.Context, bot *bots.Bot, conversationID string) {
```
crspeller: What is happening here? `b` is unused and so is `bot`? 
This feels like some architectural issues. Why is this function needed?

llmcontext/llm_context.go
```diff
@@ -231,9 +297,242 @@ func (b *Builder) getToolsStoreForUser(ctx stdcontext.Context, bot *bots.Bot, us
...
+	filtered := make([]llm.Tool, 0, len(tools))
+	for _, tool := range tools {
+		if keep(tool) {
+			filtered = append(filtered, tool)
+		}
+	}
+	return filtered
+}
+
+func retainOnlyAllowedMCPTools(tools []llm.Tool, allowlist []llm.EnabledMCPTool) []llm.Tool {
```
crspeller: Isn't this very similar `FilterMCPToolsByAllowlist` from the previous PR? a comment even names that one. Why is this here?

crspeller (changes requested):
Skipped over some of the code because I think some of my comments will chain to additional simplifications, but it just feels like it should be simpler to handle this. Are there maybe some architectural changes we can make to simplify this implementation?

api/api_llm_bridge.go
nickmisasi: Added a comment

conversation/content_block.go
nickmisasi: MCPBearName is needed but I removed the others. This made more sense when we were explicitly persisting but I agree is unnecessary with derivation - now rebuilding from the tool store, which I think has the added advantage of being able to handle tool schema changes automatically

llmcontext/llm_context.go
nickmisasi: Moved to context.go and rejigged to SetConversationID

conversations/mcp_preferences.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
```
nickmisasi: This file is a temporary shim to keep CI green until #775 when conversation wiring is included. Its used to prevent the agent loading tools that are disabled by the user preferences (via tools dropdown toggles). The next PR introduces a single pass context builder (buildConversationContextWithTools) over the conversation that every entrypoint to withUserDisabledMCPServerOptions funnels through instead passing to one of 2 helpers. 

This came about from a now-deleted Dry Run Security automated flag where tool loading could bypass user preferences.

api/api_channel.go
```diff
@@ -109,29 +116,16 @@ func (a *API) handleChannelAnalysis(c *gin.Context) {
...
-	// Check if read_channel tool is available
-	availableTools := llmContext.Tools.GetTools()
-	var toolNames []string
-	for _, tool := range availableTools {
-		toolNames = append(toolNames, tool.Name)
-	}
-
-	if llmContext.Tools.GetTool("read_channel") == nil {
-		a.pluginAPI.Log.Error("Channel analysis failed: read_channel tool not available",
+	// Validate that required tools are available for channel analysis.
```
crspeller, line 119: We don't allow the MCP server to be disabled any more so isn't this (and related helpers) all dead code?

mcp/client_manager.go
```diff
@@ -182,6 +182,10 @@ func (m *ClientManager) getClientForUser(ctx context.Context, userID string) (*U
 
 // GetToolsForUser returns the tools available for a specific user, connecting to embedded server if session ID provided.
 func (m *ClientManager) GetToolsForUser(ctx context.Context, userID string) ([]llm.Tool, *Errors) {
+	if ctx == nil {
```
crspeller: Generally nil context checks are not a thing. Why?

toolrunner/toolrunner.go
```diff
@@ -222,6 +223,24 @@ func (r *ToolRunner) runLoop(
 			return
 		}
 
+		store := toolStoreFromRequest(request)
+		if containsUnavailableTools(toolCalls, store) {
+			output <- llm.TextStreamEvent{Type: llm.EventTypeToolCalls, Value: toolCalls}
```
crspeller: Do we need to output anything to the stream here? The most this will do is cause the tools that we are rejecting to appear for a second.

toolrunner/toolrunner.go
```diff
@@ -341,15 +345,174 @@ func (r *ToolRunner) executeTools(ctx context.Context, toolCalls []llm.ToolCall,
...
+		}
+		enriched[i].Schema = tool.Schema
+		if enriched[i].ServerOrigin != "" {
+			enriched[i].MCPBareName = lookup.BareName
+		}
+	}
+	return enriched
+}
+
+func appendToolTurnAndPost(
```
crspeller, line 478: This is a very awkward helper because it mutates it's arguments and returns a value.

toolrunner/toolrunner.go
```diff
@@ -299,16 +295,26 @@ func (r *ToolRunner) executeTools(ctx context.Context, toolCalls []llm.ToolCall,
...
 		var result string
 		var resolveErr error
-		if request.Context != nil && request.Context.Tools != nil {
+		switch {
+		case request.Context == nil || request.Context.Tools == nil:
+			resolveErr = fmt.Errorf("no tool store available")
+		case request.Context.Tools.IsUnloadedMCPTool(tc.Name):
+			resolveErr = fmt.Errorf("%s", mcp.UnloadedMCPToolUserHint(tc.Name))
+		default:
+			lookup, ok := request.Context.Tools.LookupTool(tc.Name, tc.ServerOrigin)
```
crspeller, line 333: Why are we checking this again? If we get here we passed the `containsUnavailableTools` check.

conversations/regeneration.go
```diff
@@ -250,19 +249,23 @@ func (c *Conversations) regenerateViaConversation(
 		return nil, fmt.Errorf("failed to get conversation for regen: %w", err)
 	}
 
-	contextOpts := []llm.ContextOption{
-		c.contextBuilder.WithLLMContextDefaultTools(ctx, bot),
+	var contextOpts []llm.ContextOption
+	if channel.Type == model.ChannelTypeDirect || channel.Type == model.ChannelTypeGroup {
```
crspeller: Should this also be using `buildConversationContextWithTools`?

toolrunner/toolrunner.go
```diff
@@ -341,15 +345,174 @@ func (r *ToolRunner) executeTools(ctx context.Context, toolCalls []llm.ToolCall,
...
+		if mcp.IsMCPMetaTool(toolName) {
+			continue
+		}
+		if llmContext.ShouldRecordMCPDynamicSearchLoadCallSuccess(toolName) {
+			llmContext.ObserveMCPDynamicToolEvent("search_load_call_success", "success")
+		}
+	}
+}
+
+func enrichToolCallsForApproval(toolCalls []llm.ToolCall, store *llm.ToolStore) []llm.ToolCall {
```
crspeller, line 469: This and enrichToolCallFromStore look similar.

api/api_channel.go
```diff
@@ -188,6 +182,31 @@ func channelAnalysisToolBot(bot *bots.Bot) *bots.Bot {
...
+			missing = append(missing, required.ToolName)
+		}
+	}
+
+	return available, missing
+}
+
+// hasRequiredChannelAnalysisTool expects channel-analysis MCP tools to be
+// preloaded into the visible tool store by WithLLMContextPreloadedMCPTools.
+func hasRequiredChannelAnalysisTool(store *llm.ToolStore, required llm.EnabledMCPTool) bool {
```
crspeller: Do we really need this helper?

conversations/conversations.go
```diff
@@ -225,6 +226,13 @@ func (c *Conversations) ProcessDMRequest(
 	return &DMStreamResult{Stream: stream}, nil
 }
 
+func llmToolsFromContext(llmCtx *llm.Context) *llm.ToolStore {
```
crspeller: Just a nil guard, really needed?

llm/context.go
```diff
@@ -111,6 +117,14 @@ func (c *Context) SetBotFields(displayName, username, userID, defaultModel, serv
 	c.CustomInstructions = customInstructions
 }
 
+// SetConversationID late-binds a conversation ID onto an existing context.
+func (c *Context) SetConversationID(conversationID string) {
```
crspeller: This seems like a mess. Why is this here? 
Also I don't actually see where `conversationID` is used. Why is it there?

crspeller (changes requested):

api/api_channel.go
nickmisasi: Tool policy allows for disabling individual tools

toolrunner/toolrunner.go
nickmisasi: LookupTool is required to give us RuntimeName (a product of the barename with deduplication) for execution. You're right the lookup is more or less redundant but if you remove the ok check a genuinely missing tool for whatever reason would fall through to a generic ResolveTool error instead of a cleaner "unknown tool xyz" one

mcp/client_manager.go
nickmisasi: overly safe, removed

toolrunner/toolrunner.go
nickmisasi: No you're right we probably don't need this

toolrunner/toolrunner.go
nickmisasi: They are similar in signature but different in function, I did centralize some of the logic

mcp/client_manager.go
nickmisasi: Missed in initial commit, it'll be removed in the next one while it's syncing to the stack)

toolrunner/toolrunner.go
nickmisasi: Adjusted purity

crspeller (approved):

PR #778, opened 2026-05-25
Use Bifrost-provided token metadata with config fallback

crspeller:
#### Summary

The plugin already depends on Bifrost but ignored most of what it exposes about tokens: per-model `MaxInputTokens` / `MaxOutputTokens` / `ContextLength` were stripped from `ListAllModels` responses, the provider-side `CountTokensRequest` (Anthropic, OpenAI, Bedrock, Vertex) was never called, and the richer usage detail on `BifrostLLMUsage` (cached read/write tokens, reasoning tokens, cost) was discarded. As a result, `InputTokenLimit()` returned hardcoded provider defaults, `CountTokens` was a char/word heuristic, token-usage logs missed cache/reasoning/cost dimensions, and admins had to enter token limits by hand even when Bifrost knew them.

This change plumbs Bifrost's metadata through the existing wiring, with the configured value used as fallback when Bifrost has no data:

**Backend**
- Drop the hardcoded provider switch in `bifrost.LLM.InputTokenLimit()`; return the stored value as-is. `0` keeps its existing "no client-side truncation" meaning.
- Short-circuit `TruncationWrapper` when limit is `0`; otherwise run the heuristic truncation, then (when the wrapped LLM implements the new optional `llm.TokenCounter` interface and the heuristic estimate is at or above `0.8 * budget`) verify with `CountRequestTokens` and drop the oldest message + retry once if the provider count exceeds the raw limit.
- Add `OutputTokenLimit()` to the `llm.LanguageModel` interface, wired through every wrapper, fake, and the regenerated `mockery` mock.
- Extend `llm.ModelInfo` with `InputTokenLimit` / `OutputTokenLimit` / `ContextLength`; populate via a new `convertBifrostModels` helper so the webapp's model dropdown can see them.
- Add `bifrost.LLM.CountRequestTokens` wrapping `client.CountTokensRequest`.
- Extend `llm.TokenUsage` with `CachedReadTokens` / `CachedWriteTokens` / `ReasoningTokens` / `Cost`; populated through new `convertChatUsage` / `convertResponsesUsage` helpers in both streaming paths. Token-usage log gains the matching fields and bumps `TokenUsageLogSchemaVersion` to `2`. New OTel attribute keys (`agents.llm.cached_read_tokens` etc.) emitted only when non-zero, in both the Bifrost tracer and the plugin-side streaming spans.

**Webapp**
- `ServiceFields` exported and reworked: when the currently-selected model is found in the fetched-models list and has `inputTokenLimit` / `outputTokenLimit` populated, the matching `TextItem` is disabled, prefilled with the provider value, and shows "Auto-detected from provider" helper text. Independently per field. Most-recent manual values are preserved in local state across enable/disable flips so switching back to an unknown model restores them. A `useEffect` pushes the effective value into `props.service` so the saved config matches what's displayed.

**Tests**
- New Go tests for the resolution order, the truncation short-circuit, the truncation safety check (skip-far / call-near / drop-oldest cases), the `ModelInfo` conversion, the provider-side count call (via `httptest`), the usage-conversion helpers, the schema bump, and the OTel attribute emission.
- New Jest tests covering disable+prefill, the persisted-value contract, the manual-value preservation across flips, and the unknown-model fallthrough.
- New Playwright spec (`e2e/tests/system-console/token-limit-auto-detect.spec.ts`) stubbing `/admin/models/fetch` via `page.route` and asserting the disabled/prefilled fields; assigned to `e2e-shard-4`. `make check-shards` passes.

#### Ticket Link

N/A

#### Screenshots

N/A — covered by the Playwright spec above.

#### Release Note
```release-note
The Input token limit and Output token limit fields in the AI service configuration now auto-fill from the selected model when the provider reports it (Anthropic, Gemini, OpenAI, Cohere, Mistral via Bifrost), and become disabled to prevent accidental override. The hardcoded per-provider token-limit defaults have been removed: services that previously had no token limit configured will now rely on the provider's own limit and surface its error if exceeded. Token-usage logs gained ` + "`cached_read_tokens`" + `, ` + "`cached_write_tokens`" + `, ` + "`reasoning_tokens`" + `, and ` + "`cost`" + ` fields; ` + "`schema_version`" + ` is bumped to 2.
```

## Summary by CodeRabbit

* **New Features**
  * System Console auto-detects and pre-fills model input/output token limits (fields disabled when provider values exist; UI shows “Auto-detected from provider”).
  * Models now surface input/output token limits and context length; output-token limits respected across features.
  * Token-usage reporting and telemetry include cached read/write tokens, reasoning tokens and cost; token-usage schema upgraded.

* **Tests**
  * Added unit and e2e tests for token-limit auto-detect, truncation safety, richer token-usage logging, and model metadata.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mattermost/mattermost-plugin-agents/pull/778?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crspeller:
@coderabbitai resolve

llm/truncation.go
```diff
@@ -23,21 +29,64 @@ func NewLLMTruncationWrapper(llm LanguageModel) *TruncationWrapper {
...
+	count, err := w.wrapped.CountTokens(ctx, *request, opts...)
+	if errors.Is(err, ErrUnsupportedTokenCount) || err != nil || count <= limit {
+		return
+	}
+	// System prompts carry behavioral instructions; never drop them.
+	if !dropOldestNonSystemPost(request) {
+		return
+	}
+	_, _ = w.wrapped.CountTokens(ctx, *request, opts...)
+}
```
nickmisasi, line 75: I think this needs to loop? It's possible that dropping the oldest non system post might not put you far enough below the threshold, but that second count would be ignored here. We need to keep truncating until count <= limit

llm/truncation.go
```diff
@@ -23,21 +29,64 @@ func NewLLMTruncationWrapper(llm LanguageModel) *TruncationWrapper {
...
 }
 
-func (w *TruncationWrapper) CountTokens(text string) int {
-	return w.wrapped.CountTokens(text)
+// maybeTruncate heuristically truncates and, when the estimate is near the
+// budget and the model supports it, asks the provider to verify and drops the
+// oldest non-system post once if still over.
+func (w *TruncationWrapper) maybeTruncate(ctx context.Context, request *CompletionRequest, opts []LanguageModelOption) {
+	limit := w.wrapped.InputTokenLimit()
+	if limit <= 0 {
```
nickmisasi, line 45: I don't think we want to disable truncation if TokenLimit is 0? That's the default IIRC

bifrost/bifrost.go
```diff
@@ -529,29 +535,116 @@ func (b *LLM) ChatCompletionNoStream(ctx context.Context, request llm.Completion
...
+	}
+	return usage
+}
+
+var providersSupportingCountTokens = map[schemas.ModelProvider]struct{}{
+	schemas.OpenAI:    {},
+	schemas.Anthropic: {},
+	schemas.Bedrock:   {},
+	schemas.Vertex:    {},
+}
```
nickmisasi: Any reason to skip some of the other's that Bifrsot supports? Looks like [cohere](https://github.com/maximhq/bifrost/blob/dev/core/providers/cohere/count_tokens.go) and [gemini](https://github.com/maximhq/bifrost/blob/dev/core/providers/gemini/count_tokens.go) both support CountTokens

webapp/src/components/system_console/service.tsx
```diff
@@ -82,6 +85,16 @@ const ServiceFields = (props: ServiceFieldsProps) => {
...
 
+    // Cached admin entries so toggling to a Bifrost-known model and back
+    // restores the prior manual values instead of the auto-detected ones.
+    const [manualInputLimit, setManualInputLimit] = useState<number>(props.service.tokenLimit);
+    const [manualOutputLimit, setManualOutputLimit] = useState<number>(props.service.outputTokenLimit);
+    useEffect(() => {
+        setManualInputLimit(props.service.tokenLimit);
+        setManualOutputLimit(props.service.outputTokenLimit);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [props.service.id]);
```
nickmisasi, line 104: I think we should bring the exhaustive deps back? If the tokenLimit or outputTokenLimit change for whatever reason without the ID changing, this won't fire and then stale cached admin entries could be written

bifrost/tracer.go
```diff
@@ -143,10 +143,7 @@ func (t *otelTracer) PopulateLLMResponseAttributes(_ *bschemas.BifrostContext, h
...
 	}
 	if usage := chatUsage(resp); usage != nil {
-		h.span.SetAttributes(
-			telemetry.LLMInputTokens.Int64(int64(usage.PromptTokens)),
-			telemetry.LLMOutputTokens.Int64(int64(usage.CompletionTokens)),
-		)
+		setUsageAttributes(h.span, usage)
 	}
 	if bErr != nil && bErr.Error != nil {
 		h.span.SetStatus(otelcodes.Error, bErr.Error.Message)
```
nickmisasi: This wasn't from this PR but noticed while reading. The error message here needs to be sanitized or else it might dump partial API keys into the span, like it was doing for OpenAI with logs a while back.

nickmisasi (changes requested):
Overall LGTM, pointed out a few things

llm/truncation.go
crspeller: It is but for most some providers now we are able to query the max tokens. If the admin doesn't set a limit then there is no truncation. 0 = unlimited no truncation. 

I think we can just tell admins to set that limit for models that don't report. 0/5 if you have a better idea.

bifrost/bifrost.go
crspeller: Fixes this so it attempts and reads the error message instead.

nickmisasi (approved):

PR #787, opened 2026-05-28
Rebuild bots when a DB-agent's service config changes

crspeller:
#### Summary

`EnsureBots`' optimistic equality check (the fast path that decides whether to grab the cluster mutex and rebuild) was looking only at file-config bots (`b.config.GetBots()`) and the services those bots reference. **DB-backed agents** (created through the `/agents` UI) get loaded *later* in the same function. So a service config change for a service used only by a DB-backed agent was invisible to the check — both `botConfigsEqual` and `serviceConfigsEqual` returned true (file config unchanged), and the function bailed out before reaching the rebuild loop. The user-visible symptom was that any change made to a service's config (token limits, API keys, etc.) didn't take effect until the plugin was reactivated.

**Fix:** Extracted `snapshotBotsAndServices()`. It returns the full bot lineup (file bots capped by license + DB agents) and the services referenced by **all** of them. Both equality checks in `EnsureBots` now use this snapshot, so a service change for a DB-agent's service triggers a rebuild on the next `EnsureBots` call.

**Regression guard:** `TestEnsureBotsRebuildsBotWhenServiceInputTokenLimitChanges` configures an empty file-config Bots slice + a DB-backed agent referencing `svc1`, calls EnsureBots, captures the initial limit. Mutates the service to a known unusual value (250000), calls EnsureBots again, asserts the limit changed and now matches 250000. Without the fix this test fails — EnsureBots takes the fast-path skip and the bot keeps reporting the previous limit.

#### Ticket Link

N/A — found while debugging the context-usage indicator (separate PR stack).

#### Release Note
\`\`\`release-note
Fixed a bug where edits to an AI service config (token limits, API keys, model, etc.) did not take effect for DB-backed agents until the plugin was restarted.
\`\`\`

## Summary by CodeRabbit

* **Bug Fixes**
  * Bot lineup now uses a non-destructive snapshot of file and DB-backed configurations, ensuring service setting updates (e.g., input token limits) are applied and active bots are correctly activated/deactivated.

* **Tests**
  * Added tests verifying snapshotting does not mutate file-config bots and that rebuilds apply service changes for DB-backed bots.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mattermost/mattermost-plugin-agents/pull/787?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/bots/bots.go
+++ b/bots/bots.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"reflect"
+	"slices"
 	"sync"
 
 	"github.com/mattermost/mattermost-plugin-agents/assets"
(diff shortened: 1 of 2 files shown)


nickmisasi (approved):
LGTM, looks like one test failure pipeline

PR #788, opened 2026-05-28
Per-source context composition + RHS context usage indicator

crspeller:
> Stacks on #778 — review against `bifrost-token-metadata`. Will be retargeted to `master` once #778 lands.

#### Summary

Surfaces per-conversation context-window utilization in the AI RHS, with a per-source breakdown on click.

- **Per-source composition.** `CompletionRequest.Composition()` derives a per-source breakdown directly from the request — `system`, `history` (including attached-file text), `tool_defs`, `tool_results`, and `image` — so it stays consistent through truncation with no stored state. `ComputeComposition` scales the cheap estimator's proportions to the provider's authoritative total, so per-source numbers add up to the real total.
- **`GET /conversations/:conversationid/context` endpoint.** Returns the composition without making an LLM call. Auth mirrors `handleGetConversation`. Prefers the provider's `CountTokens` over the estimator and reports `total_source` as `counted` / `provider` / `estimated` so the UI knows how trustworthy the total is.
- **Bifrost `CountTokens` body-shape fix.** The `count_tokens` endpoint shares the messages schema but rejects some params — Anthropic 400s on native server tools (`web_search`, `file_search`, `code_interpreter`), OpenAI 400s on `max_output_tokens`. We strip those but keep function tool definitions, which do contribute to the input-token count (dropping them undercounts every tools-enabled bot). Pinned with regression tests for each shape.
- **Telemetry.** Five OTel keys under `agents.llm.tokens.*` on the LLM-call span, one per source, derived from the request and scaled to the provider's input total.
- **RHS indicator.** SVG ring + inline `%` in the RHS header with a popover for the breakdown. Graceful states: limit unknown → absolute count, no ring; util < 5% → hidden; util > 100% → alert icon + truncation note. Updates in near-real-time via cross-cache invalidation from the existing `invalidateConversation` callsites, so no callsite has to remember the new cache. The RHS header was reorganized so the ring appearing/disappearing doesn't shift `New chat` or the bot/tool pickers.
- NOTE: Limitation of this PR is that it doesn't properly track files. This will be fixed in a future PR when we modify how files are handled. (no sense in fixing the issue here when we are going to change file handling) 

#### Ticket Link

Jira https://mattermost.atlassian.net/browse/MM-68595

#### Screenshots
image
image
image

#### Release Note
```release-note
Added a context-usage indicator to the AI RHS header showing how much of the model's context window the current conversation is using, with a per-source breakdown (system prompt, history, images, tool definitions, tool results) on click. Added a new \`GET /conversations/:conversationid/context\` endpoint that returns the same data programmatically.
```

## Summary by CodeRabbit

* **New Features**
  * Context Usage Indicator showing token utilisation with per-source breakdown and popover
  * Per-source token composition (system, history, tool defs, tool results, images) and conversation context endpoint
  * Client API and hook for fetching/caching/invalidation of conversation context; UI wiring for thread header

* **Telemetry**
  * New per-source LLM token attributes for traces

* **Tests**
  * Comprehensive unit and integration tests covering composition, counting, telemetry and UI behaviours

* **i18n**
  * Added user-facing strings for context/token displays and tooltips


webapp/src/components/rhs/context_usage_indicator.tsx
```diff
@@ -0,0 +1,334 @@
...
+                    )}
+                </PopoverSubtitle>
+                {isOver && (
+                    <OverflowNote>
+                        <FormattedMessage defaultMessage='Context is over the limit — older messages are being dropped to fit.'/>
+                    </OverflowNote>
+                )}
+                {!hasLimit && (
+                    <EstimatedNote>
+                        <FormattedMessage defaultMessage='Provider does not publish a context window limit; utilization can&apos;t be computed.'/>
```
nickmisasi: I think `can&apos;t` get's printed literally when run through FormattedMessage

nickmisasi (approved):
one small thing to address but LGTM

PR #791, opened 2026-06-01
Lazy-load file attachments via an on-demand read_file tool

crspeller:
#### Summary

Large file attachments were inlined into the LLM context on every turn — the dominant cause of context-window blowouts (a single 5 MB text file is ~1.25M tokens, re-sent on every turn, after which truncation drops whole posts). This PR makes file loading lazy:

- **Small files** (≤ 8 KiB of readable text) are still inlined directly, to avoid an extra round trip.
- **Larger files** are shown to the LLM as a compact metadata descriptor (name, type, size, File ID) instead of their contents. The model fetches the text on demand via a new `read_file` MCP tool with ranged reads, so a large document costs almost nothing until it's actually needed. Once fetched, the tool result persists in conversation history, so it isn't re-read on later turns.

`read_file` lives on the embedded Mattermost MCP server, backed by a new authenticated plugin endpoint (`POST /api/v1/files/content`). This is necessary because the server-extracted text of PDFs/Office docs (`FileInfo.Content`) is never returned over the core REST API, so it must be read plugin-side with an explicit per-user channel-permission check. `read_file` is vetted to **auto-run in DMs** (no approval needed, matching the previous auto-inline UX), but **asks before running in a channel** — it only verifies the requesting user's own file access, not that the file belongs to the current thread, so auto-running it where other members would see the result is gated behind approval.

Images are unchanged (their token cost is negligible on modern context windows).

##### QA test steps
1. DM a bot a small `.txt` → its contents are still inlined (no tool call).
2. DM a bot a large `.txt` and a PDF → the prompt shows only metadata; ask about them → the bot calls `read_file` and answers from the contents; a follow-up turn does not re-fetch.
3. Verify a file you can access is readable, and a file in a channel you cannot access is refused.
4. `make evals-ci` (or `GOEVALS=1 LLM_PROVIDER=anthropic go test ./mcpserver/ -run TestReadFile`) — three agentic evals cover on-demand reads, paging through a large file, and unreadable binaries.

#### Ticket Link
NONE

#### Release Note
```release-note
Large file attachments are now loaded into the AI context on demand via a new read_file tool instead of being inlined automatically, sharply reducing context-window usage for conversations with attachments. Added an authenticated plugin API endpoint POST /api/v1/files/content.
```

## Summary by CodeRabbit

* **New Features**
  * Added read_file tool for on‑demand retrieval of file attachment text (respecting user channel permissions).
  * Supports paginated/ranged reads, rune-aware slicing, and automatic extraction from PDFs/Office with text/* fallback.
  * Introduced descriptor-only metadata for large or binary attachments and an “Attached files” descriptor section to enable lazy on‑demand reads.
  * read_file included in Mattermost‑vetted auto-run-in-DM tooling.

### DIFF
--- a/mcp/vetted_tools.go
+++ b/mcp/vetted_tools.go
@@ -124,6 +124,18 @@ func askToolConfigs(toolNames []string) []ToolConfig {
 	return configs
 }
 
+func autoRunEverywhereToolConfigs(toolNames []string) []ToolConfig {
+	configs := make([]ToolConfig, 0, len(toolNames))
+	for _, toolName := range toolNames {
+		configs = append(configs, ToolConfig{
+			Name:    toolName,
+			Policy:  ToolPolicyAutoRunEverywhere,
+			Enabled: true,
+		})
+	}
+	return configs
+}
+
 // githubSecurityAskTools are GitHub Copilot MCP reads that surface vulnerability
 // and secret-scanning posture; they default to ask rather than auto-run.
 var githubSecurityAskTools = map[string]struct{}{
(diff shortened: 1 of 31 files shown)


crspeller:
@coderabbitai review

mcpserver/tools/files.go
```diff
@@ -0,0 +1,93 @@
...
+	content, err := p.fileContentService.GetContent(mcpContext.Ctx, mcpContext.UserID, args.FileID, args.Offset, args.Limit)
+	if err != nil {
+		if errors.Is(err, files.ErrForbidden) {
+			return "you do not have permission to read this file", nil
+		}
+		return "failed to read file", fmt.Errorf("error reading file %s: %w", args.FileID, err)
+	}
+
+	return formatFileContent(content), nil
+}
```
nickmisasi, line 68: This doesn't have any vetting that the file that's being fetched is part of the current thread or context - it only really checks that the person can access that file. I think this is risky, especially considering we're seeding the tool config vetted as auto_run_everywhere

mcp/vetted_tools.go
```diff
@@ -242,14 +254,29 @@ var figmaVettedToolConfigs = autoRunInDMToolConfigs([]string{
...
+		"search_users",
+		"get_user_channels",
+	})
+
+	// read_file auto-runs everywhere. Before lazy file loading, an attachment's
+	// extracted content was inlined into the prompt automatically — in DMs and
+	// channels alike, with no approval. Fetching it on demand via read_file
+	// preserves that behavior rather than introducing an approval step for
+	// content the user already attached and the bot could already read.
+	configs = append(configs, autoRunEverywhereToolConfigs([]string{"read_file"})...)
```
nickmisasi: If the intention is to have this backfill I don't think it'll work, because the vetted tool configs only apply when there isn't already one

nickmisasi (commented):
Overall looks fine - but I'm worried about the auto run everywhere and lack of guardrails

mcpserver/tools/files.go
crspeller: Backed it off to the regular auto run like everything else. No risk in DMs, same risk as reading posts in other places.

mcp/vetted_tools.go
crspeller: Fixed.

nickmisasi (approved):

PR #813, opened 2026-06-10
Add create_draft tool for composing messages without posting

hanzei:
#### Summary
During my workflows, I often ask my agent to create respones in mattermost. Creating them directly as the post creation API is not a option, as I want to review and rewrite the post first. Hence, this PR add a create_draft tool for agents to use.

#### AI Summary
Added a new `create_draft` MCP tool that allows users to compose and save draft messages in Mattermost channels without immediately posting them. Drafts are saved privately for the user and appear prefilled in their message input box, allowing them to review, edit, and send the message themselves.

**Key Features:**
- Create or update draft messages in channels or threads
- Drafts are not posted—they're saved for user review before sending
- One draft per channel/thread (subsequent calls overwrite previous drafts)
- Supports optional file attachments
- Validates channel and team display names for context verification
- Tracks AI-generated drafts when enabled

**Implementation Details:**
- Added `CreateDraftArgs` struct with required parameters (channel_id, message, channel_display_name, team_display_name) and optional parameters (root_id for thread replies, attachments)
- Implemented `toolCreateDraft` function with validation of channel/team display names and file upload support
- Integrated with Mattermost's `UpsertDraft` API to save drafts
- Added comprehensive integration tests covering happy path, invalid channel ID, and missing parameters scenarios
- Updated documentation in README.md and admin guide

#### Test Plan
Added integration tests in `tools_integration_test.go` covering:
- Happy path: Successfully creating a draft and verifying it's saved (not posted to channel)
- Invalid channel ID validation
- Missing required parameters validation

Existing test infrastructure validates the tool registration and parameter schema.

#### Release Note
```release-note
Added new `create_draft` MCP tool to compose and save draft messages in Mattermost channels without posting them immediately. Drafts appear prefilled in the user's message input box for review before sending.
```

https://claude.ai/code/session_014QrBE7HT8LrhGF35HcqA3Q

## Summary by CodeRabbit

* **New Features**
  * Added a new `create_draft` built-in tool to create or update private Mattermost drafts for a channel or (optionally) a thread reply, supporting optional attachments.
  * When AI-generated tracking is enabled, drafts include AI-generated metadata.
* **Documentation**
  * Updated the admin guide and server README with `create_draft` usage details, required context fields, and attachment guidance.
* **Tests**
  * Expanded integration coverage for successful channel/thread draft creation and multiple validation error scenarios.


mcpserver/tools/posts.go
```diff
@@ -67,6 +77,8 @@ func (p *MattermostToolProvider) getPostTools() []MCPTool {
...
 
+	createDraftDesc := fmt.Sprintf("Create or update a draft message in a Mattermost channel. A draft is NOT sent — it is saved privately for the user and appears prefilled in their message input box for that channel (or thread) so they can review, edit, and send it themselves. Prefer this over create_post whenever the user wants to compose or stage a message rather than post it immediately. There is one draft per channel/thread, so calling this again for the same channel_id (and root_id) overwrites the previous draft. IMPORTANT WORKFLOW: You MUST first call get_channel_info to obtain the channel_id, channel_display_name, and team_display_name. Parameters: channel_id (required), message (required), root_id (optional - to draft a thread reply)%s. Example: {\"channel_id\": \"h5wqm8kxptbztfgzpaxbsqozah\", \"message\": \"Draft for your review\"}", attachmentsParam)
```
nickmisasi: >Prefer this over create_post whenever the user wants to compose or stage a message rather than post it immediately

This would be a big behavioural change that I think a lot of people will disagree with or be annoyed by

nickmisasi (commented):
I'm worried that this is going to skew certain models to being overly helpful and calling this tool too often. I understand it's a way you like to work, but that doesn't necessarily apply to everyone. My biggest concern is UX-wise, there's no way for the Agent to easily direct the user to the draft to actually view it. 

I think the better approach to this would be to just add custom instructions telling the Agent to provide the user with the prospective post verbatim and ask for edits before calling create_post. 

Thoughts @crspeller?

crspeller (changes requested):
Agree with @nickmisasi here. I don't see an issue with having this tool for niche workflows. But we should soften the language here and let the model know it should only use this tool when explicitly asked. Certainly because "draft me a message" probably means they want it to literally respond with the message not create a draft.

hanzei:
Thanks for the feedback @crspeller @nickmisasi :+1: 

I've soften the language a bit to not encourage the agent to use draft over post creation. PTAL

hanzei:
UX wise, I'm not too concerned. The list at `/drafts` is ordered by latest update, hence the suggestion of the agent will always be on the top. FWIW: the gmail mcp works the same way. It creates a draft for you and then asks the user to send it.

nickmisasi (commented):
I think this introduces a bit of a broken window. When you ask it to create a draft for a channel you're actively viewing, the drafts menu pops up top left, but the text box doesn't get populated until you move away and come back. 

I'm still a little worried about this even with the current language. With dynamic tool calling disabled, and a lesser model (GPT-5.2), and somewhat confusing, but not unrealistic prompting, it consistently opts to actually create the draft when my intention was to post.

Image

I know I explicitly said the words "Create a draft post" - but I think a lot of people have mentally trained themselves to say this in order to have the Agent not run off and take action too quickly

hanzei:
Thanks for checking my proposal so carefully @nickmisasi!

It seems like there is no good path forward here. The UX of drafts needs to be adjusted to fit the new workflow. We would also need URLs for drafts so external agents can refer to it. All of that is out of scope of this PR.

Closing.

PR #826, opened 2026-06-11
Add AskUserQuestion tool for interactive user questions

crspeller:
#### Summary

Adds an `AskUserQuestion` built-in tool that lets agents ask the requesting user a question with a set of options to pick from. The question rides the existing tool-approval flow: the pending tool call renders as a question card on the bot post, the user's selection becomes the tool result, and the conversation resumes with the answer.

**How it works**

- New `llm.Tool.UserInteraction` kind (`"select"`): tools answered by the user instead of executed by the server. The server and webapp branch on the declared kind, never on the tool name.
- The tool is only cataloged in contexts where someone can actually answer (new `WithLLMContextInteractive()` context option): DMs, human channel mentions with tool calling enabled, approval clicks, and regeneration. Automated invokers (bot `activate_ai`, bridge API, evals, summarization flows) never see it, and interaction tools are never auto-approved regardless of policy config.
- `POST /posts/{id}/tool_call` accepts an optional `tool_answers` map. Answers are validated against the offered options before any state changes; an invalid answer returns 400 and leaves the question pending. Answer results are terminal and shared (the user authored them), so they skip the channel Share/Keep-private stage, and the follow-up streams as soon as no share decision remains.
- New `QuestionCard` webapp component per the Figma designs: numbered options for single-select, checkboxes for multi-select, Skip/Accept, and resolved (Answered/Skipped) states.
- Mixed-batch fix: a tool with an auto-run policy that gets paused in the same batch as a question no longer asks for approval. The runner tags it (`WouldAutoExecute`, display-only), the webapp hides it while pending, and `HandleToolCall` re-checks the policy server-side on resume and executes it as auto-approved. This also fixes the pre-existing variant where an `auto_run` tool batched with an `ask`-policy tool demanded a manual click.

**QA test steps**

1. In a DM with an agent, ask something like "Ask me which of three colors I prefer using your question tool." The bot post shows a question card; pick an option and Accept — the bot continues using the selection. Repeat with Skip.
2. In a channel (with channel mention tool calling enabled), @mention the agent with the same prompt. Answer the question; the follow-up streams immediately with no Share/Keep-private step for the answer.
3. With an MCP tool set to auto-run, prompt the agent to use that tool *and* ask a question in the same turn. Only the question card shows; on answering, the auto-run tool executes and appears as a collapsed "Auto-approved" row.
4. Verify a bot-posted `activate_ai` message never produces a question card, and that another channel member sees the pending question as a generic tool card without the question contents until answered.

#### Ticket Link

NONE

#### Screenshots

image

image

#### Release Note

```release-note
Added an AskUserQuestion built-in tool that lets agents ask the requesting user a multiple-choice question through the tool-approval UI in DMs and human channel mentions. The POST /plugins/mattermost-ai/post/{post_id}/tool_call endpoint now accepts an optional tool_answers field carrying the selected options. Tools with an auto-run policy that were paused in the same batch as a call requiring user action no longer ask for manual approval; they execute automatically once the user resolves the rest of the batch.
```

## Release Notes

**New Features**
* Added interactive “question” prompts in the tool-approval UI, with selectable options, optional “something else”, and Accept/Skip actions.

**Improvements**
* Tool calls now expose user-interaction details (including “would auto-execute” markers) to improve approval and streaming behaviour.
* Approving a tool now submits any provided answers, with different DM vs channel follow-up handling.

**Bug Fixes**
* Invalid or missing user answers now return a client error (400) and do not alter pending state.


conversations/tool_approval.go
```diff
@@ -136,7 +173,47 @@ func (c *Conversations) HandleToolCall(ctx context.Context, userID string, post
...
+			block.Shared = conversation.BoolPtr(true)
+			executedAny = true
+			toolResults = append(toolResults, toolrunner.ToolResult{
+				ToolCallID: block.ID,
+				Name:       block.Name,
+				Result:     "User skipped the question",
+				IsError:    true,
+			})
+		case autoExec(llm.ToolCall{Name: block.Name, ServerOrigin: block.ServerOrigin}):
+			// The block was paused only because the rest of its batch needed
```
nickmisasi: Sort of pedantic but should we filter out tools that were previously persisted as "ask" that may have been flipped to auto run mid turn? On one hand I feel like respecting the Admin policy makes sense, but on the other I could see it being unexpected

webapp/src/components/question_card.tsx
```diff
@@ -0,0 +1,430 @@
...
+import LoadingSpinner from './assets/loading_spinner';
+
+// Parsed shape of the AskUserQuestion tool arguments. Mirrors
+// mmtools.AskUserQuestionArgs on the server.
+export interface QuestionOption {
+    label: string;
+    description?: string;
+}
+
+export interface QuestionArgs {
```
nickmisasi, line 19: From the discussion I had with Asaad/the resulting designs we had explicitly included a "free form" entry option (similar to Claude/Cursor's "Something else....") - I think we should provide that option

nickmisasi (commented):
One nit pick, and one thing I think is missing. Overall LGTM, I like the new front-end tool pattern

webapp/src/components/question_card.tsx
crspeller: image

nickmisasi (approved):
LGTM!

PR #881, opened 2026-07-13
Rich tool call rendering: readable arguments and responses, post previews, path-identical metadata

crspeller:
#### Summary

Replaces the raw-JSON display of LLM tool calls with readable, safe rendering: a generic labeled field list for arguments **and responses**, permalink-style post previews for `read_post` and `create_post`, and the server plumbing that makes a call render identically live and after reload.

**Foundation: path-identical tool metadata**
- `llm.ToolCall` gains `Title` (resolved MCP display name); the unused `Schema` field is removed from the wire (providers build tool definitions from `llm.Tool.Schema`, and requests from `id`/`name`/`arguments`/`result` — `ToolCall.Schema` never reached a provider and the webapp ignored it).
- `mcp` `GetTools` resolves the effective display title from MCP metadata (`title` > `annotations.title`, whitespace-only treated as absent) and Unicode-sanitizes `Title`/`Description` at capture (bidi/invisible-char defense).
- `ContentBlock` gains `Title`/`Description`; both block writers and `PostToBlocks`/`BlocksToPost` carry them; `SanitizeForDisplay` sanitizes them; `FilterForNonRequester` keeps them visible (tool identity, like `Name`). `redactToolCalls` now keeps `Title`/`Description` too — fixing a live↔persisted asymmetry where `Description` was dropped by omission on the live path. `buildContentBlocks` also gains the previously missing `MCPBareName`.
- A reflection-based **parity drift-guard** (`streaming/tool_call_parity_test.go` + `conversation/tool_use_writer_parity_test.go`) maps every `llm.ToolCall` JSON field to a persistence + redaction policy and fails (per-field subtests) if any writer or redactor diverges.

**Card UX (per design, Mattermost styling)**
- Each tool call renders as a bordered card (border/radius/shadow matching QuestionCard) with a chevron + tool-name header; collapsed cards are slim single-row containers.
- **Arguments and responses share the same generic treatment**: a two-column grid (bold labels left, values right) for JSON objects — strings as wrapped text, numbers/booleans/primitive arrays as value pills, nested objects as inline JSON — and clamped plain text for non-JSON responses, with **Show more** to expand. **View raw** (hosted on the shared card shell) always exposes the exact arguments payload. Accept renders as the filled primary action, Reject as the tinted secondary.
- Generic values render as plain text via styled-components — never `formatText`/markdown — so neither LLM-generated arguments nor tool responses can become a link-spoofing surface (the old code-block pipeline for results is gone). `arguments == null` (redacted) renders nothing; `{}` keeps the verbatim "No parameters required". The header toggle is keyboard-accessible (`role='button'`, Enter/Space, `aria-expanded`), and ellipsized titles expose a hover tooltip.

**Renderer registry + post preview cards**
- `tool_renderers/registry.tsx` routes each call by canonical identity (origin kind + bare name); first match wins; no match → generic card. `QuestionCard` is routed through it too.
- While awaiting approval, **`read_post` shows a Mattermost permalink-style preview of the referenced post**, and **`create_post` shows a preview of the post-to-be** — the message rendered exactly as it will appear once posted, authored by the requesting user (whose session the embedded tool executes with). So the user sees precisely what the tool will read or create before approving. The create_post preview is non-interactive (`pointer-events: none`): the post doesn't exist yet, so its permalink/avatar affordances can't lead anywhere; the read_post preview stays clickable since its post is real. Both fall back to the generic field list when the arguments don't parse, the post can't be fetched, or the call has already executed.

External MCP tools that declare a `title` now display it (sanitized, ellipsized); built-in/embedded names are unchanged. External-server render overrides are out of scope.

#### Before / after

Rendered with the real components at the current head of this branch (before = `master`'s `ToolCard`). The "after" shows both post previews (pending approval), the executed-call generic fallback, plain-text and JSON response rendering, and collapsed cards. (Screenshots are hosted on the `cursor/rich-tool-call-rendering-assets-f0d9` branch — not part of the plugin code.)

|  before  |  after  |
|----|----|
| Before: raw JSON arguments | After: read_post and create_post previews, generic field list and responses, collapsed cards |

#### Testing

- Go: table-driven tests for enrichment, block writers, redaction, sanitization, and MCP title resolution (incl. hostile-Unicode and whitespace-only titles); the parity drift-guard was verified to fail when `Title`/`Description` are dropped from `redactToolCalls`.
- Webapp: 313 Jest tests across 32 suites — field-list type dispatch, key-order stability, Show more / View raw, response rendering (JSON field list, plain text with no markdown side effects), registry routing (QuestionCard, both preview cards incl. executed/malformed/fetch-failure/external fallbacks, non-interactive create_post preview vs. interactive read_post preview), and ToolApprovalSet decision logic. ESLint + tsc clean; i18n in sync.
- e2e: `tests/tool-config/mock-api/tool-preview-cards.spec.ts` (mocked LLM, shared container) covers both cards — read_post shows the seeded post's preview and create_post shows the post-to-be's message before approval, with View raw exposing the exact payloads; assigned to `e2e-shard-1`. Existing tool-config specs pass unchanged (display names, button labels, status text, and the expand target are unchanged).

#### Ticket Link

NONE

#### Release Note
```release-note
Tool call approval cards were redesigned: arguments and responses now display as readable field lists on bordered cards, read_post and create_post calls show permalink-style previews of the referenced or to-be-created post, and a "View raw" toggle shows the exact JSON payload. External MCP tools that declare a title now display it.
```

<div><a href="https://cursor.com/agents/bc-8677d6de-9cc4-42d2-977e-a75f61cad9ec"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-8677d6de-9cc4-42d2-977e-a75f61cad9ec"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **New Features**
  * Tool cards now surface MCP-provided tool **title** and **description** consistently across live updates and reloaded conversations, including bare-name identity.
  * Added rich, structured cards for supported embedded tools, with improved “View raw” and formatted argument/result rendering.
* **Bug Fixes**
  * Hardened display sanitization for hostile bidirectional characters in tool titles, descriptions, and annotation titles.
  * Improved privacy redaction so non-requester views match between live and persisted tool-call rendering.
* **Tests**
  * Added parity and round-trip coverage for tool identity fields and redaction behaviour.


webapp/src/components/llmbot_post/turn_content_utils.ts
```diff
@@ -98,7 +98,11 @@ function toolUseBlockToToolCall(block: ContentBlock, resultMap: Map<string, Cont
     return {
         id: block.id ?? '',
         name: block.name ?? '',
-        description: '',
+
+        description: block.description ?? '',
+        title: block.title ?? undefined, // eslint-disable-line no-undefined
+        server_origin: block.server_origin ?? undefined, // eslint-disable-line no-undefined
+        mcp_bare_name: block.mcp_bare_name ?? undefined, // eslint-disable-line no-undefined
```
nickmisasi: This could be sort of a trap. Could we use an intentional null or something here instead? I'm just worried about `cannot read x of undefined` or something crashing the webapp

webapp/src/components/tool_renderers/posts/create_post.tsx
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) 2023-present Mattermost, Inc. All Rights Reserved.
```
nickmisasi, line 1: Could be worth adding a `./tool_renderers/posts/read_post.tsx` and `create_post.tsx` (or just `read.tsx` and `create.tsx`?)

nickmisasi (commented):
Flagged 2 thing but otherwise LGTM

nickmisasi (approved):
Approving to unblock and remove from my list, but I think you might have missed the 2 things i mentioned above

crspeller:
@nickmisasi Sorry about that, feedback is now addressed.

nickmisasi (approved):

PR #937, opened 2026-07-28
Add service account authentication for external MCP servers

nickmisasi:
#### Summary

Adds **service account authentication** for MCP servers, so automation-oriented agents can call external MCP servers (e.g. Jira) with admin-provisioned shared credentials instead of per-user OAuth.

**What's new**

- Each remote MCP server config gains a **Service Account Authentication** headers section in the System Console (mirrors the existing Headers editor; blank rows and names that collide after trimming/canonicalization are filtered fail-closed via `EffectiveServiceAccountHeaders()`).
- Agents gain a single all-or-nothing **"Use service accounts for authentication"** setting on the agent modal's MCPs tab, with a warning banner explaining the shared-access implications. **Only system administrators can enable it** (server-enforced; anyone who can manage the agent can turn it off).
- When enabled (and remote MCP is licensed): external MCP servers authenticate with the configured SA headers (servers without SA headers are **excluded fail-closed** from the agent's catalog — no fallback to user OAuth); embedded/plugin MCP tools run as the **agent's bot account** (channel membership is the internal access boundary); all per-user OAuth machinery (connect prompts, `mcp_oauth_needed` state, per-user tool preferences, RHS Tools popover) is bypassed/hidden. `/ai_bots` reports the *effective* mode, so the UI degrades to per-user UX on unlicensed servers.
- The contract applies in every context — DMs, channel mentions, and the AI bridge. Interactive approval is unchanged: the human initiator approves, the service account executes.
- Attribution: token usage logs gain `acting_user_id` and `tool_auth_mode`.
- Bridge: SA-flagged agents use the SA catalog regardless of the caller-supplied `user_id` (which still drives usage ACLs and attribution); discovery uses the same catalog as execution.
- Unlicensed servers: SA mode is fully off — the agent authenticates per-user like any other agent.
- Non-SA agents are unchanged everywhere, including bridge `user_id` semantics.
- Docs updated: `docs/features/multiplayer_tool_calling.md`, `docs/admin_guide.md`, `docs/features/managing_agents.md`, `public/bridgeclient/README.md`, user guide.

**Permission model while SA is on**

Enabling SA (and changing Access / MCP grants while it stays on) requires `manage_system`. Agent managers may edit day-to-day config without an admin:

| Sysadmin-only while SA on | Editable by agent managers |
| --- | --- |
| Enabling SA | Display name, custom instructions, model |
| Access (users/channels/teams/admins) | AI service (`serviceID`), Enable Tools, native tools, dynamic tool loading |
| MCP tool grants / auto-enable | Vision, reasoning, structured output, max tool turns, avatar |
| | Turning SA **off** |

Server implementation is **fail-closed**: updates apply to a proposed `BotConfig`, then `serviceAccountChangeNeedsAdmin` compares after clearing the manager-editable allowlist (`api/agent_acl.go`). Any newly added agent field is admin-only until explicitly allowlisted. Webapp soft-locks Access and MCP grant controls and keeps Save enabled for manager-editable fields; the SA checkbox stays reachable as the off-switch.

**Implementation notes**

- SA connections pool per `(bot user ID, service-account mode)` in `mcp.ClientManager`, with an isolated shared tools cache namespace (`sa:<server>`); `X-Mattermost-UserID` is the agent bot's ID. The "SA never surfaces OAuth prompts" invariant is enforced structurally in `Client.oauthNeededError`.
- Catalog selection lives at the single funnel (`llmcontext.Builder.getToolsStoreForUser`) via `UsesServiceAccountCatalog`, so conversations, the bridge, discovery, and channel analysis all inherit it with no per-site branching.
- The agent editor skips orphaned-tool reconciliation for SA agents (the editing user's per-user catalog is not authoritative for an SA agent's grants).
- New migration `000010` adds `UseServiceAccountAuth` to `Agents_UserAgents`.
- Bridge `tool_hooks` are deliberately not accounted for in SA mode (their only consumer, the Channel Automation plugin, is being retired).

**QA**

- Adding SA headers in the console, creating an SA agent (warning banner), and saving all persist correctly.
- End-to-end: chatting with an SA agent and calling an external server's tool returns the **service account's** identity (not the signed-in user), after human approval, with no OAuth prompt; RHS Tools popover is hidden.
- Regression: a normal agent keeps the per-user Tools popover with Connect for the same server.
- Non-admin managers can save custom instructions, model, display name, AI service, tools/native tools, and dynamic tool loading while SA stays on; Access and MCP grants are disabled in the UI and rejected by the API (403) if changed.
- Verified with `go test ./api/... ./mcp/...`, webapp jest for agent config/access/mcps tabs, and merge with current `master` (OAuth SDK migration + plugin MCP registration persistence).

**Screenshots**

From the perspective of an end user

When creating an Agent, they can't enable SA for that agent. SA only tools are reflected as unavailable with a popover explaining why. 
image
Same for the tools RHS
image
When the regular user is added as an administrator of an SA Agent, they are informed
image
Fields in Access and MCPs tabs are readable, but not editable:
image
The regular user (that is a delegated admin of the agent) can _disable_ SA authentication:
image

Admin perspective is the same as the above, except they are able to edit the Access and MCPs tabs.
Can configure the agent as an SA agent
image

#### Release Note

```release-note
Added service account authentication for external MCP servers. MCP server configurations accept a new Service Account Authentication headers section, and agents gained a "Use service accounts for authentication" setting that only system administrators can enable: when enabled, external MCP tool calls authenticate with the admin-configured service account headers (servers without them are excluded fail-closed) and embedded Mattermost tools run as the agent's bot account. While the setting is enabled, changing Access or MCP tool grants requires a system administrator; agent managers may still edit day-to-day settings such as display name, custom instructions, model, AI service, tools, and dynamic tool loading, and may turn the setting off or delete the agent. MCP connections now refuse HTTP redirects to a different origin and never send credentials cross-origin. Added the UseServiceAccountAuth column to the Agents_UserAgents table. Token usage logs gained acting_user_id and tool_auth_mode fields.
```

<div><a href="https://cursor.com/agents/bc-ad223957-1df3-4c36-9b75-d6789025a475"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ad223957-1df3-4c36-9b75-d6789025a475"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **New Features**
  * Added an agent-level “use service accounts for authentication” setting, persisted end-to-end and surfaced in the UI.
  * External MCP access now uses admin-configured service-account headers and fails closed for servers without valid headers.
  * Service-account agents hide the Tools UI and select the service-account tool catalog for execution.
* **Bug Fixes**
  * Enforced admin-only enablement for turning this setting on.
* **Documentation**
  * Updated admin/user and bridge/multiplayer calling guidance for service-account credential boundaries and fail-closed behaviour.
* **Tests**
  * Expanded coverage for persistence, catalogue selection/execution, and token-attribution by auth mode.


nickmisasi:
@cursor fix this PR's conflicts with master.

llmcontext/llm_context.go
```diff
@@ -251,15 +269,26 @@ func (b *Builder) getToolsStoreForUser(ctx stdcontext.Context, c *llm.Context, b
 		}
 
 		// Get tools from all connected servers
-		mcpTools, mcpErrors = b.mcpToolProvider.GetToolsForUser(ctx, userID)
+		if useServiceAccount {
```
edgarbellot, line 282: In SA mode this branch builds the embedded tool catalog for the bot user, not the requesting user, and [ensureEmbeddedSessionID(key.userID)](mcp/embedded_session_store.go#L46) mints the Mattermost session for that bot. Every embedded tool call then runs with the bot's own permissions. Nothing downstream re-checks the tool's output against what the requesting user is actually allowed to see - the code only asks "can the bot do this", never "can this user see this".

Confirmed it live: a user who was not a member of a private channel asked the SA agent to summarize it. The agent called `read_channel` as the bot (which is a member) and returned the full message history to someone who should never have seen it.

Suggested fix: the missing piece is a permission check between the tool call and its result - something that, for channel-scoped reads, intersects the target channel against the requesting user's own membership before returning content, not just the bot's. Right now that check doesn't exist anywhere in the SA path.

Same root cause also affects write tools like `add_channel_member` (unconfigured tools default to enabled/ask per `GetToolPolicy`, config/mcp_config.go) - it only didn't fire in my test because the bot lacked `manage_private_channel_members`, which is Mattermost's own guard, not this code's. If a bot ever holds a higher channel/team role, the same missing check lets that write through too.

api/api_agents.go
```diff
@@ -480,6 +500,13 @@ func (a *API) handleUpdateAgent(c *gin.Context) {
 		return
 	}
 
+	// Only the false→true transition escalates; an already-enabled flag was
```
edgarbellot: Once a system admin turns on `useServiceAccountAuth` for an agent, anyone who can manage that agent - even without `manage_system` - can freely widen it afterwards: open it to every user on the instance, auto-enable every tool, or point it at a different LLM provider.

Why: the [admin-only check](https://github.com/mattermost/mattermost-plugin-agents/blob/11cef5fbdb1af9ade955bfe77a25001b726aa883/api/api_agents.go#L503-L507) only fires on the false→true transition of the flag. Every field after that - [userAccessLevel, autoEnableNewMCPTools, adminUserIDs, serviceID](https://github.com/mattermost/mattermost-plugin-agents/blob/11cef5fbdb1af9ade955bfe77a25001b726aa883/api/api_agents.go#L288-L313) - gets overwritten under the same check as any normal agent edit: [canManageAgent](https://github.com/mattermost/mattermost-plugin-agents/blob/11cef5fbdb1af9ade955bfe77a25001b726aa883/api/api_agents.go#L179-L193), which passes for the agent's own creator or anyone in `adminUserIDs`. Neither of those requires being a system admin. So the code checks "is this person allowed to touch this agent at all", never "is this specific change one that needs admin rights".

Confirmed in my test instance: a system admin enabled SA on an agent created by, and scoped to, a regular member. That same member - still with no `manage_system` permission - then opened the agent to everyone and auto-enabled every tool, in one request that returned 200.

Is this intentional? If not, once `useServiceAccountAuth` is true, I would suggest that changes that widen its reach ([userAccessLevel](https://github.com/mattermost/mattermost-plugin-agents/blob/11cef5fbdb1af9ade955bfe77a25001b726aa883/api/api_agents.go#L295-L298), `userIDs`, `channelIDs`, `teamIDs`, `adminUserIDs`, `autoEnableNewMCPTools`, `enabledMCPTools`) should require the same [canEnableServiceAccountAuth](https://github.com/mattermost/mattermost-plugin-agents/blob/11cef5fbdb1af9ade955bfe77a25001b726aa883/api/api_agents.go#L206-L208) check as enabling the flag itself, not just `canManageAgent`.

mcp/client.go
```diff
@@ -450,9 +515,7 @@ func (c *Client) oauthNeededError(err error) error {
 
 func (c *Client) createSession(ctx context.Context, serverConfig ServerConfig) (*mcp.ClientSession, error) {
 	// Prepare headers for remote servers
-	headers := make(map[string]string)
-	headers[MMUserIDHeader] = c.userID
-	maps.Copy(headers, serverConfig.Headers)
+	headers := remoteConnectionHeaders(c.userID, serverConfig, c.serviceAccount)
```
edgarbellot, line 531: The service-account header (the shared, admin-provisioned credential) gets resent to a different host if the configured MCP server responds with an HTTP redirect.

Why: [remoteConnectionHeaders](https://github.com/mattermost/mattermost-plugin-agents/blob/11cef5fbdb1af9ade955bfe77a25001b726aa883/mcp/client.go#L122-L131) builds the SA header, and [createSession](https://github.com/mattermost/mattermost-plugin-agents/blob/11cef5fbdb1af9ade955bfe77a25001b726aa883/mcp/client.go#L516-L531) hands it to `httpClientForMCP`, which wires it into a custom `http.RoundTripper` that re-adds the header on every outgoing request - including the one Go automatically sends when following a redirect. Go's `http.Client` normally strips `Authorization` on a cross-host redirect, but that only removes what's already set on the request; it can't stop a custom transport from putting the same header back one line later.

Reproduced in my test instance: pointed this client at a test server configured with a fake SA header, made it respond with a redirect to a different host, and watched that header arrive intact at the redirect target.

Suggested fix: the `http.Client` built in `createSession`/`httpClientForMCP` needs a `CheckRedirect` that either blocks redirects to a different origin than `serverConfig.BaseURL`, or drops the injected headers before the redirected request goes out - without that, re-adding the header unconditionally on every `RoundTrip` defeats any redirect check added elsewhere.

edgarbellot (changes requested):
Great work! Let me know if you have any questions about my comments

llmcontext/llm_context.go
nickmisasi: This is something I am aware of and have done intentionally. If we were to implement your suggested fix where permissions are intersected, you'd end up in a state where the service account has _less_ access than if the invoker's permissions were just being used themselves. 

Channel membership being required is the thing that is gating access. It's actually something that's already in the product: https://github.com/mattermost/mattermost-plugin-agents/pull/611 - bot invocations of an Agent use the bot's access. I have an incoming commit to fix your finding here: https://github.com/mattermost/mattermost-plugin-agents/pull/937#discussion_r3729437042 - after which only trusted system administrators will be able to set or manage these agents (non system administrator maintainers will be read only)

Long term as possible scoped tokens, or even ABAC for bot accounts (because they're just users at the end of the day) come into play for channel access, this'll get more secure. But for now it's a deliberate, documented thing: https://github.com/mattermost/mattermost-plugin-agents/pull/937/changes#diff-358388706355238989c5323624130cf04d7e4b827ea00f92e298c1ea23d6d19dR57-R67

api/api_agents.go
```diff
@@ -480,6 +501,14 @@ func (a *API) handleUpdateAgent(c *gin.Context) {
 		return
 	}
 
+	// Any save that keeps service account auth enabled is admin-only, since every
+	// field shapes what the admin-granted credentials can do; anyone who can
+	// manage the agent may still turn it off.
+	if req.UseServiceAccountAuth && !canSaveServiceAccountAuth(a.pluginAPI, userID) {
```
crspeller: This seems clunky and involves system admins. Can we have these service accounts created on the user side and grant access to them? (This breaks Agents being managed by users, on Hub that means we are back to begging to update the PDE intake agent)

crspeller (changes requested):
Looks mostly good, but I think we need to allow non-system admins to make service accounts.

llmcontext/llm_context.go
edgarbellot: One thing I want to flag: the warning only shows up when someone opens that agent's own settings. The action that actually creates the exposure - adding the bot to a channel - happens through the normal "Add members" flow in channel settings, which has no way to know this particular user is an agent's bot. So there's a gap between where the warning lives and where the risky action actually happens.

A scenario that worries me: an admin sets up an SA agent, adds the bot to a couple of public channels and sees the warning. Months later, someone completely unrelated to this agent (who's never opened its settings) adds that same bot to a private channel for some other reason. They'd have no way of knowing what that means, since nothing surfaces there. From that point on, every user of the agent could read that channel.

This is reproducible end-to-end on the current build (after the Access/MCP-grant lockdown and the redirect fix landed): a plain user with no relation to the agent created a private channel, added the bot via a normal `POST .../members` call (no special permission required) and the agent then successfully read that channel's content on behalf of a different user who has no access to it themselves.

nickmisasi:
@crspeller @edgarbellot there was a bit of contradiction between both of your opinion's - Edgar your privilege-ratchet concern is difficult to address while also satisfying Christopher's concern around agents for all/managers needing to iterate. 

I've landed on a compromise between the fully open to all and fully closed to sysadmin approach

- **Still sysadmin only**: Enabling an Agent to be an SA agent - and, once enabled, the Access and MCPs tabs require system admin (ie, anything that could widen the blast radius from the shared credentials)
- **Editable by agent managers while SA is on:** day-to-day config - display name, instructions, model, AI service, enable/disable tools (including native) dynamic tool loading, vision, reasoning, etc. Any agent manager can also _disable_ SA if they choose. They can read but not adjust the Access or MCPs tabs. 

So to summarize - admins still gate the risky surface, but folks on Hub (like Bill) can still iterate on their agent's basic behaviour.

crspeller (approved):

llmcontext/llm_context.go
esarafianou: **Proposal**: for SA agents, add an admin-declared list of channels the Mattermost read tools may target, enforced at tool execution regardless of bot membership. Default value shouldn't be "all channels"

Reasoning: 

Existing channel access controls gate whether the agent responds in a channel and they're not consulted when a tool executes. `read_channel` uses the bot's session and sends a request to the API with the channel ID. So, as Edgar said, an agent scoped to one channel can be asked to read any channel the bot happens to be in and that membership is set by any user, ungoverned, invisible to whoever owns the agent.

The proposal is scoped to the embedded Mattermost tools and it doesn't need to touch external MCP servers. For an external server, the bound is the credential's own scope in the target system: the admin decides what the e.g. Jira account can see when they create it. The Mattermost side has no equivalent. Its "credential" is a session for a user whose reach is defined by channel membership, which the admin who enabled SA doesn't control, and which any user can extend at any time.

edgarbellot (approved):
The approach in #1012 LGTM, thank you Nick!

PR #942, opened 2026-07-30
Upgrade go-sdk to v1.7.0 (MCP 2026-07-28) and migrate OAuth stack to SDK auth/oauthex

crspeller:
#### Summary

Upgrades `modelcontextprotocol/go-sdk` v1.4.1 → **v1.7.0** so the plugin's MCP **client** and **servers** speak the latest protocol (`2026-07-28`) while staying compatible back to `2024-11-05`, and migrates our hand-rolled OAuth client onto the SDK's `auth`/`oauthex` packages.

**Why upgrade:** v1.7.0 is the first release supporting MCP `2026-07-28`; staying on v1.4.1 leaves us unable to connect to servers moving to the new protocol.

**Why migrate OAuth instead of keeping our own:** the SDK now owns the spec-compliance surface (RFC 8414/9728 discovery, RFC 7591 DCR, WWW-Authenticate parsing, error-chain preservation). Re-using it deletes ~560 lines we'd otherwise have to keep in step with the spec, and it surfaced a latent bug in our old code — refreshed tokens were never persisted, so any provider using refresh-token rotation silently broke. We implement the SDK's `auth.OAuthHandler` ourselves (rather than its turnkey handler) because our flow is asynchronous and multi-node: the 401 surfaces a link into chat, the user clicks later, and the callback can land on a different HA node — so PKCE/state/tokens must live in the KV store, not a blocking in-process handler.

**Why grants are pinned to their issuing authorization server:** without binding a grant to the exact issuer/token-endpoint/client it was obtained from, a compromised or reconfigured MCP server could, on refresh or re-auth, redirect the flow to an attacker-controlled authorization server and capture the user's refresh token or client secret. Grants now carry that binding and refresh only against it — never through re-discovery.

The commit history is organized so the mechanics ("what") are readable per change; this description covers the reasoning and the operational impact below.

#### Backwards compatibility / upgrade impact

- **No action needed at upgrade.** Existing tokens and dynamically-registered clients load in the old format and keep working; API routes, connect/disconnect UX, and the webapp contract are unchanged.
- **One-time reconnect per user.** Pre-upgrade tokens lack the authorization-server binding, so when a user's current access token expires the plugin clears the grant and shows the normal "connect" prompt once, rather than refreshing through re-discovery (the hijackable path). Timing depends on the provider's access-token lifetime (often within the hour; immediate if already expired). Anyone mid-consent during the deploy re-starts authorization (OAuth sessions live ~10 min).
- **Server modes.** The bundled MCP server serves `2026-07-28` in stateless mode; stateful sessions negotiate `2025-11-25` and older; `/sse` remains for `2024-11-05` clients.

#### Known limitation

- **Genuine legacy *go-sdk* SSE servers won't connect.** v1.7.0 sends `server/discover` before `initialize` even on the SSE fallback, and its SSE writer kills the connection when an old server rejects that method at the HTTP layer, so the `initialize` fallback can't run. This needs an upstream SDK fix (v1.7.0 is the latest release; no knob to skip discovery). Legacy SSE servers built on the TypeScript/Python SDKs are unaffected (verified in the compatibility matrix).

#### Testing

- Unit + race suites green (38 packages), lint clean. New tests cover cross-protocol negotiation, the full OAuth flow (discovery → DCR → PKCE+resource authorize → iss-verified callback → pinned refresh with rotation) against a go-sdk server, HA concurrency/lease behavior, and the split-issuer discovery topology that broke the real Atlassian MCP server.
- **Manual docker matrix:** our client verified against real TS SDK servers (SSE-only 1.5.0, 1.10.2, 1.12.1, 1.30.0) and Python SDK 2.0.0, negotiating `2024-11-05`/`2025-03-26`/`2025-11-25`/`2026-07-28`; our HTTP server verified against an old go-sdk v1.4.1 client and a current v1.7.0 client in both modes. Real Atlassian MCP OAuth connect confirmed working.
- Pre-existing, unrelated: the `//go:build integration` tests in `mcp/` have an import cycle present on `master` and don't compile.

#### Ticket Link

None

#### Release Note

```release-note
Upgraded the MCP SDK to v1.7.0, adding support for MCP protocol version 2026-07-28 in both the MCP client (tool connections) and the Mattermost MCP server, while maintaining backwards compatibility with older protocol versions (2025-11-25, 2025-06-18, 2025-03-26, 2024-11-05). The Mattermost MCP server serves protocol 2026-07-28 in stateless mode. Migrated MCP OAuth to the SDK's auth packages with stricter spec-compliant validation, fixed refreshed OAuth tokens not being persisted (which broke servers using refresh-token rotation), and pinned OAuth grants to their issuing authorization server. Users connected to an OAuth MCP server before upgrading are prompted to reconnect once, the next time their access token expires.
```

<div><a href="https://cursor.com/agents/bc-7f7b8b46-ee40-438a-891c-8bcd9f246be6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7f7b8b46-ee40-438a-891c-8bcd9f246be6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **New Features**
  * Added OAuth discovery/config caching with persisted tokens, automatic refresh, and re-auth handling.
  * Expanded OAuth support across Streamable HTTP and legacy SSE connections.
  * Added OAuth resource-metadata extraction coverage and improved protocol-version compatibility for HTTP.
* **Bug Fixes**
  * Prevented crashes when `tools/list` contains `null` tool entries.
  * Improved reliability of surfacing OAuth-needed errors from `401/403` responses.
* **Documentation**
  * Documented the `--stateless` option and HTTP protocol-version compatibility.
* **Breaking Changes**
  * Removed Dynamic Client Registration (RFC 7591) support.


mcp/oauth_manager.go
```diff
@@ -11,15 +11,24 @@ import (
...
 	"github.com/mattermost/mattermost-plugin-agents/v2/mmapi"
+	"github.com/modelcontextprotocol/go-sdk/oauthex"
 	"golang.org/x/oauth2"
 )
 
 const (
-	clientID                = "mattermost-mcp-client"
+	// oauthClientName is the human-readable RFC 7591 client_name shown on
+	// authorization server consent screens.
+	oauthClientName         = "Mattermost Agents"
```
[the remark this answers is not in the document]
crspeller: "Mattermost Agents" is a brand name so should not be translated.

crspeller:
@coderabbitai full review

mcp/oauth_transport.go
```diff
@@ -60,71 +70,36 @@ func (t *authenticationTransport) RoundTrip(req *http.Request) (*http.Response,
...
-	if token != nil {
-		oauthConfig, configErr := t.manager.createOAuthConfig(req.Context(), t.serverURL, "", t.staticCreds)
-		if configErr != nil {
-			return nil, fmt.Errorf("failed to create OAuth config: %w", configErr)
-		}
-
-		transport = &oauth2.Transport{
-			Source: oauthConfig.TokenSource(req.Context(), token),
-			Base:   transport,
+	tokenSource, err := t.handler.TokenSource(req.Context())
```
edgarbellot, line 89: This leaks the user's OAuth token to a different host if the MCP server ever redirects.

Go strips the `Authorization` header on cross-host redirects - but only if the header was set on the *original* request. Here it's set inside `RoundTrip` itself ([L78-92](https://github.com/mattermost/mattermost-plugin-agents/blob/958bf79312e8b9f22a3990561982db4648ed20d5/mcp/oauth_transport.go#L78-L92)), which runs again on every redirect hop. So the header gets reattached each time, and Go's protection never kicks in.

I confirmed this locally: a request that gets redirected to another host still arrives there with the real bearer token attached.

Impact: any MCP server that redirects to another domain (compromised, open redirect, bad proxy config) receives the user's OAuth token for that third-party account.

Only affects the legacy SSE transport ([http_client.go:51-71](https://github.com/mattermost/mattermost-plugin-agents/blob/958bf79312e8b9f22a3990561982db4648ed20d5/mcp/http_client.go#L51-L71)) - but a server can force that path just by failing the streamable attempt first.

Suggested fix: only attach the token if `req.URL` still matches the pinned server, or add a `CheckRedirect` that blocks cross-origin redirects. Same fix should apply to `doRefresh` in `oauth_refresh.go:175-190`, where a redirect from the token endpoint would leak the refresh token + client secret the same way.

mcp/oauth_store.go
edgarbellot: "Disconnect" doesn't actually disconnect anything at the provider.

[DeleteUserToken](https://github.com/mattermost/mattermost-plugin-agents/blob/958bf79312e8b9f22a3990561982db4648ed20d5/mcp/oauth_store.go#L258-L264) only deletes the token from our KV store. It never calls the authorization server's revocation endpoint. I verified this: after disconnecting, the same refresh token still works and mints a new access token at the provider.

So a user who clicks "disconnect" (or an admin doing offboarding) thinks access is cut off. It isn't. The refresh token stays valid indefinitely at the third party.

Suggested fix: call the AS's `revocation_endpoint` when deleting the token. The pieces are already there:
- `oauthex.AuthServerMeta` already exposes `RevocationEndpoint`.
- [resolveOAuthConfig](https://github.com/mattermost/mattermost-plugin-agents/blob/958bf79312e8b9f22a3990561982db4648ed20d5/mcp/oauth_manager.go#L219-L260) already reads the AS metadata (`authURL`, `tokenURL`, etc) - the revocation endpoint just needs capturing there too.
- It would slot in next to [TokenEndpoint in the stored envelope](https://github.com/mattermost/mattermost-plugin-agents/blob/958bf79312e8b9f22a3990561982db4648ed20d5/mcp/oauth_store.go#L58-L65), which already pins issuer/token endpoint/client for refresh - same pattern.

edgarbellot (changes requested):
Looking good! I only found these two issues, both of them reproducible. Let me know if you have any questions

edgarbellot (approved):
Thank you Christopher!

nickmisasi (approved):
LGTM! Excited to standardize oauth

PR #967, opened 2026-08-07
Attach Anthropic sandbox-created files to agent replies automatically

crspeller:
#### Summary

When Claude creates a file in Anthropic's code-execution sandbox, the user previously couldn't get it — Anthropic returns a provider-side file id, not the bytes. This PR attaches those files to the reply **automatically**: copying a file into the sandbox's `$OUTPUT_DIR` is the model's share gesture, since Anthropic reports file ids only for files left there. There is no attach tool, and the ids are never shown to the model.

**Pipeline:**
1. **Capture** — `ServerToolUse` carries the `file_ids` a code-execution invocation created; the toolrunner records them on the request context in observation order.
2. **Download** — `*bifrost.LLM` implements a new `llm.ProviderFileDownloader` via Bifrost's Files API, using the agent's existing credentials. Works whether Anthropic is the primary provider or a fallback route.
3. **Attach** — after the turn, the server uploads each file and attaches it via the same `AddCreatedFile` → `post.FileIds` flow as `CreateFile`, with the same shared policy checks: attachments enabled, requester upload permission, size limit (checked against provider metadata before downloading), filename sanitization, and the per-reply attachment cap. A bad file is skipped with a log, never failing the reply.

A prompt-template block teaches the model to copy deliverables into `$OUTPUT_DIR` and leave scratch files out. OpenAI code-interpreter container files use a different retrieval API and remain out of scope (documented).

**Also in this PR:**
- Native tool activity is now replayed to the model in later requests as a labeled summary (previously display-only), so follow-up turns know what ran and what was attached.
- Webapp: streamed replies split into visual rounds when provider tool activity or thinking follows text, with live/persisted parity — including fixes for duplicated text at round splits (websocket payloads are cumulative) and for activity status updates landing in already-frozen rounds.
- Review cleanups: one shared resolved-tool-call predicate for streaming and web-search annotation, `ServerToolUse.Clone()` replacing hand-rolled deep copies, dead webapp helpers removed, and assorted small Go simplifications.

**QA test steps:**
1. On an Anthropic agent with **Code Execution** enabled, ask for something that produces a file (e.g. "generate a CSV of the first 20 primes and attach it").
2. The response post carries the CSV attachment; the code-execution activity card shows the run.
3. Ask for a file to be created but *not* placed in the output directory → it is not attached.
4. Verify server file-attachment settings, size limits, and the requesting user's upload permission are enforced (rejections are logged; the reply still posts).

#### Ticket Link

None

#### Release Note
```release-note
Agents using Anthropic's Code Execution native tool now automatically attach files the model copies into the sandbox output directory to their replies. Server attachment settings, file size limits, the per-post attachment cap, and the requesting user's upload permission are enforced. Native tool activity is now replayed to the model in later conversation turns as a summarized record.
```


nickmisasi (approved):
LGTM. It looks like fallback services won't get files attached but that's probably fine. Whether security will be ok with AI generated files or not is a different story :P

PR #970, opened 2026-08-12
feat(abac): add durable service and MCP IDs

nickmisasi:
#### Summary
- Give LLM services and MCP servers stable Mattermost-style 26-character IDs so later ABAC policies can attach to a resource that survives rename, reorder, and System Console edits.
- Mint IDs on save, carry them through admin config updates (atomic `UpdateConfig`), and run a one-time store migration that rewrites legacy UUID service IDs and assigns IDs to MCP / embedded / plugin servers.
- Reject stale or conflicting identity payloads instead of silently minting a new ID (that would detach any future policy).

This is layer 1 of the ABAC stack and **can merge on its own** — it does not need the Mattermost 11.10 ABAC platform APIs.

- #970 — durable IDs (this PR)
- #971 — [PEP package and access-policy API](https://github.com/mattermost/mattermost-plugin-agents/pull/971)
- #972 — [shared access policy editor](https://github.com/mattermost/mattermost-plugin-agents/pull/972)
- #973 — [host UIs](https://github.com/mattermost/mattermost-plugin-agents/pull/973)

Reviewers: focus on ID reconciliation (`config.ReconcileServiceIDs` / MCP equivalents), the one-time `store.MigrateABACIDs` transaction, and that the webapp round-trips IDs on config save without minting extras.

#### Test Plan
- [ ] Save a new LLM service and a new remote MCP server in System Console; confirm each gets a 26-character ID in `GET /plugins/mattermost-ai/admin/config`.
- [ ] Edit name/URL of an existing service or MCP server without changing identity; confirm the ID is unchanged after save.
- [ ] Upgrade (or activate against) a config that still has UUID service IDs; confirm they are rewritten once and `fallbackServiceID` references are remapped. Automation that hard-coded the old UUIDs must re-read admin config.
- [ ] Submit a stale/corrupt payload that tries to swap an existing service's ID; confirm the save is rejected.
- [ ] `make check` (or at least `make test` for `./config ./store ./api ./mcp` plus webapp unit tests for services/MCP config).

#### Release Note
```release-note
Added durable Mattermost-style 26-character IDs for LLM services and MCP servers. Existing UUID service IDs are rewritten once on upgrade; automation that hard-coded those UUIDs must re-read GET /plugins/mattermost-ai/admin/config.
```

## Summary by CodeRabbit

- **New Features**
  - Added per-channel agent auto-reply settings, with support for root posts and threads.
  - Added administrator controls to rebuild the vector index and configure HNSW search options.
[...]

PR #971, opened 2026-08-12
feat(abac): add PEP package and access-policy API

nickmisasi:
#### Summary
- Add the plugin-side policy enforcement point (`accesscontrol`) and HTTP authoring APIs so agents, LLM services, and MCP servers can be gated by Mattermost ABAC decisions.
- Enforce at existing call sites (agent/service use, MCP tool filtering, list/picker visibility, meta-tool omission). Any failure to obtain a decision denies; resources with no policy keep today's legacy behavior.
- **Do not merge until [mattermost/mattermost#37926](https://github.com/mattermost/mattermost/pull/37926) lands and the DEV-ONLY `server/public` replace pin in `go.mod` is dropped.** The pin currently points at an in-flight branch. After #37926 merges, repoint `require` to a master / `v0.4.x` of `server/public` that contains it — never merge a branch pin or a local-path replace.

Requires Mattermost server **≥ 11.11.0** with ABAC licensed (Enterprise Advanced). Already merged platform deps: mattermost/mattermost#37509, #37510, mattermost/enterprise#2231. Still waiting on **#37926** (plugin policy save self-inclusion and host-editor robustness). This plugin already calls `SaveAccessControlPolicy`; those save-path guards only apply once #37926 is on the server.

This is layer 2 of the ABAC stack (depends on #970):

- #970 — [durable IDs](https://github.com/mattermost/mattermost-plugin-agents/pull/970)
- #971 — PEP and access-policy API (this PR)
- #972 — [shared access policy editor](https://github.com/mattermost/mattermost-plugin-agents/pull/972)
- #973 — [host UIs](https://github.com/mattermost/mattermost-plugin-agents/pull/973)

Reviewers: decision table / fail-closed vs fail-open semantics in `accesscontrol`, route authz (agent managers vs system admins), and that request bodies never supply policy identity fields.

#### Test Plan
- [ ] Confirm `go.mod` still has the DEV-ONLY `replace` for `github.com/mattermost/mattermost/server/public` and treat removing it as a merge blocker until #37926 is in the module we require.
- [ ] On Mattermost ≥ 11.11 with ABAC licensed: CRUD `GET`/`PUT`/`DELETE` `/agents/:id/access_policy` as an agent creator/admin; system-admin-only CRUD for `/admin/services/:id/access_policy` and `/admin/mcp/:id/access_policy`.
- [ ] CEL helpers (`/access_control/cel/*`) work for users who can manage an agent; `/access_control/status` is available to any authenticated user.
- [ ] Agent in attribute-based mode with a matching policy allows the user; a non-matching policy denies. Service policy deny blocks completions for agents on that service. MCP server policy deny silently omits that server's tools.
- [ ] No policy on a resource: legacy agent allow/block lists still apply; services and MCP servers stay unrestricted.
- [ ] The plugin does not load on servers older than 11.11 (`min_server_version`).
- [ ] `make check` (or `go test` for `./accesscontrol ./api ./bots ./mcp ./llmcontext ./mcpserver`).

#### Release Note
```release-note
Added attribute-based access control APIs for agents, LLM services, and MCP servers, including GET/PUT/DELETE access_policy routes and CEL helper endpoints. Requires Mattermost server v11.11.0 or later with ABAC licensed.
```

## Summary by CodeRabbit

* **New Features**
  * Added attribute-based access control for agents, services, and MCP servers.
  * Added policy management, validation, testing, autocomplete, visualization, and availability APIs.
  * Added access-policy audit events and a new attribute-based agent access level.
* **Bug Fixes**
  * Restricted searches, tools, and agent interactions now return access errors when authorization fails.
  * MCP tools and servers are filtered according to access permissions.
* **Documentation**
  * Updated the minimum supported Mattermost Server version to 11.11.0.


go.mod
```diff
@@ -236,4 +236,14 @@ require (
...
 
+// DEV-ONLY: branch pin for the in-flight ABAC plugin API (EvaluateAccessControl returning
+// model.AccessDecision + PAP/CEL methods + plugin-owned "<pluginID>:<resourceType>" policy
+// keying). Currently the pseudo-version of the mattermost branch
+// cursor/agents-abac-platform-229e head (180df00618). proxy.golang.org does not serve this
+// commit yet, so go.sum carries hashes resolved straight from the repository; builds work
+// because the go command trusts go.sum and only consults the checksum database for entries
+// missing from it. Before merge: repoint to a MASTER pseudo-version of server/public
+// containing the merged API — never merge a branch pin, and never merge a local-path replace.
+replace github.com/mattermost/mattermost/server/public => github.com/mattermost/mattermost/server/public v0.4.4-0.20260730201402-180df006186a
```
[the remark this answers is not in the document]
nickmisasi: Leaving the pin. This stack depends on the in-flight server ABAC plugin API (`EvaluateAccessControl` returning `AccessDecision`, PAP/CEL methods, plugin-owned policy keying) that is not in a released `server/public` yet. The `go.mod` comment already marks this as a merge blocker: we will repoint to a master pseudo-version once that server work lands, and will not merge a branch pin.

api/api.go
```diff
@@ -334,6 +343,10 @@ func (a *API) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Reques
 	agentRouter.PUT("/:agentid", a.handleUpdateAgent)
 	agentRouter.DELETE("/:agentid", a.handleDeleteAgent)
 	agentRouter.POST("/:agentid/avatar", a.handleUploadAgentAvatar)
+	// Access policy authoring: agent managers.
+	agentRouter.GET("/:agentid/access_policy", a.handleGetAgentPolicy)
+	agentRouter.PUT("/:agentid/access_policy", a.handlePutAgentPolicy)
+	agentRouter.DELETE("/:agentid/access_policy", a.handleDeleteAgentPolicy)
```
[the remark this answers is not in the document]
nickmisasi: Fixed: PUT/DELETE for agent, service, and MCP access policies are now in the audit registry. Records include resource type and ID only; planted CEL/policy content is asserted absent. GET stays unaudited.

conversations/handle_messages_test.go
```diff
@@ -18,6 +19,12 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// newPassthroughAccessChecker builds an ABAC checker that always reports
+// no_policy, so tests exercise pure legacy permission behavior.
+func newPassthroughAccessChecker() *accesscontrol.Checker {
+	return accesscontrol.New(accesscontrol.PassthroughClient{}, nil, accesscontrol.NoMCPServerIDs, nil)
+}
```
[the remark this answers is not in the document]
nickmisasi: @coderabbitai These declarations are not in the same package, so they do not collide or break the build.

`conversations/handle_messages_test.go` is `package conversations` (internal tests). `conversations/test_helpers_test.go` is `package conversations_test` (external black-box tests). Each package needs its own unexported helper. Removing the internal one would break `package conversations` tests that call it (`handle_messages_test.go`, `loaded_state_flow_test.go`, `ask_user_question_flow_test.go`, and others).

api/api_agents.go
```diff
@@ -655,8 +698,14 @@ func (a *API) handleListServices(c *gin.Context) {
...
 	}
 
+	bypassPolicies := a.canBypassServicePolicies(userID)
 	services := make([]ServiceInfo, 0, len(cfg.Services))
 	for _, svc := range cfg.Services {
+		if !bypassPolicies {
+			if policyErr := a.accessChecker.CanUseService(c.Request.Context(), userID, svc.ID); policyErr != nil {
+				continue
+			}
+		}
```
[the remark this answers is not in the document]
nickmisasi: @coderabbitai This is intentional PEP fail-closed behavior, not a missing 500 mapping.

`CanUseService` (via `canUseResource`) already turns any evaluation failure into `ErrAccessDenied` and logs it at the checker (`ABAC resource evaluation failed`). These HTTP call sites therefore never see a distinct infrastructure error, so branching on `errors.Is(..., ErrAccessDenied)` here would be dead code.

`TestListServicesAppliesServicePolicies` already pins the list path: an evaluation error hides every service rather than failing the catalog with 500. Returning 500 from fetch-models would contradict that fail-closed contract. Leaving both sites as-is.

server/main.go
```diff
@@ -203,7 +206,7 @@ func (p *Plugin) OnActivate() error {
 	p.configMigrated = true
 
 	// ABAC ID migrations must run after the config.json->DB migration and
-	// before bots are ensured, so EnsureBots persists agents that already use
+	// before the legacy config bot migration, so migrated bots copy
 	// already-remapped service IDs into Agents_UserAgents.
 	idsMigrated, err := runABACIDMigrations(p.API, pluginAPI, p.store, &p.configuration)
```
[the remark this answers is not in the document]
nickmisasi: This was fixed in the lower PR. ID migration runs first and the losers wait. Then all GetConfig when finished.

benrrr56-wq:
Adding a plugin-side enforcement point (accesscontrol) that gates agents, LLM services, and MCP servers by Mattermost ABAC decisions is the right placement: the PEP should sit on the request path, not enumerate-time. Two seams:
1. MCP servers present tools dynamically. If the ABAC decision is evaluated against tool meta at tools/list time but the call args change at tool/call runtime, the same decision id could authorize a different concrete call. Bind the decision to the materialized call (resolved tool plus args digest plus acting principal), not the advertised tool.
2. Record the decision id and the exact tool call it authorized on the audit path, so a later rule change cannot retroactively re-interpret what was permitted. A granted call and an evidenced call should be one record.

That decision-bound-to-call plus recomputable-evidence pattern is the core of what I build: control what the agent can do, see what actually executed, prove recomputably. On https://agentkey.us.

go.mod
crspeller: Looks like we can replace this now.

plugin.json
```diff
@@ -6,7 +6,7 @@
   "support_url": "https://github.com/mattermost/mattermost-plugin-agents/issues",
   "release_notes_url": "https://github.com/mattermost/mattermost-plugin-agents",
   "icon_path": "assets/bot_icon.png",
-  "min_server_version": "11.9.0",
+  "min_server_version": "11.11.0",
```
crspeller, line 9: Do we need a min version bump? Or can we just disable this feature?

webapp/src/components/rhs/rhs_error_boundary.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+type State = {
+    error: Error | null
+}
+
+// Catches render exceptions inside the Agents RHS so Mattermost's plugin
+// error boundary does not swallow the real message. The fallback keeps
+// data-testid="mattermost-ai-rhs" on the parent and exposes the exception
+// at data-testid="mattermost-ai-rhs-error" for e2e assertions.
+export default class RHSErrorBoundary extends React.Component<Props, State> {
```
crspeller: Looks like this agent went on a side quest and solved that RHS error too.

e2e/helpers/mmcontainer.ts
```diff
@@ -17,7 +17,8 @@ const defaultUsername        = "admin";
 const defaultPassword        = "admin";
 const defaultTeamName        = "test";
 const defaultTeamDisplayName = "Test";
-const defaultMattermostImage = "mattermostdevelopment/mattermost-enterprise-edition:release-11.9";
+// release-11.10 does not implement plugin EvaluateAccessControl (mattermost#37509 landed on master / 11.11).
```
crspeller, line 20: Just going to leave this one comment about this, but there are lots of comments in this PR that are slop.

api/api.go
```diff
@@ -65,8 +66,8 @@ type MCPClientManager interface {
 	MarkOAuthNeeded(userID, serverName, authURL string) error
 	GetEmbeddedServer() mcp.EmbeddedMCPServer
 	EnsureMCPSessionID(userID string) (sessionID string, created bool, err error)
-	GetToolsForUser(ctx context.Context, userID string) ([]llm.Tool, *mcp.Errors)
-	RefreshToolsForUser(ctx context.Context, userID string) ([]llm.Tool, *mcp.Errors, error)
+	GetUserToolsAccess(ctx context.Context, userID string) mcp.UserToolsAccess
```
crspeller: Why the rename? Of course the function should check for access controls.

accesscontrol/checker.go
```diff
@@ -0,0 +1,324 @@
...
+		return nil, errNoDecision
+	}
+	return decision, nil
+}
+
+// CanUseAgent gates end-user use of an agent, combining the resource policy
+// with the agent's UserAccessLevel. legacyCheck supplies the legacy
+// allow/block outcome; attribute-based mode never invokes it. Any failure to
+// obtain a decision denies unconditionally, in every agent mode.
+func (c *Checker) CanUseAgent(ctx context.Context, userID string, cfg *llm.BotConfig, legacyCheck func() error) error {
```
crspeller, line 98: I think this legacy check thing is something like in the first PR we can remove and just trust the migration to happen.

bots/permissions.go
```diff
@@ -110,19 +112,40 @@ func UsageRestrictionsForUserConfig(client *pluginapi.Client, cfg llm.BotConfig,
...
+// CheckUsageRestrictionsForUserConfig is the composite per-request user gate:
+// the ABAC agent decision (with the legacy UserAccessLevel switch as the
+// legacyCheck closure), then the ABAC service decision for cfg.ServiceID.
+// Every user-attributable completion entry point funnels through here.
+func (m *MMBots) CheckUsageRestrictionsForUserConfig(ctx context.Context, cfg llm.BotConfig, requestingUserID string) error {
+	legacy := func() error { return UsageRestrictionsForUserConfig(m.pluginAPI, cfg, requestingUserID) }
+	if err := m.accessChecker.CanUseAgent(ctx, requestingUserID, &cfg, legacy); err != nil {
+		return wrapDeny(err)
+	}
+	if err := m.accessChecker.CanUseService(ctx, requestingUserID, cfg.ServiceID); err != nil {
```
crspeller, line 129: I think this needs to check the whole fallback chain or something right? The fallback is automatic within Bifrost and doesn't report to us.

api/api_agents.go
```diff
@@ -746,16 +804,33 @@ func (a *API) handleFetchModelsForService(c *gin.Context) {
...
+	} else {
+		// Do not use a.bots here: agent list/get routes are not bot-middleware-gated and a.bots may be nil.
+		legacy := func() error { return bots.UsageRestrictionsForUserConfig(a.pluginAPI, *cfg, userID) }
+		agentOK = a.accessChecker.CanUseAgent(ctx, userID, cfg, legacy) == nil
+	}
+	if !agentOK {
+		return false
+	}
+
+	if a.canBypassServicePolicies(userID) {
```
crspeller, line 763: There is a bypass for system admins here but not for the actual checks, unless this is built into the ABAC policy stuff?

crspeller (changes requested):
Should we also apply restrictions to using the MCP servers though the external MCP server? 

Another thing going on here is that if a user that was previously denied an MCP server is then allowed to connect to it, it won't because the non-connected user client won't try to reconnect to that server.

accesscontrol/checker.go
nickmisasi: "legacy" in this context is the old allow/block/none lists for users and teams. It's not migration related

api/api_agents.go
nickmisasi: Intentional split - system admins skip the service policy only on list/get so they can see the catalog, since they're the only authors right now. Everything else (ie, the actual _usage_ of the service) goes through `CanUseService`

bots/permissions.go
nickmisasi: Yep good call. Rather than requiring access to the full chain, I have opted to truncate the chain at the point that a fallback is inaccessible. Ex:

A, B allowed, C denied → use A, fall back to B, drop C
A, C allowed, B denied → use A only; drop B and C

I think this will be less cumbersome from a UX perspective when configuring the fallbacks

plugin.json
nickmisasi: I'd prefer to keep it as a min bump (this will actually need to bump to v12 because of some hardening on the Plugin API side here: https://github.com/mattermost/mattermost/pull/37926 that hasn't merged yet) for simplicity. Juggling version checks to shut off entire features adds a bunch of branching I'd like to avoid. We can't just treat a missing API as disabled because things need to fail closed

go.mod
nickmisasi: We can but we'll need to repin for https://github.com/mattermost/mattermost/pull/37926 once merged.

api/api.go
nickmisasi: From the API level this is about getting the high level servers that the user has access too so it's semantically different. GetToolsForUser still exists elsewhere and is called to get the actual tools list itself. Basically, this one prevents "Github" showing up in the tools list at all, while the latter only would have "github" just with an empty list

e2e/helpers/mmcontainer.ts
nickmisasi: Did a deslop pass

webapp/src/components/rhs/rhs_error_boundary.tsx
nickmisasi: Removed from this PR

crspeller (approved):

api/api_access_control.go
```diff
@@ -0,0 +1,464 @@
...
+
+type celTestRequest struct {
+	ResourceType string `json:"resource_type"`
+	Expression   string `json:"expression"`
+	Term         string `json:"term"`
+	After        string `json:"after"`
+	Limit        int    `json:"limit"`
+}
+
+func (a *API) handleCELTest(c *gin.Context) {
```
edgarbellot, line 425: Here we serialize `result.Users` to JSON. The store's `SearchUsers` does call `u.Sanitize(map[string]bool{})` first, but clearing `email`, `first_name`/`last_name`, `auth_service`, `auth_data` only runs when `len(options) != 0`, and an empty map skips it. 

The core equivalent calls `SanitizeProfile(user, hasSystemPermission)`, whose `options` come from `PrivacySettings.ShowEmailAddress`/`ShowFullName`, so it's never empty and the redaction applies. This path bypasses those instance privacy settings and also returns the LDAP DN / SAML/OAuth identifier via `auth_data`.

Should the users be run through `SanitizeProfile` before `c.JSON`, like the core handler does?

api/api_access_control.go
```diff
@@ -0,0 +1,464 @@
...
+	if err := a.accessChecker.DeletePolicy(c.Request.Context(), userID, accesscontrol.ResourceTypeMCP, server.ID); err != nil {
+		abortPolicyRequest(c, err)
+		return
+	}
+	c.Status(http.StatusOK)
+}
+
+// celRouteAuthzRequired: system admins and agent managers pass; otherwise a
+// per-agent admin passes when agent_id names an agent they manage.
+func (a *API) celRouteAuthzRequired(c *gin.Context) {
```
edgarbellot: This function passes anyone accepted by [canConfigureAgentServices](https://github.com/mattermost/mattermost-plugin-agents/blob/f9ff1e85873d54d8549d328635f738df02f242f5/api/api_agents.go#L212-L221), which includes `PermissionManageOwnAgent` - a default permission on `system_user`, so any member clears this gate. The core equivalent requires `ManageSystem` for the unrestricted query, or self-inclusion for delegated admins; neither applies here.

Would it make sense to restrict `/cel/test` (and `/cel/autocomplete/fields`) to `ManageSystem`/`ManageOthersAgent`, or add a self-inclusion check?

accesscontrol/checker.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+		return nil
+	}
+	deny := func() error {
+		err := fmt.Errorf("agent %s: %w", cfg.ID, ErrAccessDenied)
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "access denied")
+		return err
+	}
+	attributeBased := cfg.UserAccessLevel == llm.UserAccessLevelAttributeBased
```
edgarbellot: Attribute-based agents treat `no_policy` (including the case when no policy exists) as allow, and never fall through to `legacyCheck` in that mode. Combined with [handleDeleteAgentPolicy](https://github.com/mattermost/mattermost-plugin-agents/blob/f9ff1e85873d54d8549d328635f738df02f242f5/api/api_access_control.go#L155-L170), which only requires `canManageAgent`, a single DELETE turns a policy-gated agent into one usable by anyone.

Would it be worth denying instead of allowing on `no_policy` for this mode, or reverting `UserAccessLevel` as part of the delete?

accesscontrol/checker.go
```diff
@@ -0,0 +1,285 @@
...
+	if decision == nil {
+		return nil, errNoDecision
+	}
+	return decision, nil
+}
+
+// CanUseAgent combines the resource policy with cfg.UserAccessLevel.
+// Attribute-based mode never invokes legacyCheck. Any failure to obtain a
+// decision denies in every agent mode.
+func (c *Checker) CanUseAgent(ctx context.Context, userID string, cfg *llm.BotConfig, legacyCheck func() error) error {
```
edgarbellot, line 98: Attribute-based agents treat `no_policy` (including the case when no policy exists) as allow, and never fall through to `legacyCheck` in that mode. Combined with [handleDeleteAgentPolicy](https://github.com/mattermost/mattermost-plugin-agents/blob/f9ff1e85873d54d8549d328635f738df02f242f5/api/api_access_control.go#L155-L170), which only requires `canManageAgent`, a single DELETE turns a policy-gated agent into one usable by anyone.

Should we deny instead of allowing on `no_policy` for this mode, or reverting `UserAccessLevel` as part of the delete?

api/api_agents.go
```diff
@@ -690,16 +739,31 @@ func (a *API) handleFetchModelsForService(c *gin.Context) {
 	c.JSON(http.StatusOK, models)
 }
 
-// canUserAccessAgent reports whether userID may view or use the agent (admin, then usage restrictions).
-func (a *API) canUserAccessAgent(cfg *llm.BotConfig, userID string) bool {
+// canUserAccessAgent reports whether userID may see the agent on list/get.
+// System admins see agents even when CanUseService would deny (they author
+// policies). Everyone else must pass the agent gate and CanUseService for the
+// primary service. Denied fallback hops are truncated per request, not here.
+func (a *API) canUserAccessAgent(ctx context.Context, cfg *llm.BotConfig, userID string) bool {
```
edgarbellot, line 746: This short-circuits on `cfg.IsAdmin(userID)`, so the ABAC decision is skipped for the agent's creator or anyone in `AdminUserIDs` - who aren't necessarily system admins. Combined with the delete path above, an agent admin can also remove the policy and end up using an agent it was meant to deny them.

Is this by design?

api/api.go
```diff
@@ -570,7 +601,7 @@ func (a *API) usesServiceAccountAuth(bot *bots.Bot) bool {
 }
 
 // getAIBotsForUser returns all AI bots available to a user
-func (a *API) getAIBotsForUser(userID string) []AIBotInfo {
+func (a *API) getAIBotsForUser(ctx context.Context, userID string) []AIBotInfo {
```
edgarbellot, line 606: nit: [EvaluateAccessRequest](https://github.com/mattermost/mattermost-plugin-agents/blob/71c2f88b710fad54accee4a09c5e3b147334072a/accesscontrol/pdp_client.go#L31-L53) has no caching, and `getAIBotsForUser`/`handleListAgents` call it twice per agent (agent + service) on every webapp load. On servers without Enterprise Advanced licensed, `EvaluatePluginAccessRequest` falls through to `resolvePluginPolicyExistence`, which runs an uncached SQL read (`Store().AccessControlPolicy().Get`).

So every install without ABAC pays 2N uncached DB reads per bot-list load, plus 2 more per message in auto-reply channels, for a feature they don't use. Not sure if a short TTL cache (like the 30s one on `IsAvailable`) or a fast path when `IsAvailable()` is false would be worth adding here.

api/api.go
```diff
@@ -348,6 +356,10 @@ func (a *API) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Reques
 	agentRouter.PUT("/:agentid", a.handleUpdateAgent)
 	agentRouter.DELETE("/:agentid", a.handleDeleteAgent)
 	agentRouter.POST("/:agentid/avatar", a.handleUploadAgentAvatar)
+	// Access policy authoring: agent managers.
+	agentRouter.GET("/:agentid/access_policy", a.handleGetAgentPolicy)
+	agentRouter.PUT("/:agentid/access_policy", a.handlePutAgentPolicy)
```
edgarbellot, line 361: Agent policy `PUT`/`DELETE` only requires `canManageAgent` (agent creator or `AdminUserIDs`), while service/MCP policy routes sit under `/admin` with `mattermostAdminAuthorizationRequired`. That sits next to [serviceAccountChangeNeedsAdmin](https://github.com/mattermost/mattermost-plugin-agents/blob/71c2f88b710fad54accee4a09c5e3b147334072a/api/agent_acl.go#L113-L123), which deliberately requires `manage_system` to widen `UserAccessLevel`/`UserIDs`/`AdminUserIDs` while service account auth is on - the policy is the same widening, just via a different door.

A non-admin creator can `DELETE` the policy on an attribute-based agent with service account auth and land on the `no_policy` allow, opening an agent with admin MCP credentials to everyone. 

Should we require `manage_system` for this route when `UseServiceAccountAuth` is set?

edgarbellot (changes requested):

api/api_agents.go
nickmisasi: Yes - admins (both agent and system level) may need to be able to access the agent to configure it. This access is limited only to the agents page itself, and does not allow them to skip _actually chatting with it_

The other thing to keep in mind is that while someone could theoretically ratchet themselves into being able to chat with the agent, the underlying service and tools _also_ must be accessible. Those are likely the more sensitive things anyways.

api/api_access_control.go
nickmisasi: Doing your suggestion would mean that only privileged users can test policies which doesn't follow the pattern that exists in core product. In the Channel Settings modal you can configure a Membership Policy, and as a non admin you are able to use the test access rule button as well as have autocomplete in the policy creator (which is only the text entry for system admins, in both cases)

api/api_access_control.go
nickmisasi: Yep. Matched users now go through the same profile sanitization as core. Privacy settings still drive email/fullname, sysadmins get those plus auth service, and auth_data is blanked for everyone. Left the route itself as-is (same audience as Channel Settings “Test access rule”).

accesscontrol/checker.go
nickmisasi: Agreed this was too open. Attribute-based agents now deny when there’s no policy, so deleting it no longer makes the agent usable by everyone. Creators/admins can still list/get it and PUT a new policy. Didn’t refuse DELETE or flip the access level back — recovery is just writing a policy again.

api/api.go
nickmisasi: Fair on the extra reads, especially on installs that never use ABAC. Leaving it out of this PR — not merge-blocking. Fine as a follow-up if it shows up in practice.

Main driver here is being careful not to fail open if a license was removed and downgraded vs temporarily inaccessible for whatever reason

api/api.go
nickmisasi: Yep, same door as the service-account ACL changes. With service account auth on, agent policy PUT/DELETE now require manage_system. GET is unchanged (still whoever can manage the agent). Non-SA agents are unchanged.

api/api_access_control.go
nickmisasi: Looked into this a bit more and chatted with Maria. Realized I missed a key piece. The channel settings modal only shows you the full list of people if you yourself meet the attributes. I'm pushing a commit to align with that shortly

api/api_access_control.go
```diff
@@ -0,0 +1,558 @@
...
+	}
+	return cfg
+}
+
+func auditPolicyMutation(c *gin.Context, resourceType, resourceID string) {
+	audit.AddParam(auditRec(c), audit.KeyABACResourceType, resourceType)
+	audit.AddParam(auditRec(c), audit.KeyABACResourceID, audit.TruncateID(resourceID))
+}
+
+func (a *API) handleGetAgentPolicy(c *gin.Context) {
```
edgarbellot, line 117: nit, non-blocking: Noticed `GET` still returns rule expressions unmasked - core redacts literal values the caller doesn't hold via `MaskPolicyExpressions`.

So an agent's creator (non-admin) can read the raw literal a sysadmin wrote, e.g. `clearance == "TS/SCI"`.

Is it intentional to leave `GET` unmasked?

edgarbellot (approved):
@nickmisasi Thank you for addressing everything! Just one minor question.

PR #975, opened 2026-08-14
feat: default HNSW m to 8 and rebuild the vector index without re-embedding

nickmisasi:
#### Summary
- Added a top-level `hnswM` embedding-search setting (default 8, range 2–100) so admins can shrink HNSW graph RAM without changing vector width.
- Added `POST /plugins/mattermost-ai/admin/reindex/rebuild-vector-index` and a Console **Rebuild vector index** action that drops and recreates the HNSW index with the current `m` without clearing or re-embedding posts. Search stays available across an `m` change until rebuild; a full reindex is still required for model/provider/dimension changes.
- Existing deployments stay compatible on upgrade when `hnsw_m` is missing. Rebuild is rejected (and the Console action is disabled) if stored embedding identity differs, and a successful rebuild updates only HNSW `m`.

image

#### Release Note
```release-note
Added an HNSW m setting (default 8) and a Rebuild vector index admin action that recreates the index without re-embedding posts.
```

## Summary by CodeRabbit

* **New Features**
  * Added administrator controls to rebuild the vector index without re-embedding content.
  * Added configurable HNSW M settings with validated defaults and limits.
  * Added compatibility checks for embedding models, dimensions, providers, and HNSW settings.
  * Added confirmation dialogs, progress tracking, warnings, and status messages for rebuilds.
  * Added audit records for vector-index rebuild operations.
* **Bug Fixes**
  * Improved handling of incompatible, concurrent, failed, and interrupted indexing jobs.
* **Documentation**
  * Updated administration guidance for HNSW configuration, memory usage, rebuilding, and search availability.


indexer/rebuild_vector_index.go
```diff
@@ -0,0 +1,125 @@
...
+		deleteCursorOnSuccess: false,
+		persistHNSWMOnly:      true,
+	})
+}
+
+// claimRebuildVectorIndex takes ownership of the vector-index phase key so
+// search gates and live writes skip during building. Leftover dropped/building
+// claims are adopted; leftover repairing is converted to dropped (the index
+// will be replaced) while preserving BuildStartedAt for the repair window.
+func (s *Indexer) claimRebuildVectorIndex(jobID string) (*deferredRun, error) {
```
crspeller, line 97: Should we check the current state is in a state that makes sense before continuing here? Not sure if the UI allows this or not but it probably make sense. Otherwise we could claim a partially completed index and build over that.

webapp/src/components/system_console/embedding_search/reindex_section.tsx
```diff
@@ -308,11 +316,19 @@ export const ReindexSection = ({
                             values={{nodeId: jobStatus?.node_id || 'unknown'}}
                         />
                         <StaleActions>
-                            {hasProgress && (
+                            {jobStatus?.resumable && hasProgress && (
```
crspeller: I think this hides the resume normally? Since we don't set resumable for regular indexing jobs for some reason? Maybe something deeper wrong here.

postgres/pgvector_index_def.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil || m <= 0 {
+		return 0, false
+	}
+	return m, true
+}
+
+// vectorIndexDefinitionOK is true when the catalog def is a full-table HNSW
+// index on embedding with vector_l2_ops and the expected m. Partial indexes
+// and expression indexes cannot serve all searches.
+func vectorIndexDefinitionOK(indexDef string, wantM int) bool {
```
crspeller, line 51: Is this too strict? Old indexes won't have the wantM set correctly right?

crspeller (changes requested):

postgres/pgvector_index_def.go
nickmisasi: We have to treat M missing as not ok otherwise changing to 8 then rebuilding would just run CREATE IF NOT EXIST, persist 8, and then say "everything's good" when it didn't change anything. We don't nag or auto rebuild, this just allows the next rebuild to catch up to the configured m value easily

crspeller (approved):

PR #978, opened 2026-08-14
feat: add index retention window and Catch Up hole-fill

nickmisasi:
#### Summary
- Adds `indexRetentionDays` so admins can limit indexing to the last N days (`0` indexes all posts) without dropping rows already in the index.
- Increasing N uses Catch Up from the retention floor with `NOT EXISTS`, so only the gap is embedded and search stays available. Decreasing N tightens live/reindex bounds; search can hide older rows until a later Full Reindex.
- Catch Up is a durable job operation: resume keeps hole-fill, and the job records only the snapshotted window it actually covered.
- Stacks on #976 (halfvec) and #975 (HNSW `m` rebuild).

#### Release Note
```release-note
Added an Index posts from the last N days setting (`indexRetentionDays`). Raising N and running Catch Up backfills older posts without a Full Reindex; lowering N does not delete already-indexed posts.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable semantic-search index retention by number of days.
  * Added Catch Up indexing when retention is expanded, with status guidance.
  * Added vector-index rebuild controls and expanded index health details.
  * Added clearer handling for retention changes during active indexing jobs.

* **Bug Fixes**
  * Search and index health counts now respect the retention window.
  * Incompatible indexing requests now return clear HTTP 400 errors.

* **Documentation**
  * Updated administration guidance for retention, Catch Up, and Full Reindex operations.


webapp/src/components/system_console/embedding_search/reindex_section.tsx
```diff
@@ -363,6 +387,36 @@ export const ReindexSection = ({
...
+                        <br/>
+                        <FormattedMessage defaultMessage='The index now looks further back. Run Catch Up to embed older posts that are not already in the index. Search stays available — do not Full Reindex unless you also changed the embedding model or vector precision.'/>
+                    </WarningText>
+                </WarningBanner>
+            )}
+
+            {hasLocalRetentionTighten && !hasLocalModelMismatch && !hasLocalRetentionWiden && (
+                <NoteBanner>
+                    <NoteText>
+                        <FormattedMessage defaultMessage='Lowering this does not remove already-indexed posts. Search can hide older rows. Full Reindex if you want the index smaller.'/>
```
crspeller: Do we really want to force a re-index to lower the index window?

indexer/retention.go
```diff
@@ -0,0 +1,126 @@
...
+package indexer
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/mattermost/mattermost-plugin-agents/v2/embeddings"
+)
+
+func retentionDaysPtr(days int) *int {
```
crspeller: Really? Also the value function below.

crspeller (changes requested):

webapp/src/components/system_console/embedding_search/reindex_section.tsx
nickmisasi: Thought about this a bit and I've decided to adjust so that this is ore clearly just around the initial backfill. As I had initially implemented it, if you initially backfilled 365 days, 100 days later, the index would be 465 days' worth, but you'd only be able to access 365. That felt weird IMO. Now, search once again works against the entire index. 

You can't shrink the HNSW index without recalculating it - currently, we have no support for bounded HNSW indexes (ie, CREATE INDEX IF NOT EXISTS WHERE post created after date x and before date y) and adding them would increase the complexity significantly. I think we should leave as is for now and in a later iteration look at doing an embedding deletion of any posts outside the window (when it changes), that way we don't need to maintain a partial index, and it can "cheaply" just rebuild HNSW without re-embedding.

indexer/retention.go
nickmisasi: This is primary leveraged by tests (17 call sites) where the alternative would be
```
days := 365
doSomeThingThatRequiresPointer(&days)
```

Which gets weird in for looped sub tests. 

I created a shared helper and opened https://mattermost.atlassian.net/browse/MM-70407 to address a bunch of other areas (like bifrost) where there are similar functions

crspeller (approved):

PR #983, opened 2026-08-19
Make LLM Bridge services independent of agents

crspeller:
#### Summary

**The problem.** The bridge modelled a service as reachable only *through* an agent: `/completion/service/:service` picked an arbitrary agent on that service and inherited its model override, reasoning, tools and ACLs, and discovery listed only services some agent referenced. So "call OpenAI directly" silently meant "call OpenAI as whichever agent happened to exist", and a service with no agent was unreachable. This PR makes the split honest: **agent endpoints use agents, service endpoints use configuration.**

Three decisions worth reviewer attention:

**1. Service calls deliberately have no authorization story.** Agent ACLs answer "who may talk to this agent"; a service call has no agent, so there is nothing to check. `user_id`/`channel_id` become attribution-only and inter-plugin trust is the whole boundary. Consequence: `GET /services` no longer filters by `user_id`. Nothing leaks (`BridgeServiceInfo` is `{id, name, type}`), but callers using it to scope a picker must now filter themselves. The same applies to the per-user fallback-chain trimming master added for ABAC: it wraps agent LLMs only, so a service call behaves identically with or without `user_id`.

**2. Structured output is a service property, not an agent one.** Native JSON-schema support belongs to the provider/model, so a per-agent toggle was in the wrong place. New per-service `structuredOutputPolicy` (`auto`/`native`/`prompt_fallback`; empty = `auto`). `auto` sends a native schema only when provider + model + API path is *positively known* to accept one: a rejected schema turns every structured-output feature into a hard error, while the prompt fallback only costs some accuracy.

This shifts effective defaults on upgrade **in both directions** — the riskiest part of the PR:

- Agents that had the old toggle *on* would silently drop to prompt fallback, so `store.MigrateStructuredOutputPolicies` runs at activation and sets `native` on services those agents reference. Only when the policy is unset, so it is idempotent and never overrides an admin. It must run *before* any agent save, because the new webapp omits `structuredOutputEnabled` and a save zeroes the stored value. It follows the same shape as the ABAC ID migration: one atomic `UpdateConfig` under a cluster mutex, run *after* the ID migration (it matches agents to services by ID) and *before* runtime config is loaded, so an HA follower that finds nothing left to migrate still loads the winner's policies. It scans both stored agents and legacy `config.Bots` not yet copied into the agents table, and re-runs after a late legacy-bot migration, so bots that arrive on a later config update are covered too.
- The reverse is **not** migrated: direct-OpenAI services move from prompt fallback to native strict schemas. Internal callers are safe (`jsonschema.For` already emits strict-safe schemas), but an external caller posting a loose hand-written schema can now get a provider 400. Called out in the release note, with `prompt_fallback` as the escape hatch. (No upgrade-guide entry: `docs/upgrading_to_2.0.md` documents an already-released version and shouldn't be retro-edited, and the migration makes the rest automatic for admins.) **This is the judgement call to sanity-check.**

**3. Service LLMs are cached because they are expensive.** Each Bifrost client starts a ~1,000-goroutine worker pool, so one per request is not viable. The registry caches one per service ID keyed on the exact config it was built from; superseded entries shut down once their in-flight requests drain, and deactivation is synchronous. An earlier iteration also guarded a publish-time race and kept per-service build locks; both were dropped, since the lease-time config comparison already stops a stale entry being *served*. `ReconcileServiceLLMs` was kept — without it a deleted service's client survives until deactivation.

Also:

- `auto`'s table is prefix-matched by model family, so non-text `gpt-4o-*` variants (audio/realtime/search) and Gemini embedding models are explicitly excluded — the prefix alone would wrongly mark them capable.
- An unrecognized `structuredOutputPolicy` is now a 400 on save. Previously it passed unvalidated into storage, then silently disabled every agent on that service while the System Console still showed "Auto".
- `public/bridgeclient/README.md`: the documented streaming loop asserted `event.Value.(error)`, which panics on every server-reported error — server errors arrive as a string over JSON. Fixed.

**QA steps**

1. A service with a `defaultModel` and **no agent** appears in `GET /bridge/v1/services` and completes by ID and by name, streaming and non-streaming.
2. Send `json_output_format` to a known-capable `auto` combo (native) and an unknown one (prompt fallback); then set `native`/`prompt_fallback` and confirm they override detection.
3. Upgrade an install with an agent that had structured output on → its service comes back with policy `native`.
4. `PUT /admin/config` with `"structuredOutputPolicy": "Native"` → 400, not a silently disabled service.

**Verification.** After merging `master` (ABAC IDs/PEP, service-account MCP auth, the Go modernization): `make check` passes end to end — `check-style` (golangci-lint 2.13.1, eslint, tsc), `make test` (5,468 Go tests incl. the Docker-backed `store`/`threads` suites, 734 webapp tests), `check-shards`, `check-i18n`. Both `package-lock.json` files are byte-identical to master's (`package.json` is unchanged), so `check-locks` matches master on CI's Node 24. New `store.TestMigrateStructuredOutputPolicies` covers stored agents, legacy `config.Bots`, admin-set policies, no-op runs (no config row written) and the second-run/HA-follower case against a real Postgres; `TestBuildLLMFallbackAccessOnlyForAgents` pins that a denied fallback trims an agent's chain but leaves a service call's request untouched.

#### Screenshots

Service editor shows a new "Structured output" selector defaulting to Auto, with "Native supported" and "Prompt fallback" as the alternatives; the agent editor's Advanced section no longer has a structured output control.

#### Release Note

```release-note
Bridge service endpoints now operate directly on configured services: GET /bridge/v1/services lists every bridge-capable configured service (agent-less services included; user_id is still accepted but no longer filters), and POST /bridge/v1/completion/service/:service resolves the stored service (ID before name) and runs its default model with no agent settings and no agent ACL checks. Added a per-service structuredOutputPolicy setting (auto/native/prompt_fallback; empty means auto) controlling native JSON-schema structured output for both agent and direct service calls; an unrecognized value is now rejected with HTTP 400 when saving configuration. On activation, services referenced by agents that had the deprecated per-agent structuredOutputEnabled setting turned on are migrated to structuredOutputPolicy=native. The per-agent structuredOutputEnabled setting was removed from the UI and is ignored at runtime, though it is still accepted in API payloads and retained in the database as a deprecated no-op. For direct OpenAI services the effective default changes from prompt-based JSON instructions to native strict schemas; set the service to prompt_fallback to keep the previous behaviour. Token usage logs gained service_id and service_name fields, and direct service calls log blank agent dimensions.
```

<div><a href="https://cursor.com/agents/bc-1e98f40c-d16d-485c-9128-dc2ebf9706d6?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-1e98f40c-d16d-485c-9128-dc2ebf9706d6&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

- **New Features**
  - Added per-service structured output policies: Auto, Native supported, and Prompt fallback.
  - Improved service discovery, resolution, fallbacks, eligibility checks, and model reuse.
  - Added service-specific token usage attribution.

- **Changes**
  - Removed structured output controls from agent configuration; policies are now managed per service.
  - Improved service completion permissions, tool handling, and documentation.

- **Bug Fixes**
  - Improved validation, fallback behaviour, model acquisition failures, configuration updates, and shutdown handling.


nickmisasi (commented):
Nothing to add other than Cursor's finding around the HA nodes missing the migration for local config (shoutout bugbot, solid finding). Will re review when that's addressed

api/api_agents.go
```diff
@@ -80,28 +80,31 @@ func abortAgentRequest(c *gin.Context, status int, err error) {
...
+	Model                 string               `json:"model"`
+	EnableVision          bool                 `json:"enableVision"`
+	DisableTools          bool                 `json:"disableTools"`
+	EnabledNativeTools    []string             `json:"enabledNativeTools"`
+	ReasoningEnabled      bool                 `json:"reasoningEnabled"`
+	ReasoningEffort       string               `json:"reasoningEffort"`
+	ThinkingBudget        int                  `json:"thinkingBudget"`
+	// StructuredOutputEnabled is deprecated: it is accepted and persisted for
+	// compatibility with existing callers, but ignored at runtime. Structured
+	// output is a per-service policy (ServiceConfig.StructuredOutputPolicy).
```
fmartingr:
```suggestion
	// Deprecated: StructuredOutputEnabled it is accepted and persisted for
	// compatibility with existing callers, but ignored at runtime. Structured
	// output is a per-service policy (ServiceConfig.StructuredOutputPolicy).
```

fmartingr (approved):

PR #990, opened 2026-08-25
webapp: fix stuck spinner on remounted tool-only posts

nickmisasi:
#### Summary
Tool-only agent replies leave `post.message` empty, so on remount `precontent` was treated as still starting. The **Working...** spinner (formerly Starting...) sat on top of persisted tool cards even though the agent was idle.

That showed up as:
- The requester rejecting a tool call, leaving the thread, refreshing, and coming back
- An onlooker opening a thread paused on a pending Ask Every Time approval

The spinner now shows only when there is nothing to render yet, or while generation is actually resuming (`continue`/`start`). The placeholder is **Working...** because the same row also appears mid-turn, not only at first start.

**QA**
1. In a channel where the agent has tools and in-channel tool calling is enabled, ask it to use a tool set to Ask Every Time.
2. As the requester: reject the tool call, close the thread, refresh, reopen. Confirm no Working... spinner and the rejected tool card is still visible.
3. As an onlooker: open the thread while a tool call is waiting for approval. Confirm Thinking + tool cards are visible, Working... is not, and Accept/Reject are not shown.
4. Accept a pending tool call. Confirm Working... appears while the continuation starts, then clears when the next content arrives.

#### Screenshots
| before | after |
|----|----|
| Starting... spinner stuck above rejected or pending tool cards after remount | Terminal post: Thinking + tool cards, no spinner. Live continuations show Working... |

#### Release Note
```release-note
Fixed an infinite in-progress spinner on agent posts after a tool call was rejected or when an onlooker opened a thread paused on tool approval. Relabeled the placeholder from Starting... to Working....
```

## Summary by CodeRabbit

- **New Features**
  - Added support for displaying provider server-tool activity during conversations.
  - Added clearer labels and messages for agent configuration, automatic replies, MCP tools, web search, code execution, vector indexing, and recency settings.
  - Improved rendering of search results and related error messages.

- **Bug Fixes**
  - Improved conversation ID validation and handling of invalid search results.
  - Loading status now consistently displays **“Working...”** and correctly reflects conversation state.
  - Tool-only conversations no longer show an unnecessary loading indicator.


webapp/src/components/llmbot_post/llmbot_post.tsx
```diff
@@ -404,6 +404,10 @@ export const LLMBotPost = (props: LLMBotPostProps) => {
         setExpandedReasoning((prev) => ({...prev, [roundId]: !collapsed}));
     };
 
+    // Tool-only posts leave post.message empty, so precontent stays true on
+    // remount; hide Starting... once persisted or live rounds exist.
+    const showStarting = renderedRounds.length === 0 && (precontent || (conversationLoading && !generating));
```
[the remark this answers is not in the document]
nickmisasi: Agreed — `continue` sets `precontent` on purpose while generation resumes over the prior tool round.

The gate now shows Starting... when `precontent && generating`, even if persisted rounds already exist. Remounts with existing rounds (reject / onlooker pending) still hide it because `generating` is false.

Covered in `shows Starting... after continue while generation resumes over persisted rounds`.

webapp/src/components/llmbot_post/llmbot_post.tsx
```diff
@@ -404,6 +404,13 @@ export const LLMBotPost = (props: LLMBotPostProps) => {
         setExpandedReasoning((prev) => ({...prev, [roundId]: !collapsed}));
     };
 
+    // Tool-only posts leave post.message empty, so precontent stays true on
+    // remount. Hide Working... once rounds exist, except while generation is
+    // actually resuming (continue/start still set precontent on purpose).
+    const showStarting =
```
crspeller: Should rename to working if it's called working now.

crspeller (changes requested):
This should be against master.

webapp/src/components/llmbot_post/llmbot_post.tsx
nickmisasi: Renamed `showStarting` to `showWorking` to match the label.

crspeller (approved):

PR #993, opened 2026-08-25
Parallelize and bound MCP connection and tool discovery

crspeller:
#### Summary

Parallelizes MCP connection and tool discovery while bounding reconnect load and preserving deterministic ordering, partial success, OAuth behavior, caching, and proxy collision rules.

- Uses local standard-library `sync.WaitGroup.Go` loops with buffered-channel semaphores, capped at 32 tasks independently for runtime connection, admin discovery, and external proxy discovery. There is no generic concurrency package or executor abstraction.
- Adds a separate per-`ClientManager` admission cap of 32 complete remote/plugin connection sequences across all users on a node. Embedded in-memory connections do not consume permits.
- Starts each server's 30-second connection timeout only after node admission, so queue time does not consume the upstream budget.
- Unblocks queued work on manager shutdown, prevents late/stale sessions from committing, and treats local admission failures as retryable rather than sticky upstream failures.
- Connects eligible remote, embedded, and plugin MCP servers in one runtime batch.
- Incrementally populates each user's connection cache with per-origin singleflight; remote failures remain sticky and plugin failures remain retryable.
- Resolves agent, user, admin, registration, and license eligibility before contacting a server.
- Preserves unaffected user sessions and remembered outcomes across `ReInit`. Tool-policy, `ExposeExternal`, and idle-timeout-only changes keep exact client/session pointers; URL, name, headers, credentials, enabled state, registration, plugin path, and embedded server changes invalidate only affected origins.
- Serializes runtime planning with config/plugin identity transitions while keeping admission waits, MCP network calls, and session closes outside lifecycle locks.
- Discovers admin and external proxy tools concurrently, then merges in snapshot order.
- Rejects duplicate remote server names and canonically equivalent endpoint URLs. Existing stored conflicts do not block activation; every conflicting row is skipped and reported in admin discovery.

The embedded server is reused across plugin configuration changes because its constructor does not depend on plugin config. Failed construction remains retryable, and the concrete pointer is retained until success to avoid a typed-nil interface being mistaken for a live server.

#### Measured effect

Tested end-to-end against four real `@modelcontextprotocol/server-everything` instances over streamable HTTP, each behind a TCP relay adding 250ms per client chunk:

| Remote servers | Admin discovery before | After | Runtime refresh before | After |
|---|---:|---:|---:|---:|
| 1 | 1.56s | 1.54s | 1.31s | 1.29s |
| 2 | 3.09s | 1.54s | 2.59s | 1.29s |
| 4 | 6.18s | 1.56s | 5.15s | 1.30s |

The implementation discovers all 56 remote tools plus 119 embedded tools.

[mcp_cold_discovery_before_after.log](https://cursor-plugin-pr-assets.s3.amazonaws.com/cursor/cursor/parallel-mcp-connection-discovery-2611/mcp_cold_discovery_before_after.log)

Five healthy MCP servers

[Cold discovery walkthrough](https://cursor-plugin-pr-assets.s3.amazonaws.com/cursor/cursor/parallel-mcp-connection-discovery-2611/admin_mcp_tools_cold_discovery_four_real_servers.mp4)

Stored duplicate conflicts

[Stored-conflict walkthrough](https://cursor-plugin-pr-assets.s3.amazonaws.com/cursor/cursor/parallel-mcp-connection-discovery-2611/admin_mcp_tools_reports_duplicate_server_conflicts.mp4)

#### Validation

- Cross-user real-protocol stress test: 10 users simultaneously connect to four remote and two plugin MCP servers; aggregate network concurrency remains at or below 32 and embedded connections bypass admission.
- Queue-vs-timeout, canceled cache-warmer, shutdown-unblock, admission-close race, non-sticky local failure, late-session, and stale-commit tests.
- Table-driven identity coverage for remote name/URL/header/credentials/enabled/conflict changes; plugin name/path/enabled/registration changes; embedded replacement/disable; and identity-neutral tool policy, exposure, and idle-timeout changes.
- Immediate unregister → no reconnect → re-register → exactly one fresh connection coverage.
- In-flight invalidation and concurrent ReInit/GetTools/refresh/register/unregister stress tests.
- Focused admission and preservation tests pass 20–30 repeated runs under `-race`.
- `gotestsum -- -race ./...`: 4,619 Go tests passed, 130 eval-only tests skipped.
- `go vet ./...` and repository `golangci-lint`: clean.
- E2E shard and i18n drift checks: clean.
- Final plugin bundle builds successfully.

#### Release Note

```release-note
Improved cold-start latency for MCP tools by connecting to eligible MCP servers and discovering their tools concurrently instead of one server at a time. Runtime reconnects are capped across users per node, and unaffected MCP sessions are preserved across configuration updates. MCP servers excluded by agent, user, admin, registration, or license policy are not contacted. Duplicate MCP server names and equivalent endpoint URLs are rejected when saving configuration; existing duplicates are excluded and reported in the admin MCP tools view until corrected.
```

<div><a href="https://cursor.com/agents/bc-d553b58f-c119-4221-9bd5-302d5fd52611?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d553b58f-c119-4221-9bd5-302d5fd52611&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **New Features**
  * MCP tool access can be limited to permitted server origins and licensing conditions.
  * Tool discovery and connection setup now run concurrently for faster results.
  * Remote MCP connections have a 30-second timeout.
  * Tool refreshes rediscover eligible servers and preserve partial results when some servers fail.
  * Plugin tools are discovered concurrently with deterministic conflict handling.
  * MCP catalogues now include all enabled servers.

* **Bug Fixes**
  * Configuration saves reject duplicate server names or equivalent endpoints with a clear error.
  * Disabled, conflicting, or unlicensed servers are excluded from discovery.
  * Plugin requests now stop promptly when their request context expires.


nickmisasi (commented):
I think this is a great idea as it would be awesome to speed up the loading time of the tools lists, but I’m worried about scalability particularly around reconnect stampedes. Mechanisms for single-user limits are a great start, but I think we need to consider cross-user limits on a per node basis as well. We theoretically could have had a similar problem prior to this with a very large instance - but now we’re widening the amount of reconnects by n times, it could get messy

nickmisasi (approved):

PR #995, opened 2026-08-26
Modernize and simplify the Go codebase

crspeller:
#### Summary

Modernizes the Go codebase and applies a broad, behavior-preserving simplification pass — roughly **−3,000 net lines** across ~280 files in focused commits.

**Toolchain**
- Go held at **1.26.7** in all three modules (root, `loadtest/controller`, `cmd/evalviewer`): the FIPS toolchain registry (`cgr.dev/mattermost.com/go-msft-fips`) tops out at 1.26.7 — no 1.27 image exists yet even though Microsoft shipped go v1.27.0-1. The re-upgrade to 1.27 is mechanical and documented (tag-probe + bump procedure in `build/fips.mk` and AGENTS.md); `build/fips.mk` now points at `go-msft-fips:1.26.7-dev` (digest to be pinned from this PR's CI run).
- Mechanical `go fix` modernizer pass (145 files): `new(expr)` pointer creation, `strings.Builder`, `slices.Backward` iteration, etc. (1.27-only rewrites — promoted-field literals, stdlib `uuid` — were subsequently reverted to stay compatible with the 1.26 FIPS toolchain.)
- `GOLANGCI_LINT_VERSION` bumped v2.0.2 → v2.13.1; repo lints clean with the new version uncapped.

**Dependencies dropped** (direct): `github.com/pkg/errors`, `github.com/hashicorp/go-multierror` (replaced by stdlib error wrapping).

**Dead code removed**: ~30 verified-unreachable functions/fields/params (verified via `deadcode` across all four binary entrypoints plus manual reference checks). Test-only LLM helpers moved out of the shipped binary into `llm/llmtest` (the production `llm` package no longer links `testing`).

**Deduplication highlights**
- `mcpserver/tools`: generic `mcpTool[T]` constructor replaces ~125 six-line registration literals (−326 lines; registration output verified byte-for-byte identical).
- `llm/loadtest_validation.go`: 264-line hand-copied fork of `loadtest/profile.go` replaced by one shared `loadtest/profile` package.
- `bifrost`: stream-loop duplication (watchdog, tool-call flush ×4, reasoning-end ×5) extracted; the 2,552-line `bifrost.go` split into 8 files by concern.
- `api`: streaming/no-stream bridge handler pairs merged; 7-value return tuple → plan struct; agent request shape ×4 → one embedded struct.
- Similar passes over `conversations`, `bots`, `store`, `streaming`, `mcp`, `mcpserver/auth`, `indexer`, `chunking`, `embeddings`, `format`, `evals`, `cmd/evalviewer`, `build/pluginctl`, `toolrunner`.

**Behavior-adjacent fixes**
- Error-string matching in `api/` replaced with exported sentinel errors (`indexer.ErrJobAlreadyRunning`, `ErrNotRunning`, `ErrNoPreviousIndex`, `ErrNotConfigured`, `meetings.ErrNotMeetingBotPost`, `ErrNoTranscriptionPostReference`). This also **fixes a latent bug**: the handler compared against "not a calls or zoom bot post" while the producer says "not a meeting bot post", so the intended 400 response had silently become a 500.
- `config.Container` accessors are now consistently nil-safe (previously half guarded, half panicking).
- Removed never-set-in-production search options (`MinScore`, `CreatedAfter`, `CreatedBefore`) and their SQL/filter plumbing.

Known cosmetic change: gin binding-error messages for agent create/update now include `AgentRequestFields` in the field namespace (embedded struct).

Deferred items (ctx propagation, a `ClientManager.closeChan` race, `OnActivate` restructuring, the Go 1.27 re-upgrade recipe, and more) are catalogued for follow-up PRs.

**QA**: full unit suite (including container-based postgres/store/mcpserver tests) passes; `-race` pass on concurrency-heavy packages; golangci-lint clean (uncapped); `make check-style`, `check-shards`, `check-i18n`, `check-locks` all green.

#### Ticket Link

NONE

#### Release Note

```release-note
NONE
```

## Summary by CodeRabbit

- **New Features**
  - Added service-account authentication for AI agents and MCP tool catalogs, with authorization and licensing checks.
  - Added provider file downloads, including support for files returned through fallback providers.
  - Improved chat and response streaming for reasoning, tool activity, citations, annotations, and structured output.
  - Preserved the ordering of text, reasoning, and server-tool activity in conversation history.
  - Added configurable load-test profiles and enhanced MCP tool registration and validation.

- **Changes**
  - Removed embedding search minimum-score and creation-date filters.


conversation/service.go
```diff
@@ -47,8 +60,9 @@ type BotLookup interface {
 
 // Service manages conversation entities: creation, continuation,
 // CompletionRequest building, turn writing, and title generation.
+// The embedded Store's methods are promoted onto Service.
 type Service struct {
-	store    Store
+	Store
```
nickmisasi: This feels a bit weird to me. Why have the service at all if you can just bypass it and call store directly? It gets weirder when you look at the overlap - Service and Store both define `CreateConversation`, but only Store defines `CreateTurn` so they become `s.Store.CreateConversation()` and `s.CreateTurn`

config/config.go
```diff
@@ -94,25 +94,33 @@ func (c *Container) Config() *Config {
...
 
+// load returns the current configuration, or a zero-value configuration when
+// none has been stored yet, so accessors are safe to call at any time.
+func (c *Container) load() *Config {
+	if cfg := c.cfg.Load(); cfg != nil {
+		return cfg
+	}
+	return &Config{}
+}
+
```
nickmisasi: Update explicitly supports storing `nil` still. I think we should adjust that so that store always puts &Config{} as well, or instead of, this?

nickmisasi (commented):
Overall LGTM. Flagged two things that I think we should consider/reconsider

nickmisasi (approved):

PR #1033, opened 2026-09-15
Add Cohere North as an LLM provider

crspeller:
#### Summary

Adds Cohere North (`north`) as a service type so existing Mattermost agents can use a North-hosted model as their backend.

North exposes an OpenAI Responses-compatible endpoint, so the integration rides on the existing Bifrost OpenAI provider with a custom base URL rather than a new transport:

- New `llm.ServiceTypeNorth`; validation requires an instance URL and a service token (stored in `apiKey`). The Responses API is always used (there is no Chat Completions endpoint on North).
- Base URL normalization accepts the instance root, `/api`, or `/api/v1` and always resolves to `{instance}/api`, which is where Bifrost's `/v1/responses` path must land.
- Every request to North is sent with `store: false` so agent turns are not persisted into the token owner's North conversation history. This is applied at the Bifrost provider level (`ProviderSettings.DisableStore` → `schemas.OpenAIConfig.DisableStore`), so it holds whether North is the primary or a fallback hop.
- Provider-native tools (web search, code interpreter) are declared unsupported for North; `SupportsNativeTools` now keys off `SupportedNativeToolsForServiceType` so an OpenAI-mapped type without native tools is handled correctly.
- Fallback-chain validation: Bifrost retries the primary's request body against fallbacks and strips native tools per Bifrost *provider* (North is `openai`), so `NewFromServiceConfig` rejects chains where North is a fallback for a Chat-Completions-path primary or for a primary with native tools enabled. The agent fails to start with an admin-facing error instead of 404ing or sending unsupported tools at failover.
- Model listing: North's `/v1/models` returns provider labels, not model names, so a small `net/http` fetcher paginates `GET /v2/models` instead (request context threaded through, 4 MiB body cap). Older instances 404 on it; the UI falls back to the free-text model field as it already does for other providers.
- Admin model-fetch endpoints (`/admin/models/fetch`, `/agents/models/fetch`) accept `north` and require both key and URL.
- Webapp: "Cohere North" option with North-specific labels/help text ("North instance URL", "Service token"), no Org ID, no Responses-API toggle, default model left blank, vision and effort-based reasoning enabled, native tools checklist hidden.
- Docs: new Cohere North section in `docs/providers.md`; `docs/admin_guide.md` type lists updated.

Not in scope for this version: binding to North-side Agents/knowledge/citations, per-user tokens, transcription, embeddings. Making Bifrost fallback hops fully independent of the primary (so this class of per-service gating is unnecessary) is a broader change tracked separately.

QA steps: System Console > Plugins > Agents > Add a Service > type "Cohere North"; enter the instance URL, a North auth token (from the instance `/developer` page), and a model such as `command-a-03-2025`; assign to an agent and chat (tools, streaming, and reasoning work; image input needs a vision model such as `command-a-vision-07-2025`).

Verified end-to-end against a live North instance through `bifrost.NewFromServiceConfig`: streaming tool call, stateless multi-turn replay with tool result, effort-based reasoning, and failover from an unreachable Responses primary to North all succeed, and no conversations are created in the account.

#### Ticket Link

None.

#### Screenshots

N/A (new dropdown option and per-provider fields in the existing Service form; covered by unit tests).

#### Release Note

```release-note
Added Cohere North as a supported LLM service type. Configure it with your North instance URL, a service token, and a default model in System Console > Plugins > Agents.
```

<div><a href="https://cursor.com/agents/bc-33fe88da-96b6-4481-b7e8-70bca4d79241?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-33fe88da-96b6-4481-b7e8-70bca4d79241&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

- **New Features**
  - Added Cohere North as a supported LLM provider.
  - Added North model discovery with pagination, authentication, validation, and request cancellation.
  - Added North support for vision, reasoning, streaming, and configurable fallbacks.
  - North requests always use the Responses API and do not persist responses.
  - Added North configuration fields, validation, model fetching, and localized help text.

- **Bug Fixes**
  - Improved validation for missing North API keys or instance URLs.
  - Ensured unsupported native tools are excluded for North services.

- **Documentation**
  - Documented North setup, capabilities, authentication, reasoning, model fetching, and fallback requirements.


crspeller:
@coderabbitai full review

docs/providers.md
```diff
@@ -108,6 +109,54 @@ Obtain a [Cohere API key](https://dashboard.cohere.com/api-keys), then select **
...
+
+- Provider-native tools (web search, code interpreter)
+- Transcription
+- Embeddings
+
+### Special Considerations
+
+Requests are sent with `store: false`, so agent conversations are not persisted in the token owner's North conversation history.
+
+North can only be used as a fallback for a primary service that uses the Responses API and has no provider-native tools enabled; otherwise the agent fails to start with a configuration error.
```
davidkrauser, line 156: I've been thinking about this, and I wonder if we can provide better fallback behavior here. Specifically:
- Could we refactor or use an intermediate representation of a request so that we can swap between Responses-type APIs and Completion-type APIs?
- Could we allow it to respond to requests that expect native tool calls to be enabled, but have explicit instructions in the prompt that:
  - List the tools that would've been available had tool calling been enabled
  - Return some sort of error response if it wanted to use one of those tools to answer the prompt

Ideally, we would want to just return an error if the prompt requires a tool call, but the model makes that choice, so hard to know ahead of time.

A local LLM tells me that we already do these sorts of things with:
- `BifrostChatRequest.ToResponsesRequest()` already converts
- `llm/structured_output_fallback.go` strips schema from a request and appends instructions

This isn't blocking - just trying to think of ways to make this more useful in more cases 🙂

davidkrauser (approved):

docs/providers.md
crspeller: Agreed, we should improve fallback here. There have been a bunch of cases where it would have been useful.

